PR #5, opened 2015-06-10
Return Task rather than awaiting

benaadams:
Alternative to  #4; return task directly and don't create state machine for function with async/await as all return paths return the task and there are no chained tasks.


src/AspNet5/Startup.cs
```diff
@@ -7,10 +7,10 @@ public class Startup
     {
         public void Configure(IApplicationBuilder app)
         {
-            app.Run(async (context) =>
+            app.Run( (context) =>
```
davidfowl: Fix the spacing

PR #11, opened 2015-07-19
Fix kestrel benchmarks

benaadams:
Calling Write is fatal to performance here; WriteAsync makes it happy

You might want to revisit https://github.com/aspnet/KestrelHttpServer/blob/8cb66582154eec872e1c1ecbef0afe977739f454/src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseStream.cs#L63

The .Wait() is deadly...

### DIFF
--- a/src/AspNet5/Startup.cs
+++ b/src/AspNet5/Startup.cs
@@ -8,7 +8,6 @@ namespace AspNet5
     public class Startup
     {
         private static readonly byte[] _helloWorldPayload = Encoding.UTF8.GetBytes("HelloWorld");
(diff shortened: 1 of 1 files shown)


DamianEdwards:
Yeah, even though that's what's checked in it's not what we've been testing with :smile:. Also we've been testing using the Kestrel buffer branch which effectively makes the difference not matter as it just writes to a buffer and does the actual async write later (write-behind).

All that said, this change is what we should have in this repo.

benaadams:
As replied on twitter; glad you have more general solution for kestrel itself - I can imagine very many users doing something similar in their calling code.

PR #12, opened 2015-07-22
Managed RIO Server

benaadams:
Updated commit for #10


DamianEdwards:
Latest results below. Seems CPU use went up a bit and we lost peak RPS (down from >7m) as a result.

| Connections | Pipeline Depth | Bandwidth| RPS | CPU |
| ----- | ---- | --- | --- | --- |
| 1 | 1 | 25 Mbps | ~14k | 13% |
| 8 | 16 | 1.1 Gbps | ~855k | 37% |
| 8 | 512 | 7.3 Gbps | ~5.4m | 40% |
| 32 | 16 | 2.9 Gbps | ~2.2m | 80% |
| 128 | 16 | 6.0 Gbps | ~4.5m | 100% |
| 256 | 16 | 6.8 Gbps | ~5.1m | 100% |
| 512 | 16 | 6.9 Gbps | ~5.2m | 100% |
| 512 | 150 | 8.8 Gbps | ~6.7m | 84% |
| 512 | 512 | 9.1 Gbps | ~6.9m | 67% |

Eilon:
@DamianEdwards  CLA bot is happy.

benaadams:
Hopefully this will max network before CPU

DamianEdwards:
Looking forward to trying this one :smile:

benaadams:
Tried it on an Azure G5; need to put some SIMD fixes in as on that hardware Vector<byte>.Count operates on 32 bytes at the same time rather than 16 bytes (AVX 256 rather than SSE 128)

benaadams:
wrk -c 512 --latency -t 32 http://url/ -s ./scripts/pipeline.lua -- 148

Still a bit funny on NUMA - good but not great, probably need to do thread affinities there; but on G4: 9.0 Gbps at 59% CPU:

azure-g4-windows

At 6.8M requests per second (6,807,542). Latency of 92.ms (avg), 113.81 ms (std-dev), 1.27s (max), 50% < 43ms, 90% < 288ms.

azure-g4-linux

Bobris:
You do all crazy stuff with AVX and then use this in performance critical code???

                    var date = DateTime.UtcNow.ToString("r");
                    Encoding.UTF8.GetBytes(date, 0, dateBytes.Length, dateBytes, 0);

benaadams:
> You do all crazy stuff with AVX and then use this in performance critical code???

:wink: 

Lazy fix for it coming

benaadams:
Function could be faster, but still should cut down the string allocs by 6.9M per second at 9.1 Gbps, to 512 per second; so not a bad change...

Bobris:
I think you would found DateTime.UtcNow slow too (it is basically Kernel call). This is how simply I do it in Nowin:
https://github.com/Bobris/Nowin/blob/master/Nowin/DateHeaderValueProvider.cs

benaadams:
> This is how simply I do it in Nowin

That's much better than the convoluted improvement I was planning; gone for something similar.

RichiCoder1:
No dev comment, just wanted to say awesome job @benaadams :+1:

CIPop:
This is really cool.
@benaadams it would be very interesting to see how this compares with a native-only implementation of RIO: http://ctstraffic.codeplex.com/

benaadams:
@CIPop A reference implementation! 

Super useful to know; I couldn't really find anything other than the api docs building this so its mostly trial and error iteration; will definitely have a look at it; thank you :+1:

CIPop:
Glad to help :) 
I just learned about this yesterday. I'll update the [Registered I/O Extensions tracking issue in CoreFX](https://github.com/dotnet/corefx/issues/721) as well.

davidfowl:
Closing this out

PR #22, opened 2015-10-13
RioSharp http server

aL3891:
After seeing the perf of the rio server Ben build i felt Rio was something i really wanted to know more about, so i started my own .net version of the rio api over at https://github.com/aL3891/RioSharp

Its far from done but it would be really interesting to see what kind of perf it gets :)

It features a stream interface with a rolling buffer pool, and i have the ambition hack it into kestrel as well, but im not quite there yet.. another thing that would be interesting to see the impact of is the reusing of the same buffer for multiple requests,


aL3891:
Allright, fixed some glaring bugs in riosharp and also did some preliminary perf tests. 
for whatever reason the results are very different when running as a .net 4.6 console app and as a dnx console app, the result are as follows on my machine, localhost to localhost:
.net 4.6 console app: 13500 rps
dnx46 console app: 18500 rps
ben adams rio server: 19200 rps 
basichttpserver: 19640 rps

i have a feeling something is amiss with my setup, since regular sockets are even performing better than rio, but atleast they are comparable.. i'm going to do some more tests on vms instead to try and get some more network data

aL3891:
Allright, added some sockets options and discovered that pre allocated writes is actually a slot slower if you make a lot of them, when switching back to the stream style writing with a properly sized write buffer size perf on local host jumped to 2.5 million rps (localhost to localhost)

Currently it seems like the returns on piplining starts to diminish around 4096, at least on my machine and atleast using dnx, when i run on the big framework my record is 5.4 million rps with 16 connections and a pipeline of 8192.

It would be very interesting to see what you get, especially now when i'm my results are atleast comparable to bens :)

aL3891:
@DamianEdwards i know you're busy, but just let me know if you're not interested in this stuff and i can prioritize other stuff :)

davidfowl:
@aL3891 He's at a conference

aL3891:
oh ok, thanks for the heads up!

AspNetSmurfLab:
@aL3891 you shouldn't pipeline any higher than 16 as that's what the TechEmpower benchmarks do. Anything beyond that is kinda crazy. What numbers do you get at 256 connections and pipeline depth of 16?

aL3891:
i can try, right now i've been getting errors that the completion queues are full when i use a lot of connections but that should be solvable with some tweaking

aL3891:
all right, i made some bug fixes to the library that was causing the queue lengths not to be set correctly and ran some tests. unfortunately i'm not at my reference machine right now but i ran some tests against bens implementation and got ~4500 rps with his and ~4450 with mine. (256 connections, pipeline 16 deep)

i'll run more tests when i get home and can use a more powerful box.

aL3891:
Allright did some more tests now on my reference machine, seeing somewhere around 430-480 krps with my implementation and 450-500 krps with bens. i made some changes to the buffer implementation to precaclulate alot more values and it seems to have boosted perf significantly

aL3891:
hm, looks like i messed up my rebase a little, i'm in the process of dropping anew version of riosharp that should have much better perf, but im not quite there.. i'm also waiting for rc2 to start with the asp.net host

aL3891:
closing this one and opening a new one

PR #23, opened 2015-10-23
Skip AsyncLocal

benaadams:
Take advantage of https://github.com/aspnet/Hosting/pull/436

### DIFF
--- a/src/Benchmarks/Startup.cs
+++ b/src/Benchmarks/Startup.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Server.Kestrel;
 using Microsoft.Extensions.DependencyInjection;
 
@@ -13,6 +14,7 @@ namespace Benchmarks
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddMvc();
+            services.AddSingleton(typeof(IHttpContextAccessor), typeof(InertHttpContextAccessor));
         }
 
         public void Configure(IApplicationBuilder app)
--- a/src/Benchmarks/InertHttpContextAccessor.cs
+++ b/src/Benchmarks/InertHttpContextAccessor.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+using Microsoft.AspNet.Http;
+
+namespace Benchmarks
+{
+    public class InertHttpContextAccessor : IHttpContextAccessor
+    {
+        public HttpContext HttpContext
+        {
+            get
+            {
+                return null;
+            }
+
+            set
+            {
+                return;
+            }
+        }
+    }
+}


src/Benchmarks/Startup.cs
```diff
@@ -13,6 +14,7 @@ public class Startup
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddMvc();
+            services.AddSingleton(typeof(IHttpContextAccessor), typeof(InertHttpContextAccessor));
```
rynowak: This kills the MVC

rynowak: You might be OK as long as you don't touch tempdata.

benaadams: Not for the MVC used in the benchmarks :innocent:

rynowak: Sure, but there's no guarantee that we won't take more of a dependency on this as long as it exists.

benaadams: Is off in a non-nuget benchmarking project; rather than a main branch.

Maybe should add a big WARNING comment? Not sure of phrasing though... 

Probably good for webapi based stuff; inter-machine comms; not so good for user stuff where you need to login and have html rendered.

Will get some numbers together...

rynowak: Yeah I definitely get it. Very excited to this what the numbers will look like!

My comments are more about finding a long-term alternative than about the specifics.

benaadams: Not as exciting as I hoped but not bad: 598,375 -> 701,731 so +17%

benaadams:
1.0.0-rc1-15996 coreclr x64

`~/wrk$ wrk -c 256 -t 32 ... -s ./scripts/pipeline.lua -- 16`

**Pre**
Running 10s test @ ...
  32 threads and 256 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    16.25ms   40.55ms 424.58ms   97.43%
    Req/Sec    19.81k     5.66k   40.53k    67.28%
  6043429 requests in 10.10s, 570.58MB read
  Non-2xx or 3xx responses: 6043429
Requests/sec: 598375.95
Transfer/sec:     56.49MB

**Post**
Running 10s test @ ...
  32 threads and 256 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    10.53ms   37.09ms 392.25ms   97.34%
    Req/Sec    23.22k     5.82k   42.07k    83.36%
  7087031 requests in 10.10s, 669.11MB read
  Non-2xx or 3xx responses: 7087031
Requests/sec: 701731.12
Transfer/sec:     66.25MB

DamianEdwards:
Aren't those wrk outputs saying all the requests failed?

benaadams:
Eurgh... yeah; failed really fast though?

Let me add `/plaintext` to the url and redo them

benaadams:
Without returning 404s for every request they are about the same at around 625k rps and can both happily sustain it continuously for `-d 120` rather than cutting out previously due to high GC.

AyncLocal is 80-85% cpu and Non-AsyncLocal is 70-75% cpu; so less cpu but no extra rps :sob:

DamianEdwards:
Lower CPU is great though, as we expect when we get multi-loop support back that extra CPU will be used up.

DamianEdwards:
Also you could try enabling TCP no-delay as it's supported by Kestrel now.

benaadams:
no-delay pushes up cpu a little; no rps gain - however the servers are < 1ms apart so probably just increasing packet count. However, glad its there will need it for websocket comms and for higher latency connections :+1:

src/Benchmarks/Startup.cs
benaadams: Was bad results

DamianEdwards:
I made this change along with enabling no-delay just now.

PR #25, opened 2015-11-09
Enable multi-loop & perf fixes in JsonMiddleware

AspNetSmurfLab:


DamianEdwards:
This is actually a PR from me, just from the perf lab machines.

PR #30, opened 2015-11-24
A few tweaks to EF Fortunes scenario.

anpete:


anpete:
cc @DamianEdwards

src/Benchmarks/FortunesEfMiddleware.cs
```diff
@@ -44,9 +44,7 @@ public FortunesEfMiddleware(RequestDelegate next)
 
         private static async Task<IEnumerable<Fortune>> LoadRows(ApplicationDbContext dbContext)
         {
-            var result = new List<Fortune>();
-
-            result.AddRange(await dbContext.Fortunue.ToListAsync());
+            var result = await dbContext.Fortune.ToListAsync();
```
DamianEdwards: I didn't do this because I wasn't sure if EF did anything to pre-size the list as part of the materialization. That isn't allowed as per the benchmark rules: "The list data structure must be a dynamic-size or equivalent and should not be dimensioned using foreknowledge of the row-count of the database table."

src/Benchmarks/Data/ApplicationDbContext.cs
```diff
@@ -9,6 +9,6 @@ public class ApplicationDbContext : DbContext
     {
         public DbSet<World> World { get; set; }
 
-        public DbSet<Fortune> Fortunue { get; set; }
+        public DbSet<Fortune> Fortune { get; set; }
```
DamianEdwards: doh, i mistyped that thing everywhere

DamianEdwards: I ended up fixing this as it touched a lot more places (I misnamed the migration too :frowning:)

src/Benchmarks/FortunesEfMiddleware.cs
Eilon: Does foreknowledge mean that the *test* mustn't pre-allocate, or that the "innocent" DB call can't pre-allocate? In neither case does the *test* pre-allocate anything here. (I.e. it's not telling EF that there are 2 rows and to prepare for that.)

anpete: EF doesn't do anything special here. The ToList/Async operators are actually outside of EF - In Linq-to-objects and IX-Async respectively.

src/Benchmarks/Data/ApplicationDbContext.cs
anpete: Thanks. Guess I need to fix my dev env so I can actually build this stuff :cow: :boy:

src/Benchmarks/FortunesEfMiddleware.cs
DamianEdwards: The full wording (from https://www.techempower.com/benchmarks/#section=code):

4. Using an ORM, all Fortune objects must be fetched from the Fortune table, and placed into a list data structure. Tests that do not use an ORM will be classified as "raw" meaning they use the platform's raw database connectivity.
5. The list data structure must be a dynamic-size or equivalent and should not be dimensioned using foreknowledge of the row-count of the database table.
6. Within the scope of the request, a new Fortune object must be constructed and added to the list. This confirms that the data structure is dynamic-sized. The new fortune is not persisted to the database; it is ephemeral for the scope of the request.

My interpretation of that is, as long as `ToListAsync` simply creates a new empty `List<T>` without an initial size, then loops through the results of the enumerator and calls `Add`, then we can make this change. I did a quick search to try and find `ToListAsync` via source but quickly ended up in the weeds. Any pointers?

anpete: Here you go: https://github.com/Reactive-Extensions/Rx.NET/blob/master/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L613

DamianEdwards: Perfect, thanks! Safe to make that change then :smile:

Did you want to fix this PR up so I can merge or should I just make the change?

anpete: Just make the change :+1:

PR #31, opened 2015-11-26
Dapper: micro-optimizations

mgravell:
(from: dapper team) use QueryFirstOrDefault (in 1.50-beta4); let dapper worry about connection opening if single-op; use AsList to use pre-existing buffered list when available

### DIFF
--- a/src/Benchmarks/FortunesDapperMiddleware.cs
+++ b/src/Benchmarks/FortunesDapperMiddleware.cs
@@ -46,14 +46,13 @@ namespace Benchmarks
 
         private static async Task<IEnumerable<Fortune>> LoadRows(string connectionString, DbProviderFactory dbProviderFactory)
         {
-            var result = new List<Fortune>();
+            List<Fortune> result;
 
             using (var db = dbProviderFactory.CreateConnection())
             {
                 db.ConnectionString = connectionString;
-                await db.OpenAsync();
-
-                result.AddRange(await db.QueryAsync<Fortune>("SELECT [Id], [Message] FROM [Fortune]"));
+                // note: don't need to open connection if only doing one thing; let dapper do it
+                result = (await db.QueryAsync<Fortune>("SELECT [Id], [Message] FROM [Fortune]")).AsList();
             }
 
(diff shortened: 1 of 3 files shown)


mgravell:
(nods to DNFBOT in a cheery way)

src/Benchmarks/FortunesDapperMiddleware.cs
```diff
@@ -46,14 +46,13 @@ public FortunesDapperMiddleware(RequestDelegate next, string connectionString, D
...
+            List<Fortune> result;
 
             using (var db = dbProviderFactory.CreateConnection())
             {
                 db.ConnectionString = connectionString;
-                await db.OpenAsync();
-
-                result.AddRange(await db.QueryAsync<Fortune>("SELECT [Id], [Message] FROM [Fortune]"));
+                // note: don't need to open connection if only doing one thing; let dapper do it
+                result = (await db.QueryAsync<Fortune>("SELECT [Id], [Message] FROM [Fortune]")).AsList();
```
DamianEdwards: I'm assuming Dapper doesn't do any pre-sizing of the list, yes?

mgravell: If you mean "to match the expected data": then no, dapper doesn't do that, because it is consuming a raw `IDataReader` (`DbDataReader` on core-clr) API, and **doesn't know** the number of rows until it reaches the end. In this setup, it is essentially:

    List<T> buffer = new List<T>();
    while (await reader.ReadAsync(cancel).ConfigureAwait(false)) {
        buffer.Add(func(reader));
    }
    while (await reader.NextResultAsync(cancel).ConfigureAwait(false)) { }

(where `func` is a delegate that processes each row into the target data type)

benaadams: https://www.techempower.com/benchmarks/#section=code

mgravell: Thanks @benaadams ; I have reviewed the requirements; dapper does nothing that would violate Test 4 rule 5, if that  is the concern.

PR #34, opened 2015-12-06
Use PooledContextFactory

benaadams:
Also needs https://github.com/aspnet/HttpAbstractions/pull/501 to work

Took out `SupportsSendFile` as it doesn't exist anymore

/cc @DamianEdwards

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ nuget.exe
 debugSettings.json
 project.lock.json
 version.h
+.build/
 
 # Profiler result files, just in case they are left lying around :)
 /.vs/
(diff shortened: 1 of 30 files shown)


benaadams:
Pre-allocations for DefaultContext 1.12MB

before-allocs

Post-allocations for DefaultContext 160KB
 
after-allocs

Almost all of which is logging scope

before-allocs

Not quite sure what's going on with ThreadStatic, but don't use it much - take the locks out and it goes a bit weird.

AspNetSmurfLab:
Mixed results with this. For non-pipelined tests it seems to improve things, but for pipelined tests it makes this much worse (~10% drop in RPS).

rynowak:
Sounds like contention yo

benaadams:
Interesting. I had to put the locks in else the `ThreadStatic` was behaving weird; i.e. not behaving as one. Contention would make sense if it wasn't.

Need to do some investigation and might need to raise a corelcr bug.

benaadams:
Rookie mistake - didn't actually make the `ThreadStatic` `static`

DamianEdwards:
@benaadams would you mind rebasing this so we can re-measure pls?

benaadams:
O_o did I rebase on wrong thing?

benaadams:
Submitted new PR  #49

PR #36, opened 2015-12-09
Add Jil JSON test

NickCraver:
This moves the JSON.Net test to /json/newtonsoft (to match the other multi-implementation test patterns) and adds /json/jil as another JSON serialization test.

### DIFF
--- a/src/Benchmarks/JsonJilMiddleware.cs
+++ b/src/Benchmarks/JsonJilMiddleware.cs
@@ -7,19 +7,18 @@ using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Http;
-using Newtonsoft.Json;
+using Jil;
 
 namespace Benchmarks
 {
-    public class JsonMiddleware
+    public class JsonJilMiddleware
     {
         private static readonly Task _done = Task.FromResult(0);
-        private static readonly PathString _path = new PathString("/json");
-        private static readonly JsonSerializer _json = new JsonSerializer();
+        private static readonly PathString _path = new PathString("/json/jil");
 
         private readonly RequestDelegate _next;
         
-        public JsonMiddleware(RequestDelegate next)
+        public JsonJilMiddleware(RequestDelegate next)
         {
             _next = next;
         }
@@ -36,7 +35,7 @@ namespace Benchmarks
 
                 using (var sw = new StreamWriter(httpContext.Response.Body, Encoding.UTF8, bufferSize: 30))
                 {
-                    _json.Serialize(sw, new { message = "Hello, World!" });
+                    JSON.Serialize(new { message = "Hello, World!" }, sw);
                 }
 
                 return _done;
@@ -46,11 +45,11 @@ namespace Benchmarks
         }
     }
     
-    public static class JsonMiddlewareExtensions
+    public static class JsonJilMiddlewareExtensions
     {
-        public static IApplicationBuilder UseJson(this IApplicationBuilder builder)
+        public static IApplicationBuilder UseJilJson(this IApplicationBuilder builder)
         {
-            return builder.UseMiddleware<JsonMiddleware>();
+            return builder.UseMiddleware<JsonJilMiddleware>();
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/Benchmarks/project.json
```diff
@@ -15,7 +15,8 @@
     "Microsoft.AspNet.Server.WebListener": "1.0.0-*",
     "Microsoft.AspNet.StaticFiles": "1.0.0-*",
     "Microsoft.Extensions.WebEncoders": "1.0.0-*",
-    "Newtonsoft.Json": "7.0.1"
+    "Newtonsoft.Json": "7.0.1",
+    "Jil": "2.13.0-*"
```
davidfowl: Can we use a specific version here?

src/Benchmarks/JsonJilMiddleware.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public Task Invoke(HttpContext httpContext)
+        {
+            // We check Ordinal explicitly first because it's faster than OrdinalIgnoreCase
+            if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal) ||
+                httpContext.Request.Path.StartsWithSegments(_path, StringComparison.OrdinalIgnoreCase))
+            {
+                httpContext.Response.StatusCode = 200;
+                httpContext.Response.ContentType = "application/json";
+                httpContext.Response.ContentLength = 30;
```
mgravell: This is copied from JsonMiddleware.cs, I note - these hard-coded 30s smell a bit like cheating, though, no? (in both cases; I'm not limiting this to #36)

davidfowl: Cheating why?

benaadams: If you want to go fast you should be doing `httpContext.Response.Headers["Content-Length"] = "30";` as is done in the [plaintext middleware](https://github.com/aspnet/benchmarks/blob/master/src/Benchmarks/PlaintextMiddleware.cs#L34)

As the numeric form [does a bunch of conversions](https://github.com/aspnet/HttpAbstractions/issues/407)

mgravell: @davidfowl it just seems odd to hard-code the content length, rather than either ask the serializer what it *will* be, or (more commonly) buffer the output, write the length of the buffer, then write the buffer. I realize the intent is to remove the buffer, but .... meh, maybe I'm over-thinking it.

NickCraver: My only goal was to put the two on equal footing - I'm all for changing both to be more realistic, though.

src/Benchmarks/project.json
NickCraver: Yeah sure - I wasn't sure what the desire was since Jil's in beta for a bit while some missing corefx bits are added (waiting on https://github.com/dotnet/corefx/issues/4543 for another rev). I'll update to be specific though.

mgravell: Everything except Newtonsoft.Json uses wildcard... just sayin'

There are pros and cons of either approach; it saved me having to issue a PR when I last rev'd "dapper" ;p

DamianEdwards:
@NickCraver would you mind rebasing this so we can measure?

NickCraver:
@DamianEdwards done!

davidfowl:
This PR is stale now, needs a do-over.

NickCraver:
I've paused RC2 work since things are still changing and we don't want to eat the time on the library side. I've ported Sigil (needed for Jil) but Jil still needs a fair amount of work for core. Given that many of the API differences that make it more difficult to port (and maintain performance) might be coming back...I'm personally inclined to leave it until that happens and I'm fairly sure Kevin's in the same boat.

That being said, I know Kevin would be happy to take an RC2 pull request on it. Due to RC1 tooling craziness current master doesn't support core at all (it was torn out to get a release out the door), so be aware it's pretty much a non-core to RC2 core port in the current state. I had a branch going for this, but to get Jil updated and run the performance tests I also need protobuf and we hit `coreclr.exe` crashing exceptions on that front. Theoretically, Jil is closer but so far away from an RC2 update on that branch it may be better to start over. Completing the set of libraries needed for testing is also non-trivial. I'm not sure if `SimplerSpeedTester` which Jil also uses is even planned for a port, that one also means changing all benchmark code which creates a conundrum of what comparisons are even valid. That'd have to be a cross-branch port for apples to apples ensuring no performance degradation.

TL;DR: I've done this PR twice now - don't have it in me to do it a third time. But, if someone else is willing to RC2-ify Jil then I'm happy to review it. If you want to close this out that's disappointing but may be best. It's a shame we never got a test run.

PR #38, opened 2015-12-14
Add missing dependency to project.json.

cesarbs:
cc @DamianEdwards @halter73


halter73:
:shipit:

cesarbs:
Should I merge to `master` too?

DamianEdwards:
Nah, from now on we'll only update `master` when we push framework releases, like the other repos.

PR #39, opened 2015-12-17
reuseStreams = true

benaadams:
"Amortize stream costs" #487

I think this is the correct config?


DamianEdwards:
I deleted the `hosting.json` file. If we want to do this we should just set it during configuration of the `WebHostBuilder`. Do you have measurements of the impact?

benaadams:
Superseded by #53

PR #42, opened 2015-12-25
OrdinalIgnoreCase is unnecessary

benaadams:
https://www.techempower.com/benchmarks/#section=code&hw=peak&test=json
> General requirements
3. Specific characters and character case matter.

### DIFF
--- a/src/Benchmarks/FortunesDapperMiddleware.cs
+++ b/src/Benchmarks/FortunesDapperMiddleware.cs
@@ -30,9 +30,7 @@ namespace Benchmarks
 
         public async Task Invoke(HttpContext httpContext, HtmlEncoder htmlEncoder)
         {
-            // We check Ordinal explicitly first because it's faster than OrdinalIgnoreCase
-            if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal) ||
-                httpContext.Request.Path.StartsWithSegments(_path, StringComparison.OrdinalIgnoreCase))
+            if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal))
             {
                 var rows = await LoadRows(_connectionString, _dbProviderFactory);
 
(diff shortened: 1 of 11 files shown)


DamianEdwards:
Technically, I believe they're referring to casing in responses, not URIs. The language around URIs is "The **recommended** URI is /json"

benaadams:
yeah but linux stuff is generally case sensitive so bbc news is linux; you can tell as if you go here its ok

`http://www.bbc.co.uk/news/scotland` but if you capitalize it 404s `http://www.bbc.co.uk/news/Scotland`

benaadams:
yeah but linux stuff is generally case sensitive so bbc news is linux; you can tell as if you go here its ok

`http://www.bbc.co.uk/news/scotland` but if you capitalize it 404s `http://www.bbc.co.uk/news/Scotland`

If you are being compared to linux frameworks on linux; should use the same rules? Windows is just generally more forgiving.

DamianEdwards:
Sure, but the other examples I tried in the TechEmpower benchmarks didn't care about casing AFAICT. All that said, we probably don't need to have this line anyway (everything here is controlled).

benaadams:
Ordinal has improves https://github.com/dotnet/coreclr/pull/2825

DamianEdwards:
Wow, that's cool. We'll get that benefit whether we remove the case-insensitive check or not though.

DamianEdwards:
Bah, OK you win, removing code is good :smile:

benaadams:
> We'll get that benefit whether we remove the case-insensitive check or not though.

Not if you fall through to next item

DamianEdwards:
Oh good point, although that's less of an issue with my latest changes (only enable specific scenarios).

benaadams:
And.... just for completeness https://tools.ietf.org/html/rfc7230#section-2.7.3
> The scheme and host are case-insensitive and normally provided in lowercase; all other
   components are compared in a case-sensitive manner.

However with controllers and methods being camel case and paths generally being lowercase (ignoring @shanselman's wild blog post paths); probably would want to be case insensitive for general aspnet path testing.

PR #44, opened 2015-12-29
Update Json.NET to 8.0.1

JamesNK:
Update Json.NET to 8.0.1
Add JsonArrayPool

### DIFF
--- a/src/Benchmarks/JsonMiddleware.cs
+++ b/src/Benchmarks/JsonMiddleware.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
 
 using System;
+using System.Buffers;
 using System.IO;
 using System.Text;
 using System.Threading.Tasks;
@@ -16,6 +17,7 @@ namespace Benchmarks
         private static readonly Task _done = Task.FromResult(0);
         private static readonly PathString _path = new PathString("/json");
         private static readonly JsonSerializer _json = new JsonSerializer();
+        private static readonly JsonArrayPool<char> _arrayPool = new JsonArrayPool<char>(ArrayPool<char>.Shared);
 
         private readonly RequestDelegate _next;
         
@@ -35,8 +37,11 @@ namespace Benchmarks
                 httpContext.Response.ContentLength = 30;
 
                 using (var sw = new StreamWriter(httpContext.Response.Body, Encoding.UTF8, bufferSize: 30))
+                using (var jsonWriter = new JsonTextWriter(sw))
                 {
-                    _json.Serialize(sw, new { message = "Hello, World!" });
+                    jsonWriter.ArrayPool = _arrayPool;
+
+                    _json.Serialize(jsonWriter, new { message = "Hello, World!" });
                 }
 
                 return _done;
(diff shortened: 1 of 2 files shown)


JamesNK:
I haven't run the benchmarks to see the impact of using an array pool. Unlike JsonTextReader, JsonTextWriter doesn't allocate a very large buffer by default - it's only about 70 bytes - so the overhead of getting and returning the array from the pool might not be worth it.

EamonNerbonne:
Anecdotally: in private code I use an array of `ConcurrentQueues` to manage an object pool, and the overhead of the array access and ConcurrentQueue update seems to be worth it even for trivial objects - as the the GC overhead of a single object costs more than a ConcurrentQueue dequeue/enqueue pair, even if it's tiny.

src/Benchmarks/JsonMiddleware.cs
```diff
@@ -53,4 +58,34 @@ public static IApplicationBuilder UseJson(this IApplicationBuilder builder)
             return builder.UseMiddleware<JsonMiddleware>();
         }
     }
+
+    public class JsonArrayPool<T> : IArrayPool<T>
```
DamianEdwards: Why does that need to be wrapped? i.e. why not just assign `ArrayPool<char>.Shared` directly to `jsonWriter.ArrayPool`?

DamianEdwards:
@JamesNK would you mind rebasing this please?

src/Benchmarks/JsonMiddleware.cs
JamesNK: Because Json.NET doesn't have a dependency on System.Buffers.

rynowak:
This should also be 8.0.2. 8.0.1 is [**over**](https://www.youtube.com/watch?v=YlGqN3AKOsA)

src/Benchmarks/JsonMiddleware.cs
DamianEdwards: Ah got it. `IArrayPool<T>` is in JSON.NET then, yes?

AspNetSmurfLab:
OK so the results are in... and it actually makes it worse. Note, the GC and allocated MB numbers come from the `GC` API, so take them with a grain of salt.

| Impl | RPS | GC Gen 0/1/2 | Allocated MB |
| ---- | ---- | --- | --- |
| Baseline (no pooling) | 214k | 29/3/1 | 16.6 |
| Baseline + allocation of unused `JsonTextWriter` | 210k | 30/3/1 | 164.74 |
| Baseline + allocation & use of `JsonTextWriter` w/o pooling | 207k | 28/3/1 | 127.8 |
| Baseline + allocation & use of `JsonTextWriter` with pooling | 204k | 27/3/1 | 142.94  |

I think the JSON payloads in this case are just too small (~30 bytes) to see any benefit. It's pretty clear from the allocated MB number that the cost of allocating the `JsonTextWriter` for each request far outweighs any benefits from pooling the char array.

What's more interesting is the overhead of using the pool vs. not when the `JsonTextWriter` is allocated (the last two results) is measurably worse in this case. i.e. setting the `JsonTextWriter.ArrayPool` property resulted in a perf decrease. I didn't expect that.

JamesNK:
Good to know.

The big buffer is on deserializing. JsonTextReader's buffer starts at char[2000], JsonTextWriter starts at char[32], only growing when it needs to when writing large strings with escaped characters.

benaadams:
Shared ArrayPool is 50 x 1MB arrays O_o

Might be better with something like:
```
private static readonly ArrayPool<char> _pool  = ArrayPool<char>.Create(2000, 64);
private static readonly JsonArrayPool<char> _arrayPool = new JsonArrayPool<char>(_pool);
```

JamesNK:
https://github.com/dotnet/corefx/blob/master/src/System.Buffers/src/System/Buffers/ArrayPool.cs#L21

yeah, that's pretty excesive

rynowak:
The `ArrayPool` uses bucketing by size, so the 'max' in this case is the maximum size that will use pooling. That comment also isn't quite right, it's the default max, not the default buffer size.

You generally get what you ask for.

benaadams:
> The ArrayPool uses bucketing by size, so the 'max' in this case is the maximum size that will use pooling. That comment also isn't quite right, it's the default max, not the default buffer size.

Ahh, looking deeper can see its doing this now. Looks like the docs are already being updated for clarity: https://github.com/dotnet/corefx/pull/5816/files

EamonNerbonne:
Even if it's not the pooling, it's odd that the memory usage grows by so much when allocating an unused `JsonTextWriter`.

PR #47, opened 2016-01-20
Use AsNoTracking instead of ChangeTracker

anpete:
… to prevent expensive initialization of ChangeTracker.

cc @DamianEdwards @halter73 @divega 

Related to: aspnet/EntityFramework#4133

### DIFF
--- a/src/Benchmarks/SingleQueryEfMiddleware.cs
+++ b/src/Benchmarks/SingleQueryEfMiddleware.cs
@@ -35,10 +35,8 @@ namespace Benchmarks
             {
                 var db = (ApplicationDbContext)httpContext.RequestServices.GetService(typeof(ApplicationDbContext));
 
-                db.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;
-
                 var id = _random.Next(1, 10001);
-                var row = await db.World.FirstAsync(w => w.Id == id);
+                var row = await db.World.AsNoTracking().FirstAsync(w => w.Id == id);
                 var result = JsonConvert.SerializeObject(row, _jsonSettings);
 
                 httpContext.Response.StatusCode = StatusCodes.Status200OK;
--- a/src/Benchmarks/Data/ApplicationDbSeeder.cs
+++ b/src/Benchmarks/Data/ApplicationDbSeeder.cs
@@ -8,8 +8,8 @@ namespace Benchmarks.Data
 {
     public class ApplicationDbSeeder
     {
-        private object _locker = new object();
-        private bool _seeded = false;
+        private readonly object _locker = new object();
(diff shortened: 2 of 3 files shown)


src/Benchmarks/SingleQueryEfMiddleware.cs
```diff
@@ -35,10 +35,8 @@ public SingleQueryEfMiddleware(RequestDelegate next)
             {
                 var db = (ApplicationDbContext)httpContext.RequestServices.GetService(typeof(ApplicationDbContext));
 
-                db.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;
-
                 var id = _random.Next(1, 10001);
-                var row = await db.World.FirstAsync(w => w.Id == id);
+                var row = await db.World.AsNoTracking().FirstAsync(w => w.Id == id);
```
rynowak: Would it be beneficial/allowed to cache the expression? We used codegen to do this in Razor because these can be deceptively expensive.

benaadams: Probably as its still doing the full ORM? (Test type 2: Single database query / Requirements) https://www.techempower.com/benchmarks/#section=code

anpete: @benaadams Yep, still full ORM. Could be worth measuring after this change.

DamianEdwards:
Do you have any numbers?

anpete:
@DamianEdwards I will try and get some. Everything is broken right now.

src/Benchmarks/SingleQueryEfMiddleware.cs
divega: I am not sure there is currently a way to have the query expression created only once and stashed in a static field because the DbContext instance is used to bootstrap the process. 

FWIW if we thought many customers would be ok with the ugly code required to stash the expression in a static field just to avoid the overhead of re-creating the expression on every request we could as well consider (re)adding the `CompiledQuery` feature to the backlog, since that would also improve the query translation lookup.

divega:
LGTM. 

@anpete I assume you also looked into splitting the initialization of the `ChangeTracker` to make setting the flag less expensive?

anpete:
@DamianEdwards I ran a micro-benchmark and the numbers look worse because it looks like we initialize ChangeTracker internally anyway :-/ If I can fix that then we can revisit this change.

@divega Not really. ChangeTracker just calls GetService to resolve an IChangeTrackerFactory, which causes a non-trivial amount of services to be resolved.

DamianEdwards:
@anpete yeah that's what I found originally which is why the code is the way it is :smile:. If you can fix that that'd be great.

PR #48, opened 2016-01-26
Allow configuration of scenarios

DamianEdwards:
Allows configuration of which scenarios to enable via interactive prompts in the console, cmd line args, or a `scenarios.json` file. This is to ensure scenarios don't suffer any performance impact from other bits in the app not required for that scenario.

@davidfowl

### DIFF
--- a/src/Benchmarks/Startup.cs
+++ b/src/Benchmarks/Startup.cs
@@ -2,11 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
 
 using System;
-using System.Collections.Generic;
 using System.Data.Common;
 using System.Data.SqlClient;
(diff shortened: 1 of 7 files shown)


src/Benchmarks/Startup.cs
```diff
@@ -16,28 +16,27 @@ namespace Benchmarks
...
-            {
-                // This allows passing of config values via the cmd line, e.g.: dnx web --app.EnableDbTests=true
-                builder.AddConfiguration("app.", env.Configuration);
-            }
-
             Configuration = builder.Build();
 
             Configuration.Bind(StartupOptions);
+
+            Scenarios = scenarios ?? new Scenarios();
```
davidfowl: Nuke

PR #51, opened 2016-01-27
Add graphs to readme.md

benaadams:
Adds graphs shown on https://github.com/benaadams/d3renderer pulls data
from this repo

Needs PR #50 else images will be broken as no data in repo

~~Change version path (e.g. v1) to cache bust github's image cache - may
need to update csv files in earlier commit~~

##### Plaintext
plaintext

##### JSON
json

##### Fortunes
frotunes

##### Single Query
singlequery

##### Multi Query
multiquery

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Benchmarks for ASP.NET Core including (but not limited to) scenarios from the [T
 ![multiquery](https://d3renderer.azurewebsites.net/multiquery/v1)
 
 ##### Fortunes
-![frotunes](https://d3renderer.azurewebsites.net/fortunes/v1)
+![fortunes](https://d3renderer.azurewebsites.net/fortunes/v1)
 
 # Running the benchmarks
 
--- a/README.md
+++ b/README.md
@@ -1,13 +1,20 @@
 # Benchmarks
 Benchmarks for ASP.NET Core including (but not limited to) scenarios from the [TechEmpower Web Framework Benchmarks](https://www.techempower.com/benchmarks/).
 
-
 ##### Plaintext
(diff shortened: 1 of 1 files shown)


benaadams:
@DamianEdwards any tweaks to graphs?

benaadams:
If the data is in this repo you can check what the graphs look like before merging at https://github.com/benaadams/d3renderer/blob/master/README-aspnet.md

simonmurdock:
Hello, sorry to butt in, but I thought I could provide some feedback...

Use a legend so there are fewer text overlapping issues
Display the value of the point on hover, as some are so far away from the Y-Axis It's hard to tell

benaadams:
@simonmurdock doesn't look like can add hover/tooltips - github only allows "images" in markdown -luckily svg counts as an image. I have added `<title />` element to the lines and points; however while the work on the raw svg alas with they don't seem to work when it is used as `` or they get lost in translation in github's "user content" image cache

simonmurdock:
I was hoping SVG would allow some fancyness like that. Shame that Github remove it. Maybe they're doing it on purpose for security etc.

README.md
```diff
@@ -1,6 +1,21 @@
...
+![json](https://d3renderer.azurewebsites.net/json/v1)
+
+##### Single Query
+![singlequery](https://d3renderer.azurewebsites.net/singlequery/v1)
+
+##### Multi Query
+![multiquery](https://d3renderer.azurewebsites.net/multiquery/v1)
+
+##### Fortunes
+![frotunes](https://d3renderer.azurewebsites.net/fortunes/v1)
```
Eilon: Typo in link name :smile:

PR #53, opened 2016-02-18
Add Header+Stream pooling

benaadams:
Headers were previously reused for every request on a connection; now they are created for each request. This override restores the previous behaviour (though slightly different - as it is cross connection and runs with a lower idling memory set)

Ref https://github.com/aspnet/KestrelHttpServer/pull/608

/cc @DamianEdwards

### DIFF
--- a/src/Benchmarks/appsettings.json
+++ b/src/Benchmarks/appsettings.json
@@ -1,4 +1,6 @@
 ﻿{
   "ConnectionString": "Server=(localdb)\\mssqllocaldb;Database=aspnet5-Benchmarks;Trusted_Connection=True;MultipleActiveResultSets=true",
-  "Scenarios": "Plaintext,Json"
+  "Scenarios": "Plaintext,Json",
+  "kestrel.maxPooledStreams": 256,
+  "kestrel.maxPooledHeaders": 256
 }


DamianEdwards:
@halter73 @mikeharder Can we take a look at this one please. This might be the reason we've seen a slight drop-off in the results on the SmurfLab.

mikeharder:
@DamianEdwards Sure, I will compare plaintext RPS before and after this change.

benaadams:
The change to "Scenarios": "Plaintext,Json",` is likely an oversight

mikeharder:
@benaadams The only change i see related to `"Scenarios": "Plaintext,Json"` is an added comma, which is necessary to make valid JSON.  Are you talking about the added comma or something else?

benaadams:
Heh, yeah was the comma - my brain is addled going from film shoot to film shoot :)

mikeharder:
@DamianEdwards @benaadams I'm not seeing a statistically significant change in RPS from this PR.  RPS was 980k for dev, and 988k for this PR, with a STDEV of 15k.  Is this PR expected to improve plaintext RPS?

benaadams:
It should revert a change which will reduce sustained perf (eg 5 mins), also you may have to set the numbers to be as high as the connections count

mikeharder:
@benaadams Is there a benchmark I can run to see the difference resulting from this change?

benaadams:
Is default now

PR #62, opened 2016-03-03
Convert to CLI.

NTaylorMullen:
/cc @DamianEdwards


src/Benchmarks/project.json
```diff
@@ -14,19 +15,18 @@
...
       "dependencies": {
         "NETStandard.Library": "1.0.0-*",
         "System.Runtime.Serialization.Primitives": "4.1.0-*"
       }
     },
-    "net451": {}
+    "net451": { }
+  },
+  "tools": {
+    "dotnet-ef": "1.0.0-*"
```
NTaylorMullen: This actually causes explosions currently because of our incoherent set of packages. Once the CI cycles through it'll start working.

DamianEdwards:
@NTaylorMullen were you actually able to run the EF migrations and successfully run one of the EF-related scenarios?

NTaylorMullen:
Talked offline and decided to work around blocking ef issues by doing the following:

- Create a separate Program.Main code path to do manual migrations. This would enable migrations to occur by invoking the entry point with a specific flag.
- Don't rely on EF for migrations, do it manually.

mikeharder:
@DamianEdwards Even if there is more work remaining for EF, it would be great to merge this now to update the build scripts.  The current build scripts download an old version of the CLI that no longer works with MVC.

src/Benchmarks/Startup.cs
```diff
@@ -160,6 +160,8 @@ public void Configure(IApplicationBuilder app, ApplicationDbSeeder dbSeeder)
 
             if (Scenarios.Any("Db"))
             {
+                dbContext.Database.EnsureCreated();
```
NTaylorMullen: So this is our work around for now to the broken ef migration tool. I'm only doing `EnsureCreated()` here but if we wanted to be extra sure the application boots with the correct Database I could drop the `EnsureDeleted()` prior to this. Given how infrequently the DB models change in this repo I felt that just ensuring its created is sufficient. That way startup time will be snappy :smile:.

Users can always delete the db manually if they want to.

NTaylorMullen:
:up: :date:

DamianEdwards:
:shipit:

Issue #66, opened 2016-04-04
Enable Server GC in project.json


mikeharder:
We should enable Server GC in project.json, once this feature is added to the CLI (https://github.com/dotnet/cli/issues/2121).

PR #69, opened 2016-04-06
Enable Server GC in project.json (#66)

mikeharder:
Requires newer CLI than currently used by our repos.  Verified to work on 1.0.0-beta-002252.

@halter73

### DIFF
--- a/src/Benchmarks/project.json
+++ b/src/Benchmarks/project.json
@@ -40,5 +40,10 @@
     "wwwroot",
     "node_modules",
     "bower_components"
-  ]
+  ],
+  "runtimeOptions": {
+    "configProperties": {
+      "System.GC.Server": true
+    }
+  }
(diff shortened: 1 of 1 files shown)


halter73:
This depends on https://github.com/dotnet/cli/pull/2227 which should be in our next dotnet cli drop. We should verify this actually works once we have it.

:shipit:

rynowak:
Is this needed to activate server GC? I was under the impression it was the default https://github.com/dotnet/cli/blob/rel/1.0.0/Documentation/specs/runtime-configuration-file.md#runtimeoptions-section-runtimeconfigjson

mikeharder:
The default is "false" as of CLI 1.0.0-beta-002252.  I assume it will stay this way and the spec needs to be updated, but you could check with the CLI team.

PR #70, opened 2016-04-07
Server, client, and driver apps to automate running benchmarks

mikeharder:
@halter73, @davidfowl


src/dotnet-benchmark-client/Controllers/JobsController.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            if (job == null || job.Id != 0 || job.State != ClientState.Waiting)
+            {
+                return BadRequest();
+            }
+
+            job = _jobs.Add(job);
+
+            Response.Headers["Location"] = $"/jobs/{job.Id}";
+            return new StatusCodeResult((int)HttpStatusCode.Accepted);
```
rynowak: FYI `CreatedAtAction("GetById", new { id = job.Id })` + changing the status code can do this for you as well

rynowak: We also have a `StatusCodes` static class with the status code constants defined as ints.

benaadams:
Will this become part of the CI? :+1:

src/dotnet-benchmark-client/Startup.cs
```diff
@@ -0,0 +1,205 @@
...
+            process.ErrorDataReceived += (_, e) =>
+            {
+                Log(e.Data);
+                job.Error += (e.Data + Environment.NewLine);
+            };
+
+            process.Exited += (_, __) =>
+            {
+                double rps = -1;
+                var match = Regex.Match(job.Output, @"Requests/sec:\s*([\d.]*)");
```
rynowak: Allocations on this are going to be :banana:s - did you learn nothing from my talk? :trollface:

src/dotnet-benchmark-client/project.json
```diff
@@ -0,0 +1,36 @@
...
+    "warningsAsErrors": true,
+    "emitEntryPoint": true,
+    "preserveCompilationContext": true
+  },
+  "dependencies": {
+    "Microsoft.AspNetCore.Mvc": "1.0.0-*",
+    "Microsoft.AspNetCore.Server.Kestrel": "1.0.0-*",
+    "Microsoft.Extensions.CommandLineUtils": "1.0.0-*",
+    "System.Diagnostics.Process": "4.1.0-*",
+    "Newtonsoft.Json": "8.0.*",
```
rynowak: If you're referencing MVC you shouldn't really need to reference this unless you explicitly want to float on the patch version

src/dotnet-benchmark-client/Startup.cs
davidfowl: omg regex.

src/dotnet-benchmark-client/project.json
davidfowl: Yea lets not do this for JSON.NET. Just pick a version we don't need to float.

davidfowl:
> Will this become part of the CI?

Yah

src/dotnet-benchmark-client/project.json
mikeharder: Assuming packages correctly follow semver, why not always float?  For that matter, why not float up to the minor version (e.g. "8.*")?

src/dotnet-benchmark-client/Controllers/JobsController.cs
mikeharder: I tried using `CreatedAtAction("GetById", new { id = job.Id })`, but it fails with the error:


```
System.InvalidOperationException: No route matches the supplied values.
   at Microsoft.AspNetCore.Mvc.CreatedAtActionResult.OnFormatting(ActionContext context)
```

rynowak: Ah.... yeah I advised you wrongly sorry. `CreatedAtAction("GetById", new { id = job.Id })` is going to interpret the anonymous object as the payload, not as route values. https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs#L887

`CreatedAtAction("GetById", null, new { id = job.Id })`  would do the trick, but it's kinda odd to pass in a null payload.

Anyways, if you feel like removing the hardcoding the URL, you should also be able to just do `Url.Action("GetById", new { id = job.Id })`. I don't really think this is important, just wanted to make mention

src/Benchmarks.ServerJob/project.json
```diff
@@ -0,0 +1,26 @@
+{
+  "version": "1.0.0-*",
+  "description": "Class to transfer data to/from the benchmark server.",
```
halter73: I think instead of dealing with the boilerplate to create a project for the ServerJob/ClientJob transfer classes, it would be easier just to have some shared .cs file [like this](https://github.com/aspnet/EntityFramework/blob/c28771c3244a7c19d42fb58d500c3b455942ac71/src/Microsoft.EntityFrameworkCore/project.json#L19).

src/Benchmarks.ClientJob/project.json
```diff
@@ -0,0 +1,26 @@
+{
+  "version": "1.0.0-*",
+  "description": "Class to transfer data to/from the benchmark client.",
+  "repository": {
```
halter73: Note: version/description/repository sections aren't really necessary for test/sample/tool projects that we aren't creating packages for. Doesn't hurt though.

src/Repository/IRepository.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Repository
+{
+    public interface IRepository<T> where T: IIdentifiable
```
halter73: Why not just use IDictonary/Dictionary?

src/BenchmarksServer/GitCommands.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace BenchmarkServer
+{
+    public static class Git
```
halter73: Nit: File name should match class name.

src/BenchmarksDriver/Program.cs
```diff
@@ -0,0 +1,319 @@
...
+namespace BenchmarkDriver
+{
+    public class Program
+    {
+        private static readonly HttpClient _httpClient = new HttpClient();
+
+        private static readonly Dictionary<Scenario, ClientJob> _clientJobs =
+            new Dictionary<Scenario, ClientJob>()
+            {
+                { Scenario.Plaintext, new ClientJob() { Connections = 256, Threads = 32, Duration = 10, PipelineDepth = 16} },
```
halter73: Can we make this configurable via driver command line args?

src/Benchmarks.ServerJob/project.json
mikeharder: A separate library seems to more correctly express the relationship between components, so I think it should always be chosen over shared sources, unless there's a strong reason to do otherwise.

davidfowl:
This PR has surpassed the reviewable state. I'd suggest merging and sending smaller PRs for changes and review feedback

src/Repository/IRepository.cs
mikeharder: The main reason is that IRepository expresses exactly the contract my application needs, and methods can be added/removed as needed.  Using a framework interface like IDictionary removes this flexibility.

Another reason is that IRepository is an ordered list, while IDictionary is unordered.

src/BenchmarksServer/GitCommands.cs
mikeharder: Fixed.

src/BenchmarksDriver/Program.cs
mikeharder: It's easy to add this feature, but I'd rather wait until someone actually wants to use it.  If you do  have a real use for it now, I can add it.

I would still want to keep the defaults hardcoded for each scenario, to make it easier for everyone to use the same parameters unless they explicitly want something different.

mikeharder:
@davidfowl: Sounds good to me, the next PR will be fairly small (adding support for arbitrary repos).

halter73:
:shipit:

PR #71, opened 2016-04-09
Support arbitrary repositories and branches or commits

mikeharder:
@halter73 @davidfowl 

Example:
```
dotnet run -- -n plaintext -s http://adx-3u-perfsvr:5001 -c http://adx-3u-svr02:5002
  -o KestrelHttpServer:user/test-branch -o HttpAbstractions:user/test-branch
```

### DIFF
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -47,7 +47,7 @@ namespace BenchmarkDriver
                 "Connection string of SQL Database to store results", CommandOptionType.SingleValue);
 
             var sourceOption = app.Option("-o|--source",
-                "Source dependency. Format is 'repo:branch'. " +
+                "Source dependency. Format is 'repo@branchOrCommit'. " +
                 "Repo can be a full URL, or a short name under https://github.com/aspnet.",
                 CommandOptionType.MultipleValue);
 
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -73,16 +73,16 @@ namespace BenchmarkDriver
                     var sources = new List<Source>();
                     foreach (var source in sourceOption.Values)
                     {
-                        var split = source.LastIndexOf(':');
-                        var repository = source.Substring(0, split);
-                        var branch = source.Substring(split + 1);
+                        var split = source.IndexOf('@');
+                        var repository = (split == -1) ? source : source.Substring(0, split);
+                        var branch = (split == -1) ? null : source.Substring(split + 1);
 
                         if (!repository.Contains(":"))
                         {
                             repository = $"https://github.com/aspnet/{repository}.git";
                         }
 
-                        sources.Add(new Source() { Branch = branch, Repository = repository });
+                        sources.Add(new Source() { BranchOrCommit = branch, Repository = repository });
                     }
                     serverJob.Sources = sources;
 
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -47,7 +47,7 @@ namespace BenchmarkDriver
                 "Connection string of SQL Database to store results", CommandOptionType.SingleValue);
 
             var sourceOption = app.Option("-o|--source",
-                "Source depencency. Format is 'repo:branch'. " +
+                "Source dependency. Format is 'repo:branch'. " +
                 "Repo can be a full URL, or a short name under https://github.com/aspnet.",
                 CommandOptionType.MultipleValue);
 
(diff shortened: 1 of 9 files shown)


BrennanConroy:
This doesn't look like it supports arbitrary commits, wouldn't that be something we would want?

mikeharder:
@BrennanConroy This is a good suggestion, I will add this in a subsequent PR.  For syntax to the BenchmarkDriver app, I was thinking of using "repo:branch#commit", like so:

```
-o KestrelHttpServer:user/test-branch#a1b2c3d
```

rynowak:
Why not `@` - that's what github does

mikeharder:
OK, then:

```
-o KestrelHttpServer:user/test-branch@a1b2c3d
```

mikeharder:
I chatted with @halter73, we have improved the design.  In the driver, you can specify either a branch or a commit after the '@' character:

```
-o KestrelHttpServer@user/test-branch
-o KestrelHttpServer@a1b2c3d
```

The code can handle these in exactly the same way:

```
git clone repo
git checkout branch|commit
```

I will update this PR with these changes.

src/BenchmarksServer/Startup.cs
```diff
@@ -174,36 +169,46 @@ public static int Main(string[] args)
...
+            {
+                globalJson["projects"].Add(Path.Combine("..", dir, "src"));
+            }
+            File.WriteAllText(globalJsonPath, JsonConvert.SerializeObject(globalJson, Formatting.Indented));
 
-                // Add references to libuv (required until https://github.com/aspnet/KestrelHttpServer/pull/731)
-                var benchmarksProjectJson = Path.Combine(path, _benchmarksDir, "src", "Benchmarks", "project.json");
-                dynamic projectJson = JsonConvert.DeserializeObject(File.ReadAllText(benchmarksProjectJson));
+            // If any dir named "KestrelHttpServer", add libuv reference to Benchmarks
+            // (required until https://github.com/aspnet/KestrelHttpServer/pull/731)
```
halter73: No longer required: https://github.com/aspnet/KestrelHttpServer/pull/731

src/BenchmarksServer/Startup.cs
```diff
@@ -174,36 +169,46 @@ public static int Main(string[] args)
             }
         }
 
-        private static void CloneAndRestore(string path, ServerJob job)
+        private static string CloneAndRestore(string path, ServerJob job)
         {
-            var benchmarksRepoUrl = string.IsNullOrEmpty(job.BenchmarksRepo) ? _defaultBenchmarksRepoUrl : job.BenchmarksRepo;
-            var benchmarksBranch = string.IsNullOrEmpty(job.BenchmarksBranch) ? _defaultBranch : job.BenchmarksBranch;
-            Git.Clone(path, benchmarksRepoUrl, benchmarksBranch, _benchmarksDir);
+            var benchmarksDir = Git.Clone(path, _benchmarksRepoUrl);
```
halter73: Might it be simpler just to specify some random clone directory than read from stderr?

src/BenchmarksDriver/Program.cs
```diff
@@ -74,12 +68,24 @@ public static int Main(string[] args)
...
-                        BenchmarksBranch = benchmarksBranchOption.Value(),
-                        BenchmarksRepo = benchmarksRepoOption.Value(),
-                        KestrelBranch = kestrelBranchOption.Value(),
-                        KestrelRepo = kestrelRepoOption.Value(),
                     };
 
+                    var sources = new List<Source>();
+                    foreach (var source in sourceOption.Values)
+                    {
+                        var split = source.LastIndexOf(':');
```
halter73: Use '@' as discussed.

src/BenchmarksServer/Startup.cs
mikeharder: Will remove in next version of PR.

src/BenchmarksServer/Startup.cs
mikeharder: It's nice to have meaningful directory names if you need to debug something, and the code to read from stderr is already written so it's easier to leave it at this point.

src/BenchmarksDriver/Program.cs
```diff
@@ -46,16 +46,10 @@ public static int Main(string[] args)
...
-            var benchmarksRepoOption = app.Option("--benchmarksRepo",
-                "URL of Benchmarks repo.  Default is 'https://github.com/aspnet/benchmarks.git'.", CommandOptionType.SingleValue);
-
-            var kestrelBranchOption = app.Option("--kestrelBranch",
-                "Kestrel branch.  If specified, Benchmarks is configured to use Kestrel from sources rather than packages.",
-                CommandOptionType.SingleValue);
-            var kestrelRepoOption = app.Option("--kestrelRepo",
-                "URL of Kestrel repo.  Default is 'https://github.com/aspnet/KestrelHttpServer.git'.", CommandOptionType.SingleValue);
+            var sourceOption = app.Option("-o|--source",
+                "Source dependency. Format is 'repo:branch'. " +
```
halter73: Update help text to use '@'.

halter73:
:shipit: after help text update.

PR #73, opened 2016-04-11
allow benchmarks to run within techempower automation on linux

nathana1:
2 fixes for running in the techempower suite:
- Console.ReadKey throws when the process is run in the background, so allow interactivity to be toggled off (remains on by default)
- the python JSON parser can't read responses that begin with a BOM

### DIFF
--- a/src/Benchmarks/Middleware/JsonMiddleware.cs
+++ b/src/Benchmarks/Middleware/JsonMiddleware.cs
@@ -8,18 +8,16 @@ using System.Threading.Tasks;
 using Benchmarks.Configuration;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.WebUtilities;
 using Newtonsoft.Json;
 
 namespace Benchmarks.Middleware
 {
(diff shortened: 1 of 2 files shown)


nathana1:
Add some folks:
@halter73 @SajayAntony @DamianEdwards @sivagms

src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -31,9 +34,9 @@ public Task Invoke(HttpContext httpContext)
             {
                 httpContext.Response.StatusCode = 200;
                 httpContext.Response.ContentType = "application/json";
-                httpContext.Response.ContentLength = 30;
+                httpContext.Response.ContentLength = _contentLength;
 
-                using (var sw = new StreamWriter(httpContext.Response.Body, Encoding.UTF8, bufferSize: 30))
+                using (var sw = new StreamWriter(httpContext.Response.Body, _encoding, bufferSize: _contentLength))
```
rynowak: If you use `HttpResponseStreamWriter` it will handle this for you

nathana1:
New improved version coming soon.

PR #74, opened 2016-04-11
Remove web.config

JunTaoLuo:
I came across this repo as part of cleaning up and relocating `web.config`. Do we plan to run benchmarks with kestrel behind IIS or IISExpress? If we do, we need to call `UseIISIntegration()` in `Startup`. If we don't, we should just remove the `web.config` unless there is another reason to keep it. @DamianEdwards @halter73 @muratg


JunTaoLuo:
Chatted offline with @halter73 and @mikeharder. The decision is to remove this until there is a scenario where we want to test behind IIS.

PR #75, opened 2016-04-12
Run client job once as a warmup, to reduce variance and make the resu…

mikeharder:
…lts more representative of real-world performance.

The diff looks larger than the actual change.  Most of the changes is just extracting a RunClientJob() method from the existing Run() method.

@halter73

### DIFF
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Data.SqlClient;
+using System.Diagnostics;
 using System.Net.Http;
 using System.Text;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


src/BenchmarksDriver/Program.cs
```diff
@@ -141,103 +142,113 @@ public static int Main(string[] args)
...
+                Log("Warmup");
+                await RunClientJob(scenario, clientUri, serverBenchmarkUri);
 
-                    var clientJobsUri = new Uri(clientUri, "/jobs");
+                Log("Benchmark");
+                var clientJob = await RunClientJob(scenario, clientUri, serverBenchmarkUri);
 
-                    var clientJob = new ClientJob(_clientJobs[scenario]) { ServerBenchmarkUri = serverBenchmarkUri };
-                    var clientContent = JsonConvert.SerializeObject(clientJob);
+                Debug.Assert(clientJob.State == ClientState.Completed);
```
halter73: Make `Trace.Assert`

halter73:
Ship after fix. I like the plot you made.

src/BenchmarksDriver/Program.cs
mikeharder: Why use Trace.Assert() instead of Debug.Assert()?  AFAIK, the only difference is the former works in both Release and Debug, which doesn't really matter since this is test driver code, but I've always used Debug.Assert() to avoid slowing down Release builds.

PR #77, opened 2016-04-22
Adding MVC scenarios for automation

rynowak:
/cc @mikeharder

### DIFF
--- a/src/Benchmarks.ServerJob/Scenario.cs
+++ b/src/Benchmarks.ServerJob/Scenario.cs
@@ -3,11 +3,21 @@
 
 namespace Benchmarks.ServerJob
 {
+    /// <summary>
+    /// The scenario is used both as a commandline argument to the Benchmarks process and also to compute
+    /// the path portion of the URL. These definitions should match those in the Benchmarks process.
+    /// 
+    /// Use [ScenarioPath(...)] when the path portion of the URL doesn't match the scenario name exactly.
+    /// </summary>
     public enum Scenario
     {
         Plaintext,
         Json,
+
+        [ScenarioPath("/mvc/plaintext")]
         MvcPlaintext,
+
+        [ScenarioPath("/mvc/json")]
         MvcJson,
     }
 }
--- a/src/Benchmarks.ServerJob/Scenario.cs
+++ b/src/Benchmarks.ServerJob/Scenario.cs
@@ -6,6 +6,8 @@ namespace Benchmarks.ServerJob
     public enum Scenario
     {
         Plaintext,
-        Json
+        Json,
+        MvcPlaintext,
+        MvcJson,
     }
 }
--- a/src/Benchmarks/Benchmarks.xproj
+++ b/src/Benchmarks/Benchmarks.xproj
@@ -9,7 +9,7 @@
     <ProjectGuid>ec62acf4-8b19-41c2-b699-d75cab7763df</ProjectGuid>
     <RootNamespace>Benchmarks</RootNamespace>
     <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">..\..\artifacts\obj\$(MSBuildProjectName)</BaseIntermediateOutputPath>
-    <OutputPath Condition="'$(OutputPath)'=='' ">..\..\artifacts\bin\$(MSBuildProjectName)\</OutputPath>
+    <OutputPath Condition="'$(OutputPath)'=='' ">..\..\artifacts\</OutputPath>
   </PropertyGroup>
   <PropertyGroup>
     <SchemaVersion>2.0</SchemaVersion>
--- a/src/Benchmarks/Configuration/Scenarios.cs
+++ b/src/Benchmarks/Configuration/Scenarios.cs
@@ -53,8 +53,11 @@ namespace Benchmarks.Configuration
         [ScenarioPath("/fortunes/dapper")]
         public bool DbFortunesDapper { get; set; }
 
-        [ScenarioPath("/mvc/plaintext", "/mvc/json")]
-        public bool MvcApis { get; set; }
+        [ScenarioPath("/mvc/plaintext")]
+        public bool MvcPlaintext { get; set; }
+
+        [ScenarioPath("/mvc/json")]
+        public bool MvcJson { get; set; }
 
         [ScenarioPath("/mvc/view")]
         public bool MvcViews { get; set; }
--- a/src/Benchmarks/Startup.cs
+++ b/src/Benchmarks/Startup.cs
@@ -84,7 +84,7 @@ namespace Benchmarks
                     //.AddApplicationPart(typeof(Startup).GetTypeInfo().Assembly)
                     .AddControllersAsServices();
 
-                if (Scenarios.MvcApis)
+                if (Scenarios.MvcJson)
                 {
                     mvcBuilder.AddJsonFormatters();
                 }
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -23,8 +23,10 @@ namespace BenchmarkDriver
         private static readonly Dictionary<Scenario, ClientJob> _clientJobs =
             new Dictionary<Scenario, ClientJob>()
             {
-                { Scenario.Plaintext, new ClientJob() { Connections = 256, Threads = 32, Duration = 10, PipelineDepth = 16} },
-                { Scenario.Json, new ClientJob() { Connections = 256, Threads = 32, Duration = 10} },
+                { Scenario.Plaintext, new ClientJob() { Connections = 256, Threads = 32, Duration = 10, PipelineDepth = 16 } },
+                { Scenario.Json, new ClientJob() { Connections = 256, Threads = 32, Duration = 10 } },
+                { Scenario.MvcPlaintext, new ClientJob() { Connections = 256, Threads = 32, Duration = 10, PipelineDepth = 16 } },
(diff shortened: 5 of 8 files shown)


rynowak:
Currently verifying this locally

src/Benchmarks/Benchmarks.xproj
```diff
@@ -9,7 +9,7 @@
     <ProjectGuid>ec62acf4-8b19-41c2-b699-d75cab7763df</ProjectGuid>
     <RootNamespace>Benchmarks</RootNamespace>
     <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">..\..\artifacts\obj\$(MSBuildProjectName)</BaseIntermediateOutputPath>
-    <OutputPath Condition="'$(OutputPath)'=='' ">..\..\artifacts\bin\$(MSBuildProjectName)\</OutputPath>
+    <OutputPath Condition="'$(OutputPath)'=='' ">..\..\artifacts\</OutputPath>
   </PropertyGroup>
```
rynowak: VS wanted to make this change, so I figured why not. This output path was generated as part of an RC1 xproj, and won't work with dotnet CLI + p2p references which is why VS changed it.

src/Benchmarks/Configuration/Scenarios.cs
```diff
@@ -53,8 +53,11 @@ public Scenarios(IScenariosConfiguration scenariosConfiguration)
         [ScenarioPath("/fortunes/dapper")]
         public bool DbFortunesDapper { get; set; }
 
-        [ScenarioPath("/mvc/plaintext", "/mvc/json")]
-        public bool MvcApis { get; set; }
+        [ScenarioPath("/mvc/plaintext")]
+        public bool MvcPlaintext { get; set; }
+
+        [ScenarioPath("/mvc/json")]
+        public bool MvcJson { get; set; }
```
rynowak: Needs to be split so we can track these as separate scenarios

src/Benchmarks/Startup.cs
```diff
@@ -84,7 +84,7 @@ public void ConfigureServices(IServiceCollection services)
                     //.AddApplicationPart(typeof(Startup).GetTypeInfo().Assembly)
                     .AddControllersAsServices();
 
-                if (Scenarios.MvcApis)
+                if (Scenarios.MvcJson)
                 {
                     mvcBuilder.AddJsonFormatters();
```
rynowak: Not needed for plaintext

davidfowl:
:shipit:

src/BenchmarksDriver/Program.cs
```diff
@@ -23,8 +23,10 @@ public class Program
         private static readonly Dictionary<Scenario, ClientJob> _clientJobs =
             new Dictionary<Scenario, ClientJob>()
             {
-                { Scenario.Plaintext, new ClientJob() { Connections = 256, Threads = 32, Duration = 10, PipelineDepth = 16} },
-                { Scenario.Json, new ClientJob() { Connections = 256, Threads = 32, Duration = 10} },
+                { Scenario.Plaintext, new ClientJob() { Connections = 256, Threads = 32, Duration = 10, PipelineDepth = 16 } },
+                { Scenario.Json, new ClientJob() { Connections = 256, Threads = 32, Duration = 10 } },
+                { Scenario.MvcPlaintext, new ClientJob() { Connections = 256, Threads = 32, Duration = 10, PipelineDepth = 16 } },
+                { Scenario.MvcJson, new ClientJob() { Connections = 256, Threads = 32, Duration = 10} },
```
rynowak: I haven't done any private testing yet with MVC JSON, but the settings for MVC plaintext are (in my experience) the best.

src/Benchmarks.ServerJob/Scenario.cs
```diff
@@ -3,9 +3,21 @@
...
     {
         Plaintext,
-        Json
+        Json,
+
+        [ScenarioPath("/mvc/plaintext")]
+        MvcPlaintext,
+
+        [ScenarioPath("/mvc/json")]
+        MvcJson,
```
rynowak: @mikeharder - needed to add this (or something like it) to be able to work with multiple path segments. If you don't like this approach let's discuss

src/BenchmarksServer/Startup.cs
```diff
@@ -171,20 +172,34 @@ public static int Main(string[] args)
...
         private static string CloneAndRestore(string path, ServerJob job)
         {
-            var benchmarksDir = Git.Clone(path, _benchmarksRepoUrl);
+            // It's possible that the user specified a custom branch/commit for the benchmarks repo,
+            // so we need to add that to the set of sources to restore if it's not already there.
+            //
+            // Note that this is also going to de-dupe the repos if the same one was specified twice at
+            // the command-line (last first to support overrides).
+            var repos = new HashSet<Source>(job.Sources.Reverse(), SourceRepoComparer.Instance);
+            repos.Add(new Source() { Repository = _benchmarksRepoUrl });
```
rynowak: Funnily enough to be able to test changes locally you need to pass a custom commitish for aspnet/Benchmarks, which this didn't support.'

I'm realizing now that the `.Reverse()` probably belongs in the commandline driver code and not here. I'll remove it.

rynowak:
Updated. I had add support for a few things to the server to get this running, but it's working now end to end.

halter73:
:ship:

src/BenchmarksServer/Startup.cs
mikeharder: I would add a comment that `repos.Add(new Source() { Repository = _benchmarksRepoUrl });` will be a no-op if the HashSet already contains a benchmarks repo.  I didn't know HashSet.Add() worked this way before checking the docs.

src/BenchmarksServer/Startup.cs
```diff
@@ -171,20 +172,34 @@ public static int Main(string[] args)
...
 
             // Clone
+            string benchmarksDir = null;
             var dirs = new List<string>();
             foreach (var source in job.Sources)
             {
                 var dir = Git.Clone(path, source.Repository);
+                if (string.Equals(source.Repository, _benchmarksRepoUrl, StringComparison.OrdinalIgnoreCase))
```
mikeharder: I think we should use a weaker check for repository equality, to support (for example) specifying a fork of benchmarks like `http://github.com/foo/benchmarks.git`.  We could use the text between the last slash character and the next dot character (if it exists).  Because I think the trailing `.git` is optional, and we'd want `aspnet/benchmarks.git` to match `aspnet/benchmarks`.

src/BenchmarksServer/Startup.cs
```diff
@@ -194,8 +209,8 @@ private static string CloneAndRestore(string path, ServerJob job)
             }
```
mikeharder: I would exclude the benchmarks dir itself from this foreach loop, since we don't need to add `benchmarks\src` to its own global.json.

src/BenchmarksServer/Startup.cs
```diff
@@ -266,5 +281,35 @@ private static Process StartProcess(string hostname, string benchmarksRepo, Serv
...
+            return $"http://{hostname}:5000/{path.ToLower()}";
+        }
+
+        private class SourceRepoComparer : IEqualityComparer<Source>
+        {
+            public static readonly SourceRepoComparer Instance = new SourceRepoComparer();
+
+            public bool Equals(Source x, Source y)
+            {
+                return string.Equals(x.Repository, y.Repository);
```
mikeharder: As mentioned above, I would use a weaker equality check that just compares the repo name (after last slash and before last dot).

src/BenchmarksServer/Startup.cs
```diff
@@ -266,5 +281,35 @@ private static Process StartProcess(string hostname, string benchmarksRepo, Serv
...
+            }
+
+            public int GetHashCode(Source obj)
+            {
+                return StringComparer.OrdinalIgnoreCase.GetHashCode(obj.Repository);
```
mikeharder: Update `GetHashCode()` to match changes to `Equals()`.

src/BenchmarksServer/Startup.cs
rynowak: sure

src/BenchmarksServer/Startup.cs
rynowak: Yeah seems like a good idea

src/BenchmarksServer/Startup.cs
rynowak: done

rynowak:
@mikeharder updated to address feedback. Tested locally with:
```
dotnet run -- -n Json -s http://rynowak-lefty:5001/ -c http://rynowak-goat-01:5002/
dotnet run -- -n MvcJson -s http://rynowak-lefty:5001/ -c http://rynowak-goat-01:5002/ -o Benchmarks@mvc-scenarios
dotnet run -- -n MvcJson -s http://rynowak-lefty:5001/ -c http://rynowak-goat-01:5002/ -o http://github.com/aspnet/Benchmarks.git@mvc-scenarios
dotnet run -- -n MvcJson -s http://rynowak-lefty:5001/ -c http://rynowak-goat-01:5002/ -o http://github.com/aspnet/Benchmarks@mvc-scenarios
```

mikeharder:
:shipit: for the existing code.  Did you want to update makefile.shade in this PR as well?

rynowak:
Yeah, thanks for the reminder, just did it.

mikeharder:
:shipit:

rynowak:
3855812385530ffefb85eedbca26e34178ba2e8c

PR #78, opened 2016-04-26
Make stuff portable

BrennanConroy:
@mikeharder @pakrym

### DIFF
--- a/src/Benchmarks.ClientJob/project.json
+++ b/src/Benchmarks.ClientJob/project.json
@@ -13,12 +13,9 @@
     "Newtonsoft.Json": "8.0.3"
   },
   "frameworks": {
-    "netcoreapp1.0": {
+    "netstandard1.0": {
       "dependencies": {
-        "Microsoft.NETCore.App": {
-          "version": "1.0.0-*",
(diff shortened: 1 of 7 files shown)


src/Benchmarks.ClientJob/project.json
```diff
@@ -13,9 +13,12 @@
     "Newtonsoft.Json": "8.0.3"
   },
   "frameworks": {
-    "netstandardapp1.5": {
```
rynowak: Aren't `ClientJob`, `ServerJob` and `Repository` class libraries?

BrennanConroy: Maybe, then they should have been targeting netstandard1.x instead of netstandardapp

mikeharder: They are supposed to be class libraries, I authored the project.json wrong.  Can you change them now?

We should also add support for NetFx 4.5, I can add this in a later change.

BrennanConroy: Netstandard1.0 :)

mikeharder:
:shipit:

PR #79, opened 2016-04-26
Remove "--infer-runtimes" from call to "dotnet restore"

mikeharder:
- No longer needed since benchmarks and all dependencies use netstandard or netcoreapp

@BrennanConroy, @rynowak, @JunTaoLuo

### DIFF
--- a/src/BenchmarksServer/Startup.cs
+++ b/src/BenchmarksServer/Startup.cs
@@ -221,7 +221,7 @@ namespace BenchmarkServer
             // Restore in each dir
             foreach (var dir in dirs)
             {
-                ProcessUtil.Run("dotnet", "restore --infer-runtimes", workingDirectory: Path.Combine(path, dir));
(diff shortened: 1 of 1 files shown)


rynowak:
:heart_eyes_cat: - this will hopefully unblock using private builds of MVC

rynowak:
:shipit:

mikeharder:
@rynowak: While making this change, I noticed that `dotnet restore` takes quite a while when specifying a source repo like `Mvc`.  Would it be sufficient to run `dotnet restore` in `Repo\src` rather than `Repo\`?  If so, this should improve build times quite a bit.

rynowak:
That would definitely help as long as we only need to build repos that have a `src` (all of our repos follow this rule). MVC has like 14 projects in src and like 90 total

mikeharder:
OK, I'll create a separate PR with that change since it's a bit more risky.

PR #80, opened 2016-04-26
Run "dotnet restore" in "repo\src" rather than "repo\"

mikeharder:
- Reduces restore time of Mvc by 30s

@rynowak


muratg:
:+1: :shipit:

PR #81, opened 2016-04-27
Use static to get base path

rynowak:
FYI only - will merge to unblock perf CI

### DIFF
--- a/src/Benchmarks/Startup.cs
+++ b/src/Benchmarks/Startup.cs
@@ -18,11 +18,11 @@ namespace Benchmarks
 {
     public class Startup
     {
-        public Startup(IApplicationEnvironment appEnv, IHostingEnvironment hostingEnv, Scenarios scenarios)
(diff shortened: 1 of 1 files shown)


src/Benchmarks/Startup.cs
```diff
@@ -18,11 +18,11 @@ namespace Benchmarks
 {
     public class Startup
     {
-        public Startup(IApplicationEnvironment appEnv, IHostingEnvironment hostingEnv, Scenarios scenarios)
+        public Startup(IHostingEnvironment hostingEnv, Scenarios scenarios)
         {
             // Set up configuration sources.
             var builder = new ConfigurationBuilder()
-                .SetBasePath(appEnv.ApplicationBasePath)
+                .SetBasePath(PlatformServices.Default.Application.ApplicationBasePath)
```
rynowak: /cc @pranavkm

mikeharder:
Was this a reaction to a breaking change?

PR #84, opened 2016-05-04
Only resolve response feature once

benaadams:


src/Benchmarks/Middleware/PlaintextMiddleware.cs
```diff
@@ -26,12 +26,13 @@ public Task Invoke(HttpContext httpContext)
...
-                httpContext.Response.ContentType = "text/plain";
+                var response = httpContext.Response;
+                response.StatusCode = 200;
+                response.ContentType = "text/plain";
                 // HACK: Setting the Content-Length header manually avoids the cost of serializing the int to a string.
                 //       This is instead of: httpContext.Response.ContentLength = _helloWorldPayload.Length;
-                httpContext.Response.Headers["Content-Length"] = "13";
-                return httpContext.Response.Body.WriteAsync(_helloWorldPayload, 0, _helloWorldPayload.Length);
+                response.Headers["Content-Length"] = "13";
+                return response.Body.WriteAsync(_helloWorldPayload, 0, _helloWorldPayload.Length);
```
rynowak: Isn't this going through the `HttpResponse` not the `IHttpResponseFeature`? I think there would be more we could shave off using the feature.

benaadams: Ah yeah, it still goes though the feature anyway

feature

rynowak:
Why close it though? What would the numbers look like if we retrieved the feature and used that directly?

This is something that I've been diving into with MVC as well, and it seems like this area is a bottleneck we need to solve.

benaadams:
Added more info to the other PR

PR #88, opened 2016-05-16
React to rename of Microsoft.EntityFrameworkCore.Commands and dotnet-ef

mikeharder:
@natemcmaster

### DIFF
--- a/src/Benchmarks/project.json
+++ b/src/Benchmarks/project.json
@@ -6,7 +6,6 @@
   },
   "dependencies": {
     "Dapper": "1.50.0-*",
-    "Microsoft.EntityFrameworkCore.Commands": "1.0.0-*",
     "Microsoft.EntityFrameworkCore.SqlServer": "1.0.0-*",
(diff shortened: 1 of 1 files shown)


src/Benchmarks/project.json
```diff
@@ -6,7 +6,6 @@
   },
   "dependencies": {
     "Dapper": "1.50.0-*",
-    "Microsoft.EntityFrameworkCore.Commands": "1.0.0-*",
```
natemcmaster: You also need " "Microsoft.EntityFrameworkCore.Tools": "1.0.0-*" in dependencies.

https://docs.efproject.net/en/latest/cli/dotnet.html#installation

NTaylorMullen: And as a build-time only dependency.

mikeharder:
@natemcmaster, @NTaylorMullen: Updated

natemcmaster:
:shipit:

PR #91, opened 2016-05-24
UseServer(string) was removed

rynowak:
Changing to call kestrel directly for now to unblock perf runs

### DIFF
--- a/src/Benchmarks/Program.cs
+++ b/src/Benchmarks/Program.cs
@@ -34,8 +34,7 @@ namespace Benchmarks
                 .Build();
 
             var webHostBuilder = new WebHostBuilder()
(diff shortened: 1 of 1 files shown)


rynowak:
Going to merge this to unblock perf testing, will follow up on the right thing to do here.

rynowak:
@davidfowl - you removed this `UseServer(string)` - what should this code do instead?

@mikeharder - is being able to set the server at the command line still relevant?

mikeharder:
@rynowak: I doubt anyone needs to set the server at the command line, but I wasn't the original author of benchmarks.  @DamianEdwards?

rynowak:
IMO let's just resolve to leave this as-is. If someone needs it in the future, I'm not sure we'll be able to envision their needs today.

PR #93, opened 2016-05-26
Modify all global.json, rather than just benchmarks

mikeharder:
- Required to handle cross-repo breaking changes

@cesarbs, @halter73, @rynowak


rynowak:
:shipit:

PR #95, opened 2016-05-27
add new command line options

nathana1:
* Add new configuration options to toggle interactive thread and choose between kestrel and weblistener
* Add ability to specify multiple scenarios in configuration and a new [default] option for just middleware plaintext and json scenarios
* Fix json scenario so it doesn't emit a BOM

These make it possible to the benchmarks in the TechEmpower automation.

/cc @DamianEdwards @halter73 @mikeharder

### DIFF
--- a/src/Benchmarks/Configuration/ConsoleHostScenariosConfiguration.cs
+++ b/src/Benchmarks/Configuration/ConsoleHostScenariosConfiguration.cs
@@ -22,19 +22,17 @@ namespace Benchmarks.Configuration
             var scenarioConfig = new ConfigurationBuilder()
                 .AddJsonFile("scenarios.json", optional: true)
                 .AddCommandLine(_args)
-                .Build()
-                .GetChildren()
-                .ToList();
+                .Build();
 
             var enabledCount = 0;
-
-            if (scenarioConfig.Count > 0)
+            var configuredScenarios = scenarioConfig["scenarios"];
+            if (!string.IsNullOrWhiteSpace(configuredScenarios))
             {
                 Console.WriteLine("Scenario configuration found in scenarios.json and/or command line args");
-
-                foreach (var scenario in scenarioConfig)
+                var choices = configuredScenarios.Split(',');
+                foreach (var choice in choices)
                 {
-                    enabledCount += scenarios.Enable(scenario.Value);
+                    enabledCount += scenarios.Enable(choice);
                 }
             }
             else
--- a/src/Benchmarks/Program.cs
+++ b/src/Benchmarks/Program.cs
@@ -33,8 +33,9 @@ namespace Benchmarks
                 .AddJsonFile("hosting.json", optional: true)
                 .Build();
 
+            Server = config["server"] ?? "Kestrel";
+
             var webHostBuilder = new WebHostBuilder()
-                .UseKestrel()
                 .UseContentRoot(Directory.GetCurrentDirectory())
                 .UseConfiguration(config)
                 .UseStartup<Startup>()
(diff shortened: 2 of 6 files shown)


src/Benchmarks/Configuration/ConsoleHostScenariosConfiguration.cs
```diff
@@ -11,6 +11,7 @@ namespace Benchmarks.Configuration
     public class ConsoleHostScenariosConfiguration : IScenariosConfiguration
     {
         private readonly string[] _args;
+        private readonly string ScenarioKey = "scenarios";
```
mikeharder: Rename from `ScenarioKey` to `_scenariosKey`.  Note the plural.

src/Benchmarks/Configuration/WebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    public static class WebHostBuilderExtensions
+    {
+        private static readonly string ServerKey = "server";
+        private static readonly string KestrelThreadCountKey = "threadCount";
+        private static readonly string Kestrel = "Kestrel";
+        private static readonly string WebListener = "WebListener";
+        public static IWebHostBuilder UseServer(this IWebHostBuilder builder, IConfigurationRoot config, out string serverName)
+        {
+            var serverValue = config[ServerKey] ?? Kestrel;
```
mikeharder: I think it would be clearer to put this code directly in `Program.Main()`, rather than creating an extension method.

src/Benchmarks/Program.cs
```diff
@@ -16,6 +16,8 @@ public class Program
     {
         public static string[] Args;
         public static string Server;
+        
+        private static readonly string NonInteractiveKey = "NonInteractive";
```
mikeharder: If this is only used once, it's fine to hardcode it rather than define a field.

src/Benchmarks/Configuration/WebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+using System;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+
+namespace Benchmarks.Configuration
+{
+    public static class WebHostBuilderExtensions
+    {
+        private static readonly string ServerKey = "server";
```
mikeharder: I would just hardcode these rather than defining constants, if they're only used a small number of times.

src/Benchmarks/Configuration/ConsoleHostScenariosConfiguration.cs
mikeharder: Actually just hardcode this rather than define a field.

src/Benchmarks/Program.cs
```diff
@@ -44,11 +46,16 @@ public static void Main(string[] args)
...
-            Server = "kestrel";
-
             var webHost = webHostBuilder.Build();
-
-            StartInteractiveConsoleThread();
+            
+            Console.WriteLine($"Using server {Server}");
+            Console.WriteLine($"Server GC is currently {(GCSettings.IsServerGC ? "ENABLED" : "DISABLED")}");
+            
+            var nonInteractiveValue = config[NonInteractiveKey];
```
mikeharder: Why does benchmarks need the ability to *not* start the interactive thread?

nathana1: When running automated in the techempower framework it needs to be a background task, so it doesn't actually have a console.  In that case console.readkey throws and server doesn't start.

mikeharder:
:shipit:

nathana1:
Resolves #98.

nathana1:
Pushed to dev.

PR #101, opened 2016-06-07
Update Newtonsoft.Json version

rynowak:
This was apparently missed when the version was bumped elsewhere.

Removed version references in a few places where we need to do the same thing as the framework. Let's discuss if there's a good reason to have these.

### DIFF
--- a/src/Benchmarks.ServerJob/project.json
+++ b/src/Benchmarks.ServerJob/project.json
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "Repository": "1.0.0-*",
-    "Newtonsoft.Json": "8.0.3"
+    "Newtonsoft.Json": "9.0.1-beta1"
   },
   "frameworks": {
     "netstandard1.0": {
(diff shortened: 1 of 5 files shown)


src/BenchmarksServer/project.json
```diff
@@ -15,7 +15,6 @@
     "Microsoft.AspNetCore.Server.Kestrel": "1.0.0-*",
     "Microsoft.Extensions.CommandLineUtils": "1.0.0-*",
     "System.Diagnostics.Process": "4.1.0-*",
-    "Newtonsoft.Json": "8.0.*",
```
rynowak: This is pinning the version to a lower version than MVC requires, which is what is causing the problem. Floating seems like a better option for this and and our other entry points.

src/Benchmarks.ServerJob/project.json
```diff
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "Repository": "1.0.0-*",
-    "Newtonsoft.Json": "8.0.3"
```
rynowak: I don't see a reason why we'd want this to be less than our "framework minimum" even though it hypothetically could be.

mikeharder:
I don't know anything about our versioning strategy for external dependencies, but this looks reasonable.

:shipit:

PR #107, opened 2016-07-25
add missing techempower db scenarios

### LINKED ISSUE #59: Implement Database update Techempower test
The DB Update test ([#5 at this link](https://www.techempower.com/benchmarks/#section=code)) should also be implemented with the same set of tests as the query tests.

nathana1:
Resolves #59.

- Adds the missing scenarios

- Fixes threading issues with Random and DbContext

- Fixes encoding issues with Fortunes scenario

/cc @DamianEdwards @mikeharder @SajayAntony


src/Benchmarks/Controllers/DbControllerBase.cs
```diff
@@ -0,0 +1,71 @@
...
+                if (_efDb == null)
+                {
+                    _efDb = HttpContext.RequestServices.GetRequiredService<EfDb>();
+                }
+
+                return _efDb;
+            }
+        }
+
+        protected static int ValidateQueries(int queries)
```
mikeharder: Why is this needed?  Since we control both the client and server, I don't think we'd need to validate input in the server like this.

src/Benchmarks/Controllers/DbControllerBase.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        private RawDb _rawDb;
+        private DapperDb _dapperDb;
+        private EfDb _efDb;
+
+        protected RawDb RawDb
+        {
+            get
+            {
+                if (_rawDb == null)
```
mikeharder: I would use `Lazy<T>` to simplify this code.

src/Benchmarks/Data/DapperDb.cs
```diff
@@ -58,6 +58,28 @@ public DapperDb(IRandom random, DbProviderFactory dbProviderFactory, IOptions<Ap
...
+        public async Task<World[]> LoadMultipleUpdatesRows(int count)
+        {
+            var result = await LoadMultipleQueriesRows(count);
+
+            using (var db = _dbProviderFactory.CreateConnection())
+            {
+                db.ConnectionString = _connectionString;
+                await db.OpenAsync();
+
+                for (int i = 0; i < count; i++)
```
mikeharder: Is it by design that we want to perform `count` individual updates to the database, rather than batching the updates to improve performance?

src/Benchmarks/Data/EfDb.cs
```diff
@@ -26,19 +26,33 @@ public Task<World> LoadSingleQueryRow()
...
+        public async Task<World[]> LoadMultipleUpdatesRows(int count)
+        {
+            var result = await LoadMultipleQueriesRows(count, true);
+
+            for (int i = 0; i < count; i++)
+            {
+                result[i].RandomNumber = _random.Next(1, 10001);
+            }
+
+            await _dbContext.SaveChangesAsync();
```
mikeharder: I assume this batches the updates, which may not be an apples-to-apples comparison with the DapperDb scenario.  We should have someone from the EF team review this.

src/Benchmarks/Controllers/DbControllerBase.cs
nathana1: TechEmpower's scenario doc calls out doing this validation.

src/Benchmarks/Controllers/DbControllerBase.cs
```diff
@@ -0,0 +1,71 @@
...
+            get
+            {
+                if (_efDb == null)
+                {
+                    _efDb = HttpContext.RequestServices.GetRequiredService<EfDb>();
+                }
+
+                return _efDb;
+            }
+        }
```
rynowak: Why not create a controller for each of these and just use UI?

src/Benchmarks/Controllers/MultipleUpdatesController.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Benchmarks.Controllers
+{
+    public class MultipleUpdatesController : DbControllerBase
+    {
+        [HttpGet("updates/raw")]
+        [Produces("application/json")]
+        public async Task<object> Raw(int queries = 1)
+        {
+            var count = ValidateQueries(queries);
+            return await RawDb.LoadMultipleUpdatesRows(count);
```
rynowak: Why not just non-async and `return RawDb.LoadMultipleUpdatesRows(count);` This is simpler and doesn't create a state machine

src/Benchmarks/Data/RawDb.cs
```diff
@@ -88,6 +88,45 @@ public RawDb(IRandom random, DbProviderFactory dbProviderFactory, IOptions<AppSe
             return result;
         }
 
+        public async Task<World[]> LoadMultipleUpdatesRows(int count)
+        {
+            var result = await LoadMultipleQueriesRows(count);
+
+            using (var db = _dbProviderFactory.CreateConnection())
```
mikeharder: Currently this creates and closes two DB connections (one in `LoadMultipleQueriesRows`, and again in `LoadMultipleUpdatesRows`).  Perf might be better if a single DB connection was used.  Again, not sure if multiple connections was an intentional choice.  Also, connection pooling (if enabled) might mean there is little perf difference.

src/Benchmarks/Data/RawDb.cs
```diff
@@ -88,6 +88,45 @@ public RawDb(IRandom random, DbProviderFactory dbProviderFactory, IOptions<AppSe
...
+                random.ParameterName = "@Random";
+                id.DbType = DbType.Int32;
+                cmd.Parameters.Add(random);
+
+                for (int i = 0; i < count; i++)
+                {
+                    id.Value = result[i].Id;
+                    var randomNumber = _random.Next(1, 10001);
+                    random.Value = randomNumber;
+                    using (var rdr = await cmd.ExecuteReaderAsync(CommandBehavior.SingleRow))
```
mikeharder: Same question as above about batching of updates.

src/Benchmarks/Startup.cs
```diff
@@ -58,7 +60,7 @@ public void ConfigureServices(IServiceCollection services)
 
             if (Scenarios.Any("Ef"))
             {
-                services.AddScoped<EfDb>();
+                services.AddTransient<EfDb>();
```
mikeharder: Why was EF changed from scoped (once per http request) to transient (once per service request)?

src/Benchmarks/Data/DapperDb.cs
nathana1: I guess it depends on if dapper is considered an orm.  The scenario doc says for raw accesses to do individual queries and updates and an orm can do whatever it does under the covers.  The current dapper implementation does individual queries for the multiple queries scenario and dapper doesn't have first class support for updates, so multiple update queries seemed to be in the spirit of the scenario.

mikeharder: OK, I will review https://www.techempower.com/benchmarks/#section=code and see if I agree.

src/Benchmarks/Data/EfDb.cs
anpete: This will batch. We should play with RelationalDbContextOptionsBuilder.MaxBatchSize to find sweet spot here.

src/Benchmarks/Controllers/DbControllerBase.cs
mikeharder: Alternatively, it's fine to leave this alone for now, and we can use another PR to make code cleanup improvements.  For example, I think `RawDb.cs` can be refactored to remove duplicate code.

src/Benchmarks/Data/DapperDb.cs
anpete: Agree in general with @nathana1 here. Also, EF can read and write using raw SQL, too. See the FromSql and ExecuteSqlCommand APIs. Should we have separate benchmarks for these?

src/Benchmarks/Data/EfDb.cs
anpete: Also EF is transactional here. I'm assuming the others are not. Is this fair?

src/Benchmarks/Data/RawDb.cs
anpete: EF can do external connection management here, too.

src/Benchmarks/Data/EfDb.cs
```diff
@@ -26,19 +26,33 @@ public Task<World> LoadSingleQueryRow()
...
+                var world = withTracking ? _dbContext.World.AsTracking() : _dbContext.World;
+                result[i] = await world.FirstAsync(w => w.Id == id);
             }
 
             return result;
         }
 
+        public async Task<World[]> LoadMultipleUpdatesRows(int count)
+        {
+            var result = await LoadMultipleQueriesRows(count, true);
```
anpete: Use "withTracking: true" for clarity.

src/Benchmarks/Data/DapperDb.cs
```diff
@@ -58,6 +58,28 @@ public DapperDb(IRandom random, DbProviderFactory dbProviderFactory, IOptions<Ap
...
+                {
+                    var randomNumber = _random.Next(1, 10001);
+                    await db.ExecuteAsync("UPDATE world SET randomNumber = @Random WHERE id = @Id", new { Random = randomNumber, Id = result[i].Id });
+                    result[i].RandomNumber = randomNumber;
+                }
+
+                db.Close();
```
mikeharder: I think `db.Close()` is unnecessary, since it will be disposed by the `using` statement.

src/Benchmarks/Data/DapperDb.cs
mikeharder: The spec seems pretty clear that batching is allowed (Test 5, Requirement 9):

> Use of batch updates is acceptable but not required.

So I think Raw, Dapper, and EF should all batch updates if possible.

src/Benchmarks/Data/EfDb.cs
mikeharder: Spec says the following about transactions (Test 5, Requirement 10):

> Use of transactions is acceptable but not required. If transactions are used, a transaction should only encapsulate a single iteration, composed of a single read and single write. Transactions should not be used to consolidate multiple iterations into a single operation.

I assume it will be fastest to disable transactions for all (Raw, Dapper, and EF)?

src/Benchmarks/Data/RawDb.cs
mikeharder: I think we should use a single DB connection since the spec doesn't say otherwise, and it may be faster.

src/Benchmarks/Data/RawDb.cs
```diff
@@ -88,6 +88,45 @@ public RawDb(IRandom random, DbProviderFactory dbProviderFactory, IOptions<AppSe
...
+                    {
+                        await rdr.ReadAsync();
+                        result[i].RandomNumber = randomNumber;
+                    }
+                }
+
+                db.Close();
```
mikeharder: I think db.Close() is unnecessary, since it will be disposed by the using statement.

src/Benchmarks/Data/EfDb.cs
anpete: So, we are currently in violation of the spec because we have a single transaction around all writes. Easiest fix right now is to move SaveChanges inside the loop, which gets us to a transaction per iteration. Disabling transactions completely will require injecting a custom service; let me know if you want to proceed with that and I can figure out how to do it.

src/Benchmarks/Data/DapperDb.cs
anpete: The fastest way is likely to use a TVP to stream the updated values to SQL Server. Statement concatenation (what we do in EF) may be faster when the number of updates is low.

src/Benchmarks/Data/EfDb.cs
anpete: So, discussed with some folks and we think this should be the fastest approach for EF:

1) Disable EF transactions via service replacement. 
2) No-tracking query of the entities.
3) Spin the loop updating the entities.
4) Call DbContext.UpdateRange
5) Call SaveChanges

src/Benchmarks/Data/RawDb.cs
anpete: Why is this using ExecuteReader? It should likely be ExecuteNonQuery.

src/Benchmarks/Data/EfDb.cs
nathana1: Can you send a code sample with transactions disabled?

src/Benchmarks/Data/NoTransactionSqlServerConnection.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Storage;
+using Microsoft.EntityFrameworkCore.Storage.Internal;
+using Microsoft.Extensions.Logging;
+
+namespace Benchmarks.Data
+{
+    class NoTransactionSqlServerConnection : SqlServerConnection
```
mikeharder: Is this the recommended way to disable transactions in EF?

mikeharder:
:shipit: once we decide on batched updates, and you've confirmed we're using the recommended best practice to disable transactions in EF.

src/Benchmarks/Data/NoTransactionSqlServerConnection.cs
divega: There is no recommended way to disable transactions in EF because we have never recommended it to anyone before. If it turns out to be really much faster we can consider having an easier opt-in.

src/Benchmarks/Middleware/MultipleQueriesEfMiddleware.cs
```diff
@@ -21,20 +22,19 @@ public class MultipleQueriesEfMiddleware
         };
 
         private readonly RequestDelegate _next;
-        private readonly EfDb _db;
 
-        public MultipleQueriesEfMiddleware(RequestDelegate next, EfDb db)
+        public MultipleQueriesEfMiddleware(RequestDelegate next)
```
mikeharder: When I run a benchmark with these changes, the first 50 or so requests work fine, but then RPS goes to 0.  If I revert this file, RPS is similar to the dev branch.

mikeharder: It appears the changes in this file significantly increase the load on SQL, which causes SQL to be the bottleneck instead of the web server.

mikeharder: It's possible the changes in this PR are correct, and the code in the dev branch was incorrectly caching too much in memory.  Either way, let's understand the difference and which is more representative of the intended scenario.

src/Benchmarks/Data/ApplicationDbContext.cs
```diff
@@ -22,7 +27,34 @@ public ApplicationDbContext(IOptions<AppSettings> appSettings)
...
+            }
+
+            public NoTxSqlServerOptionsExtension(NoTxSqlServerOptionsExtension copyFrom) : base(copyFrom)
+            {
+            }
+            public override void ApplyServices(IServiceCollection services)
+            {
+                base.ApplyServices(services);
+                services.Remove(services.First((sd) => sd.ServiceType == typeof(ISqlServerConnection)));
+                services.AddScoped<ISqlServerConnection, NoTransactionSqlServerConnection>();
```
ajcvickers: This is an unusual way to do service replacement. I'm guessing this is because you don't want EF to share the same service provider as the application and you don't want to manage the service provider yourself?

The more normal way to do this would be to change AddDbContext as follows:
```C#
services
    .AddScoped<ISqlServerConnection, NoTransactionSqlServerConnection>()
    .AddEntityFrameworkSqlServer()
    .AddDbContext<ApplicationDbContext>(
        (serviceProvider, options) => options
            .UseInternalServiceProvider(serviceProvider)
            .UseSqlServer(Configuration.GetConnectionString("DefaultConnection")));
```

That being said, if you want to do it this way, then its probably fine.

src/Benchmarks/Startup.cs
```diff
@@ -49,7 +51,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton<ApplicationDbSeeder>();
             services.AddEntityFrameworkSqlServer()
```
ajcvickers: If you are not using the application service provider for internal EF services, then there is no point in adding all these services, so you should remove AddEntityFrameworkSqlServer.

nathana1:
Opened #119 and #120 to track remaining issues.  Pushed as https://github.com/aspnet/benchmarks/commit/87965610a0fac32dfa29445cb44affdbfe7bd364

PR #108, opened 2016-07-26
Add HTTPS to Benchmarks

mikeharder:
- Configure Kestrel to use self-signed cert
- Add scheme parameter to driver and server

/cc @DamianEdwards @pakrym @SajayAntony @nathana1


halter73:
Is "testCert.pfx" the one from the Kestrel repo? Adding this reminds me that the cert expires September 15th of this year. I'm thinking maybe we should replace it, but I'm not sure if it matters since we don't validate the cert.

mikeharder:
@halter73: Yes I copied the cert from the Kestrel repo.  wrk disables cert validation so I doubt it matters.

src/Benchmarks/Program.cs
```diff
@@ -50,6 +50,8 @@ public static void Main(string[] args)
                 var threads = GetThreadCount(config);
                 webHostBuilder = webHostBuilder.UseKestrel((options) =>
                 {
+                    options.UseHttps("testCert.pfx", "testPassword");
```
halter73: Kestrel does a little bit of work creating a `ConnectionFilterContext` a `LibuvStream` and invoking an `HttpsConnectionFilter.OnConnectionAsync` which basically no-ops but still goes through the async state machine.

It's probably best not to call this in the benchmarks app if the URL scheme isn't HTTPS.

SajayAntony:
:shipit:

rynowak:
:shipit:

mikeharder:
I still need to add the SQL code, but it's straightforward so I'll merge this once its done.

halter73:
:ship:

PR #109, opened 2016-07-30
Add support for Kestrel ConnectionFilters

mikeharder:
- Add `connectionFilter` option to Benchmarks and automation
- Add class `Benchmarks.PassthroughConnectionFilter` which just wraps the previous stream in a passthrough stream
- Add scenario for Plaintext with PassthroughConnectionFilter to `makefile.shade` target run by CI machine

@halter73, @DamianEdwards, @nathana1, @rynowak


src/Benchmarks/Program.cs
```diff
@@ -133,6 +140,21 @@ private static int GetThreadCount(IConfigurationRoot config)
             var threadCountValue = config["threadCount"];
             return threadCountValue == null ? -1 : int.Parse(threadCountValue);
         }
+
+        private static IConnectionFilter GetConnectionFilter(IConfigurationRoot config, IConnectionFilter prevFilter)
+        {
+            var connectionFilterValue = config["connectionFilter"];
+            if (String.IsNullOrEmpty(connectionFilterValue))
```
davidfowl: nit: lowercase S (we changed to that style)

src/Benchmarks/Program.cs
```diff
@@ -133,6 +140,21 @@ private static int GetThreadCount(IConfigurationRoot config)
...
+        private static IConnectionFilter GetConnectionFilter(IConfigurationRoot config, IConnectionFilter prevFilter)
+        {
+            var connectionFilterValue = config["connectionFilter"];
+            if (string.IsNullOrEmpty(connectionFilterValue))
+            {
+                return null;
+            }
+            else
+            {
+                var connectionFilterType = Type.GetType(connectionFilterValue, throwOnError: true);
```
davidfowl: Do you envision us having different connection filters with no arguments? Why not make it a boolean flag? On or off?

mikeharder: I could see us testing other connection filters like Logging, or a dev may want to add a custom connection filter to a branch of Benchmarks and test that.  Making the connection filter a type string gives more flexibility without needing to change the `BenchmarksDriver` or `BenchmarksServer`.

davidfowl: The problem is that those connection filters are non trivial to create as they have arguments. You'll end up hard coding the construction of specific filters. For example, logging https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Filter/LoggingConnectionFilter.cs, you'd have a pass an `ILogger` and for SSL https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionFilter.cs, you need to pass options. That's why the extension methods exist.

halter73: Benchmarking the logging filter probably isn't too important.  HTTPS is handled by another PR. We can always change it later, but I think this is fine for now.

mikeharder:
While I agree we may never actually use this to test another connection filter, it seems like a better design than adding a very specific parameter (and column in SQL) like "passthroughConnectionFilter".

src/Benchmarks/Program.cs
mikeharder: While I agree we may never actually use this to test another connection filter, it seems like a better design than adding a very specific parameter (and column in SQL) like "passthroughConnectionFilter".

makefile.shade
```diff
@@ -16,9 +16,16 @@ k-standard-goals
 #benchmark-tests .compile
   @{
     var sql = string.IsNullOrEmpty(E("BENCHMARK_SQL")) ? string.Empty : "-q " + E("BENCHMARK_SQL");
-    var scenarios = new string[] { "plaintext", "json", "mvcplaintext", "mvcjson" };
+    var scenarios = new string[]
+    {
+      "-n plaintext",
+      "-n plaintext --connectionFilter Benchmarks.PassthroughConnectionFilter",
```
rynowak: camelCase doesn't seem like the right convention - should this be something like `--connection-filter`?

The other thought I had is that it might be better to call it "connection filter type" so it's explicit that's what it does. I could go either way

src/Benchmarks.ServerJob/ServerJob.cs
```diff
@@ -13,6 +13,8 @@ public class ServerJob : IIdentifiable
     {
         public int Id { get; set; }
 
+        public string ConnectionFilter { get; set; }
```
rynowak: I think this is fine for now, but if we do more of this we'll have to add "scenario args" because this doesn't scale super well

src/Benchmarks/Program.cs
```diff
@@ -133,6 +140,21 @@ private static int GetThreadCount(IConfigurationRoot config)
             var threadCountValue = config["threadCount"];
             return threadCountValue == null ? -1 : int.Parse(threadCountValue);
         }
+
+        private static IConnectionFilter GetConnectionFilter(IConfigurationRoot config, IConnectionFilter prevFilter)
```
rynowak: It might be simpler to just handle all of the setting of `options.ConnectionFilter` in this method and remove the null-check + if in the calling code.

For instance this method will set `options.ConnectionFilter` to `null` when nothing is specified on the command line, but will daisy chain if a connection filter is set. This doesn't seem intended.

makefile.shade
rynowak: Another thought, should we add a "display name" or something like that? I'm not sure we want to surface plaintext & plaintext + connection filter as the same thing. I'm happy to treat this as a future enhancement

rynowak:
:shipit: - all my feedback here is pretty minor and all about how to approach our future needs

makefile.shade
mikeharder: The existing convention used in the benchmarks app was camelCase for the config key: https://github.com/aspnet/benchmarks/blob/dev/src/Benchmarks/Program.cs#L133.  Do you think this should be changed, or just the command-line parameters?

src/Benchmarks/Program.cs
mikeharder: I don't understand what you think is unintended.  If the `--connectionFilter` command-line arg is set, then we create an instance of the type, passing either the previous connection filter or the special `NoOpConnectionFilter`.  If the `--connectionFilter` command-line arg is not set, then `options.ConnectionFilter` is not changed at all.

makefile.shade
mikeharder: Regarding the "display name", we will not surface "plaintext" and "plaintext + connection filter" as the same thing.  The "ConnectionFilter" column in SQL will be used to filter the data appropriately.

rynowak: > The existing convention used in the benchmarks app was camelCase for the config key

Hmm, I guess this is one of those cases where things will seem weird one way or another. I was just remarking that seeing a camelCase commandline argument is strange. Best leave it then if this is going to make things more confusing.

src/Benchmarks/Program.cs
rynowak: yeah you're right.

PR #118, opened 2016-08-17
Remove ErrorHandlerMiddleware

mikeharder:
- Default behavior of Kestrel is to send a 500 response (with no body) and log the exception, which is sufficient.

@nathana1, @rynowak


nathana1:
:shipit:

PR #121, opened 2016-08-26
add support to run against postgresql

nathana1:


src/Benchmarks/Data/ApplicationDbContext.cs
```diff
@@ -19,17 +19,27 @@ public class ApplicationDbContext : DbContext
         public ApplicationDbContext(IOptions<AppSettings> appSettings)
         {
             _appSettings = appSettings.Value;
+            UseBatchUpdate = _appSettings.Database != "postgresql";
```
mikeharder: Why is batch update disabled for postgres?

src/Benchmarks/Data/ApplicationDbContext.cs
```diff
@@ -19,17 +19,27 @@ public class ApplicationDbContext : DbContext
...
         }
 
         public DbSet<World> World { get; set; }
 
         public DbSet<Fortune> Fortune { get; set; }
 
+        public bool UseBatchUpdate { get; }
```
mikeharder: NIT: I would make this property return `_appSettings.Database != "postgresql"`, rather than assigning the value in the ctor.  I think it's cleaner since the semantics of the property are in the property getter itself, rather than in the class ctor.

src/Benchmarks/Configuration/AppSettings.cs
```diff
@@ -7,5 +7,6 @@ namespace Benchmarks.Configuration
     public class AppSettings
     {
         public string ConnectionString { get; set; }
+        public string Database { get; set; } = "sqlserver";
```
mikeharder: I would prefer to use an enum rather than strings.  Create a enum named something like `Database` with values `SqlServer` and `PostgreSQL`.  Then change this property in AppSettings from `string` to `Database`, and the framework will handle converting the string in `appsettings.json` to the correct enum value.

src/Benchmarks/Startup.cs
```diff
@@ -54,8 +55,15 @@ public void ConfigureServices(IServiceCollection services)
             
             if (Scenarios.Any("Raw") || Scenarios.Any("Dapper"))
             {
-                // TODO: Add support for plugging in different DbProviderFactory implementations via configuration
-                services.AddSingleton<DbProviderFactory>(SqlClientFactory.Instance);
+                var database = Configuration["database"] ?? "sqlserver";
```
mikeharder: Can this be changed to use the strongly-typed AppSettings, rather than the string from config?

src/Benchmarks/Data/ApplicationDbContext.cs
nathana1: Postgres has a problem with deadlocking when batch updates aren't sorted by key and ef doesn't emit batches as sorted.

mikeharder:
:shipit:

nathana1:
Resolved with https://github.com/aspnet/benchmarks/commit/a7a2448fe2282969034f51d66b823a4558197934.

PR #123, opened 2016-09-06
Use cached repsonse

benaadams:


nathana1:
Looks fine to me.

/cc @mikeharder

mikeharder:
@benaadams: Seems fine, but does this result in a measurable improvment?

benaadams:
> HttpContext.Response is already a simple field getter?

It isn't yet... https://github.com/aspnet/HttpAbstractions/pull/704

mikeharder:
@benaadams: Assuming this is the correct class, how is this not a simple field getter?

https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http/DefaultHttpContext.cs#L101

Also, I cannot detect any improvement in RPS from this change.

benaadams:
Ahh.. you are right its the next bit that is the look up.

PR #124, opened 2016-09-08
improve how the ef dbcontext is setup (#120)

### LINKED ISSUE #120: improve how the ef dbcontext is setup
From the comments in https://github.com/aspnet/benchmarks/pull/107:

This is an unusual way to do service replacement. I'm guessing this is because you don't want EF to share the same service provider as the application and you don't want to manage the service provider yourself?
The more normal way to do this would be to change AddDbContext as follows:
services
    .AddScoped<ISqlServerConnection, NoTransactionSqlServerConnection>()
    .AddEntityFrameworkSqlServer()
    .AddDbContext<ApplicationDbContext>(
        (serviceProvider, options) => options
            .UseInternalServiceProvider(serviceProvider)
            .UseSqlServer(Configuration.GetConnectionString("DefaultConnection")));
That being said, if you want to do it this way, then its probably fine.

nathana1:
Resolves #120.

/cc @mikeharder @anpete


mikeharder:
:shipit:

PR #127, opened 2016-09-20
HTTP benchmarking tool

mikeharder:
@davidfowl @mikaelm12 @halter73 @cesarbs @anurse 

Simple HTTP benchmarking tool in C#

### DIFF
--- a/experimental/HttpBenchmark/Program.cs
+++ b/experimental/HttpBenchmark/Program.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Sockets;
@@ -38,6 +39,8 @@ namespace HttpBenchmark
                 CommandOptionType.SingleValue);
             var keepaliveOption = app.Option("-k|--keepalive", "Reuse connections for multiple requests.  Default is TRUE.",
                 CommandOptionType.SingleValue);
+            var methodOption = app.Option("-m|--method", "Request method.  Default is GET.", CommandOptionType.SingleValue);
+            var bodyOption = app.Option("-b|--body", "Request body.", CommandOptionType.SingleValue);
 
             var urlArgument = app.Argument("url", "URL to benchmark");
 
@@ -67,6 +70,15 @@ namespace HttpBenchmark
                     keepaliveValue = "true";
                 }
 
+                var method = methodOption.Value();
+                if (string.IsNullOrEmpty(method))
+                {
+                    method = "GET";
+                }
+                method = method.ToUpperInvariant();
+
+                var body = bodyOption.Value();
+
                 var url = urlArgument.Value;
 
                 int connections;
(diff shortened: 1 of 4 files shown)


experimental/HttpBenchmark/Program.cs
```diff
@@ -0,0 +1,290 @@
...
+            var app = new CommandLineApplication()
+            {
+                Name = "HttpBenchmark",
+                FullName = "HTTP Benchmark",
+                Description = "HTTP benchmarking tool"
+            };
+
+            app.HelpOption("-?|-h|--help");
+
+            var connectionsOption = app.Option("-c|--connections", "Number of connections.  Default is 256.", CommandOptionType.SingleValue);
```
cesarbs: Put default values in constants and `$"{use}"` them here.

experimental/HttpBenchmark/Program.cs
```diff
@@ -0,0 +1,375 @@
...
+                    }
+                    else if (line == "Transfer-Encoding: chunked")
+                    {
+                        chunked = true;
+                        break;
+                    }
+                }
+            }
+
+            if (contentLength != null)
```
rynowak: @rynowak
rynowak Pending
Would it be possible to also check the status code here? I have the heebie-jeebies about any sort of automation that can report perf information when the run wasn't successful.

experimental/HttpBenchmark/Program.cs
```diff
@@ -0,0 +1,375 @@
...
+
+            // Calculate response length
+            using (var socket = CreateSocket(uri).Result)
+            {
+                Console.WriteLine("Request:");
+                Console.Write(requestString);
+                socket.Send(requestBytes);
+
+                var responseBuffer = new byte[1024];
+                int bytesReceived = 0;
```
halter73: This will probably need to be a long.

mikeharder: I updated the tool to support responses up to 1 MB, rather than the 1KB previously supported.  The `bytesReceived` is still fine as an `int`, unless we want to support responses larger than `int.MaxValue`.

experimental/HttpBenchmark/Program.cs
mikeharder: I updated the tool to verify the status code of the initial response used to calculate the response length.  However, subsequent responses are not yet verified.

Currently, the tool takes a lot of shortcuts and makes a lot of assumptions to achieve high performance.  It executes a single request and does expensive response parsing to determine the response length and verify the status code.  For the load test, it assumes all responses have the same content length, and it doesn't verify the status code.  This would break in a lot of scenarios.  For example, a chunked response can be chunked in many different ways yet all are the same effective response.

A better implementation would be to fully parse every response, handle chunking, etc.  And do this efficiently enough so the client doesn't become a bottleneck for a high-performance server like Kestrel Plaintext.  This can be added if and when this tool becomes popular.

experimental/HttpBenchmark/Program.cs
```diff
@@ -0,0 +1,384 @@
...
+                lastElapsed = elapsed;
+
+                WriteResult(_requests, currentRequests, currentElapsed);
+
+                if (elapsed > duration)
+                {
+                    Console.WriteLine();
+                    Console.WriteLine($"Average RPS: {Math.Round(_requests / elapsed.TotalSeconds)}");
+
+                    Environment.Exit(0);
```
anurse: Maybe I'm just being too much of a purist, but I kinda feel like this is a bit of a weird way to shut down the benchmark. It took me a bit to realize the benchmark finished here. I guess I'd kinda prefer something like a timed cancellation token or something that collected the results after the run... but I also see the value in collecting as we go.

I guess I'm OK with this model, but could we put some comments in the benchmark loops themselves indicating how they are terminated? `while(true)` always makes me nervous, especially if I can't see the `break` anywhere in the loop :)

mikeharder: Changed to use timed `CancellationToken`.

experimental/HttpBenchmark/Program.cs
rynowak: The main thing I care about here is the status code. We can't identify all bugs in a performance testing scenario, but status code will catch the common ones which have to do with the site just plain not working.

experimental/HttpBenchmark/Program.cs
```diff
@@ -0,0 +1,380 @@
...
+                    app.ShowHelp();
+                    Console.WriteLine(e);
+                    return 2;
+                }
+            });
+
+            return app.Execute(args);
+        }
+
+        private static int Run(int connections, TimeSpan duration, int pipeline, bool keepalive, string method, string body, Uri uri)
```
troydai: nit: borderline too long.

experimental/HttpBenchmark/Program.cs
```diff
@@ -0,0 +1,380 @@
...
+
+                var elapsed = _stopwatch.Elapsed;
+                var currentElapsed = elapsed - lastElapsed;
+                lastElapsed = elapsed;
+
+                WriteResult(_requests, currentRequests, currentElapsed);
+            }
+
+            Console.WriteLine();
+            Console.WriteLine($"Average RPS: {Math.Round(_requests / _stopwatch.Elapsed.TotalSeconds)}");
```
troydai: nit: this not a complain but more like a feature request. can we save the result in a json file. it would be useful for pipeline the tools.

mikeharder: @troydai: Feel free to submit a PR to add this feature.

PR #130, opened 2016-09-23
Lowercase table names

roji:
This makes raw queries compatible with PostgreSQL. Also modified raw SQL commands to use lowercase column and table names.

### DIFF
--- a/src/Benchmarks/Data/DapperDb.cs
+++ b/src/Benchmarks/Data/DapperDb.cs
@@ -40,7 +40,7 @@ namespace Benchmarks.Data
         async Task<World> ReadSingleRow(DbConnection db)
         {
             return await db.QueryFirstOrDefaultAsync<World>(
-                    "SELECT Id, RandomNumber FROM World WHERE Id = @Id",
+                    "SELECT id, randomnumber FROM world WHERE Id = @Id",
                     new { Id = _random.Next(1, 10001) });
         }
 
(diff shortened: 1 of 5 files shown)


src/Benchmarks/Data/DapperDb.cs
```diff
@@ -40,7 +40,7 @@ public DapperDb(IRandom random, DbProviderFactory dbProviderFactory, IOptions<Ap
         async Task<World> ReadSingleRow(DbConnection db)
         {
             return await db.QueryFirstOrDefaultAsync<World>(
-                    "SELECT Id, RandomNumber FROM World WHERE Id = @Id",
+                    "SELECT id, randomNumber FROM world WHERE Id = @Id",
```
davidfowl: the casing here is inconsistent with update

roji: You're right, sorry, will amend right away

mikeharder:
I thought PostgreSQL was supposed to be working already, after https://github.com/aspnet/benchmarks/commit/a7a2448fe2282969034f51d66b823a4558197934.  @nathana1?

nathana1:
I wonder if we're getting different behavior when PostgreSQL is populated with the TechEmpower scripts vs. the db seeder included with the our benchmarks (I used the TechEmpower script, I recommended Shay use the seeder to save some steps).  

The TechEmpower script (https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/config/create-postgres.sql) creates the tables twice:

CREATE TABLE  World (
...
CREATE TABLE  "World" (

So I think those are treated as world and World by PostgreSQL.

roji:
Yeah, that would explain it. PostgreSQL automatically lower-cases unquoted names, and the seeder included in the benchmarks creates "World".

mikeharder:
@nathana1: Do you think we should merge this PR?  You are the best person to decide.

nathana1:
I did a quick pass on this branch to make sure it doesn't break SqlServer and it looks good.  

:shipit:

PR #133, opened 2016-10-03
Only enable scenarios starting with the partial name

mikeharder:
- Previous code enabled scenarios containing the partial name anywhere, which made it impossible to enable "Plaintext" without "MvcPlaintext".

@DamianEdwards, @nathana1


mikeharder:
There appears to be no impact on Plaintext RPS, meaning that enabling MvcPlaintext added no overhead.

PR #134, opened 2016-10-04
WIP: Add scenario for plaintext with response caching

mikeharder:
@JunTaoLuo, @BrennanConroy

### DIFF
--- a/src/Benchmarks/Middleware/CachedPlaintextMiddleware.cs
+++ b/src/Benchmarks/Middleware/CachedPlaintextMiddleware.cs
@@ -1,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+using System;
+using System.Text;
+using System.Threading.Tasks;
+using Benchmarks.Configuration;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+
+namespace Benchmarks.Middleware
+{
+    public class CachedPlaintextMiddleware
+    {
+        private static readonly PathString _path = new PathString(Scenarios.GetPath(s => s.CachedPlaintext));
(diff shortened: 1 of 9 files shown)


src/Benchmarks/Middleware/CachedPlaintextMiddleware.cs
```diff
@@ -0,0 +1,43 @@
...
+        private readonly RequestDelegate _next;
+
+        public CachedPlaintextMiddleware(RequestDelegate next)
+        {
+            _next = next;
+        }
+
+        public Task Invoke(HttpContext httpContext)
+        {
+            if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal))
```
JunTaoLuo: Is there no way to avoid this? Not sure how much overhead this adds.

mikeharder: It should be fine, since it's done by all middlewares in Benchmarks including Plaintext.

JunTaoLuo: Hmm I see that the other middlewares also uses the same pattern.

src/Benchmarks/Middleware/CachedPlaintextMiddleware.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public Task Invoke(HttpContext httpContext)
+        {
+            if (httpContext.Request.Path.StartsWithSegments(_nocachePath, StringComparison.Ordinal))
+            {
+                return PlaintextMiddleware.WriteResponse(httpContext.Response);
+            }
+            else if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal))
+            {
+                httpContext.Response.Headers["cache-control"] = "public, max-age=1";
```
JunTaoLuo: Not really realistic, but we want to simulate entries that expire.

PR #136, opened 2016-10-05
Add scenario for MemoryCache

mikeharder:
@JunTaoLuo, @BrennanConroy, @Tratcher

### DIFF
--- a/src/Benchmarks/Middleware/MemoryCachePlaintextMiddleware.cs
+++ b/src/Benchmarks/Middleware/MemoryCachePlaintextMiddleware.cs
@@ -34,8 +34,10 @@ namespace Benchmarks.Middleware
                 {
                     payload = Encoding.UTF8.GetBytes("Hello, World!");
 
-                    var cacheEntry = _memoryCache.CreateEntry(_key);
-                    cacheEntry.Value = payload;
+                    using (var cacheEntry = _memoryCache.CreateEntry(_key))
+                    {
+                        cacheEntry.SetValue(payload).SetAbsoluteExpiration(TimeSpan.FromSeconds(1));
+                    }
                 }
 
                 httpContext.Response.StatusCode = 200;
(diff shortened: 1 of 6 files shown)


BrennanConroy:
:shipit:
Although do we want to add a scenario with reads and writes? What is the default expiration of a memory cache entry?

src/Benchmarks/Middleware/MemoryCachePlaintextMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+        public Task Invoke(HttpContext httpContext)
+        {
+            if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal))
+            {
+                byte[] payload;
+                if (!_memoryCache.TryGetValue(_key, out payload))
+                {
+                    payload = Encoding.UTF8.GetBytes("Hello, World!");
+
+                    var cacheEntry = _memoryCache.CreateEntry(_key);
```
Tratcher: using, this doesn't actually add the entry without it.

src/Benchmarks/Middleware/MemoryCachePlaintextMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+                    cacheEntry.Value = payload;
```
Tratcher: Add an expiration

Tratcher:
There is no default expiration time, only a memory preasure trigger

src/Benchmarks/Middleware/MemoryCachePlaintextMiddleware.cs
mikeharder: Yes, i will add a 1-second expiration for parity with the ResponseCaching scenario.

src/Benchmarks/Middleware/MemoryCachePlaintextMiddleware.cs
mikeharder: Got it.  It wasn't clear how this was supposed to work from the interfaces.  I assumed `CreateEntry()` added it to the cache as well.

src/Benchmarks/Startup.cs
```diff
@@ -115,6 +115,11 @@ public void ConfigureServices(IServiceCollection services)
                 }
             }
 
+            if (Scenarios.Any("MemoryCache"))
```
Tratcher: Shouldn't this match MemoryCachePlaintext?

mikeharder: It's future-proofing in case we add more scenarios which start with "MemoryCache".

JunTaoLuo:
:ship::it:

PR #137, opened 2016-10-06
Add scenario for ResponseCaching when client sends no-cache header

mikeharder:
- Requires adding header support to Benchmark automation
- Split each ResponseCaching scenario into its own middleware, to ensure each scenario only requires one path comparison

@JunTaoLuo, @BrennanConroy, @DamianEdwards


BrennanConroy:
`Hit` is a weird way to describe that scenario, since it is hitting and missing the cache

src/Benchmarks/Middleware/ResponseCachingPlaintextNocacheMiddleware.cs
```diff
@@ -0,0 +1,42 @@
...
+        public ResponseCachingPlaintextNocacheMiddleware(RequestDelegate next)
+        {
+            _next = next;
+        }
+
+        public Task Invoke(HttpContext httpContext)
+        {
+            if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal))
+            {
+                httpContext.Response.Headers["cache-control"] = "public, max-age=1";
```
BrennanConroy: Forgot this is the response

mikeharder:
@BrennanConroy: We want to test 3 scenarios.  One where most requests are served from the cache (currently named "Hit"), one where the response doesn't include a "cache-control" header so is never served from the cache (currently named "Miss"), and one where the client sends a "no-cache" header so the ResponseCaching middleware should short-circuit early (currently named "Nocache").  I'm happy to rename them if you can think of better names.

src/BenchmarksClient/Startup.cs
```diff
@@ -97,7 +96,15 @@ public static int Main(string[] args)
                 var job = allJobs.FirstOrDefault();
                 if (job != null)
                 {
-                    var jobLogText = $"[ID:{job.Id} Connections:{job.Connections} Threads:{job.Threads} Duration:{job.Duration} Pipeline:{job.PipelineDepth}]";
+                    var jobLogText = $"[ID:{job.Id} Connections:{job.Connections} Threads:{job.Threads} " +
+                        $"Duration:{job.Duration} Pipeline:{job.PipelineDepth}";
+
+                    if (job.Headers != null)
+                    {
+                        jobLogText += $" Headers:{PP(job.Headers)}";
```
JunTaoLuo: What's PP?

src/BenchmarksClient/Startup.cs
```diff
@@ -194,5 +212,10 @@ private static void Log(string message)
             var time = DateTime.Now.ToString("hh:mm:ss.fff");
             Console.WriteLine($"[{time}] {message}");
         }
+
+        private static string PP(string[] args)
```
JunTaoLuo: Oh, It's a function. Would have been nice if this had a more descriptive name.

src/BenchmarksDriver/Program.cs
```diff
@@ -401,5 +428,10 @@ private static void Log(string message, Reporter reporter)
             var time = DateTime.Now.ToString("hh:mm:ss.fff");
             reporter.WriteLine($"[{time}] {message}");
         }
+
+        private static string PP(string[] args)
```
JunTaoLuo: Hmm should this be added to some util class so it's not duplicated?

JunTaoLuo:
Would recommend the names ServeCached (or ServeCachedResponse), NeverCached (Or NeverCacheResponse), and RequestNoCache

src/BenchmarksDriver/Program.cs
mikeharder: I'll change this to an extension method in `Benchmarks.ClientJob` so it can be shared across the Driver and Client.

mikeharder:
@JunTaoLuo: So the full scenario names would be `ResponseCachingPlaintextServeCached`, `ResponseCachingPlaintextNeverCached`, and `ResponseCachingPlaintextRequestNoCache`, with matching URLs.  These are pretty long, but I'm fine with them.

JunTaoLuo:
@mikeharder Maybe start using abbreviations ResponseCachingPlaintext => RCP?

PR #139, opened 2016-10-10
More robust delete

BrennanConroy:
@mikeharder

### DIFF
--- a/src/BenchmarksServer/Startup.cs
+++ b/src/BenchmarksServer/Startup.cs
@@ -241,7 +241,19 @@ namespace BenchmarkServer
             {
                 info.Attributes = FileAttributes.Normal;
             }
-            dir.Delete(recursive: true);
+
+            for (var i = 0; i < 3; i++)
+            {
+                try
+                {
+                    dir.Delete(recursive: true);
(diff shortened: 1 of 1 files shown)


BrennanConroy:
Pretty sure its just the OS hasn't released the Benchmarks.dll yet since we delete immediately after killing the process.

mikeharder:
@BrennanConroy: I'm not so sure.  Here's the managed exception when trying to delete a directory which contains a file in us:

```Unhandled Exception: System.IO.IOException: The process cannot access the file 'test.txt' because it is being used by an
other process.```

The exception we have been seeing matches the one thrown when trying to delete a directory which contains a read-only file:

```Unhandled Exception: System.UnauthorizedAccessException: Access to the path 'test.txt' is denied.```

The code in `DeleteDir()` tries to prevent this by setting `FileAttributes.Normal` on every file before deleting the directory, but this seems to be occasionally insufficient.

mikeharder:
@BrennanConroy: Lets merge https://github.com/aspnet/benchmarks/pull/142 and wait for another repro before attempting a fix.

PR #141, opened 2016-10-11
Add scenario for response caching with varyby rules

JunTaoLuo:
cc @mikeharder @BrennanConroy

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -30,6 +30,7 @@ k-standard-goals
       "-n ResponseCachingPlaintextCached",
       "-n ResponseCachingPlaintextResponseNoCache",
       "-n ResponseCachingPlaintextRequestNoCache",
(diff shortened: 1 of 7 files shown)


makefile.shade
```diff
@@ -30,6 +30,7 @@ k-standard-goals
       "-n ResponseCachingPlaintextCached",
       "-n ResponseCachingPlaintextResponseNoCache",
       "-n ResponseCachingPlaintextRequestNoCache",
+      "-n ResponseCachingPlaintextVaryByCached",
```
BrennanConroy: Do we want this in the automated runs?

BrennanConroy:
🐑 :it:

makefile.shade
mikeharder: One more test is fine if we think it will be useful.  We can always remove tests if the run starts taking too long.

PR #142, opened 2016-10-11
Log handles and file attributes if delete fails

mikeharder:
@BrennanConroy


BrennanConroy:
:shipit:

BrennanConroy:
Do these programs do all files in the directory?

mikeharder:
@BrennanConroy: They should.  `handle.exe` displays handles to objects starting with the command-line parameter.  And the `/s` parameter to `attrib.exe` is for recursive.

PR #149, opened 2016-10-28
Use prepared SQL statements

mikeharder:
- Improves PostgreSQL performance by 10-15%

@nathana1


src/Benchmarks/Data/RawDb.cs
```diff
@@ -39,6 +39,9 @@ public RawDb(IRandom random, DbProviderFactory dbProviderFactory, IOptions<AppSe
         
         async Task<World> ReadSingleRow(DbConnection connection, DbCommand cmd)
         {
+            // Prepared statements improve PostgreSQL performance by 10-15%
+            cmd.Prepare();
```
nathana1: Does this work?  I needed to put it after db.OpenAsync when I was testing with fortunes.

PR #157, opened 2016-11-17
Use EF's new user-cached compiled queries and update scenario tweak

anpete:
Avoids the internal automatic query cache. Typically yields 3-5% perf. improvement. Tweaked the update scenario to do an explicit state update vs. automatic change detection.

### DIFF
--- a/src/Benchmarks/Data/EfDb.cs
+++ b/src/Benchmarks/Data/EfDb.cs
@@ -1,10 +1,13 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved. 
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
 
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Benchmarks.Configuration;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Query;
 using Microsoft.Extensions.Options;
 
 namespace Benchmarks.Data
@@ -14,19 +17,23 @@ namespace Benchmarks.Data
         private readonly IRandom _random;
         private readonly ApplicationDbContext _dbContext;
         private readonly bool _useBatchUpdate;
-
+        
         public EfDb(IRandom random, ApplicationDbContext dbContext, IOptions<AppSettings> appSettings)
         {
             _random = random;
             _dbContext = dbContext;
-            _dbContext.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;
             _useBatchUpdate = appSettings.Value.Database != DatabaseServer.PostgreSql;
         }
+
+        private static readonly Func<ApplicationDbContext, int, Task<World>> _firstWorldQuery
+            = EF.CompileAsyncQuery((ApplicationDbContext context, int id) 
+                => context.World.First(w => w.Id == id));
 
         public Task<World> LoadSingleQueryRow()
         {
             var id = _random.Next(1, 10001);
-            return _dbContext.World.FirstAsync(w => w.Id == id);
+
+            return _firstWorldQuery(_dbContext, id);
         }
 
         public async Task<World[]> LoadMultipleQueriesRows(int count)
@@ -36,7 +43,8 @@ namespace Benchmarks.Data
             for (var i = 0; i < count; i++)
             {
                 var id = _random.Next(1, 10001);
-                result[i] = await _dbContext.World.FirstAsync(w => w.Id == id);
+
+                result[i] = await _firstWorldQuery(_dbContext, id);
             }
 
             return result;
@@ -49,9 +57,10 @@ namespace Benchmarks.Data
             for (var i = 0; i < count; i++)
             {
                 var id = _random.Next(1, 10001);
-                var result = await _dbContext.World.AsTracking().FirstAsync(w => w.Id == id);
+                var result = await _firstWorldQuery(_dbContext, id);
 
-                result.RandomNumber = _random.Next(1, 10001);
+                _dbContext.Attach(result).Property("RandomNumber").CurrentValue = _random.Next(1, 10001);
+
                 results[i] = result;
 
                 if (!_useBatchUpdate)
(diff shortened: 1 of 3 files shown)


mikeharder:
I will test the perf impact of this change.  Should a similar change be made to the Fortunes query?

anpete:
Updated the Fortunes query.

mikeharder:
I benchmarked this PR on Azure F4 VMs.  Results:

Scenario | Baseline | PR | Improvement
------------ | ------------- | ------------ | -------------
Single Query (/db/ef) | 9,103 | 16,247 | 78%
Multiple Queries (/queries/ef?queries=20) | 730 | 1,157 | 58%
Multiple Updates (/updates/ef?queries=20) | 564 | 1,157 | Large number of non-2XX responses
Fortunes (/fortunes/ef) | 8,809 | 11,577 | 31%

This PR causes large improvements to most scenarios.  However, before this PR is merged the non-2XX responses for the "multiple updates" scenario should be understood.  This PR causes the number of non-2XX responses to increase from a handful (2-3) to thousands.  The baseline itself should not have any errors either, so it's possible the baseline is itself flawed, and this PR just amplifies the existing problem.

My next step will be investigating the non-2XX responses in both the baseline and this PR.

mikeharder:
Here's the exception *with* the PR:

```
System.InvalidOperationException: The instance of entity type 'World' cannot be tracked because another instance of this type with the same key is already being tracked. When adding new entities, for most key types a unique temporary key value will be created if no key is set (i.e. if the key property is assigned the default value for its type). If you are explicitly setting key values for new entities, ensure they do not collide with existing entities or temporary values generated for other new entities. When attaching existing entities, ensure that only one entity instance with a given key value is attached to the context.
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Add(TKey key, InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.StartTracking(InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState oldState, EntityState newState, Boolean acceptChanges)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityGraphAttacher.PaintAction(EntityEntryGraphNode node)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityEntryGraphIterator.TraverseGraph(EntityEntryGraphNode node, Func`2 handleNode)
   at Microsoft.EntityFrameworkCore.DbContext.SetEntityState[TEntity](TEntity entity, EntityState entityState)
   at Benchmarks.Data.EfDb.<LoadMultipleUpdatesRows>d__7.MoveNext() in C:\Git\Benchmarks\src\Benchmarks\Data\EfDb.cs:line 62
```

I'm still trying to repro the exception without the PR.

mikeharder:
I've seen two exceptions thrown in the dev branch (*without* this PR):

* updates/ef: An item with the same key has already been added. Key: Benchmarks.Data.EfDb (https://github.com/aspnet/benchmarks/issues/162)
* updates/ef: Invalid operation. The connection is closed. (https://github.com/aspnet/benchmarks/issues/163)

The first is likely a bug in the service initialization code, and happens every time after the server is started.  The second happens rarely, and probably needs to be caught and retried somewhere.

benaadams:
Shouldn't be ChangeTracking should it?

src/Benchmarks/Data/EfDb.cs
```diff
@@ -49,9 +57,10 @@ public Task<World> LoadSingleQueryRow()
             for (var i = 0; i < count; i++)
             {
                 var id = _random.Next(1, 10001);
-                var result = await _dbContext.World.AsTracking().FirstAsync(w => w.Id == id);
+                var result = await _firstWorldQuery(_dbContext, id);
+
+                _dbContext.Attach(result).Property("RandomNumber").CurrentValue = _random.Next(1, 10001);
```
divega: Attach will fail any time _random.Next(1, 10001) happen to returns return the same value twice in the same call to `LoadMultipleUpdatesRows()`

divega: I think the best solution of this is to use `Find()` and to fix `Find()` to take advantage of compiled queries (I have created https://github.com/aspnet/EntityFramework/issues/7248 for this). Once we do this code can be changed to:

``` 
C#
var id = _random.Next(1, 10001);
var result = await _dbContext.World.Find(id);
var entry = _dbContext.Entry(result);
if (entry.State == EntityState.Detached)
{
    entry.State = EntityState.Unchanged;
}
entry.Property("RandomNumber").CurrentValue = _random.Next(1, 10001);
```

divega: In the meanwhile I cannot think of a way to do this that takes advantage of compiled queries and doesn't make the code much uglier

cc @anpete @ajcvickers @mikeharder

divega: Although after talking to @ajcvickers, I think there is a way if we make this particular query tracked. I will put that together.

divega: @mikeharder I think this should work for now, can you try it?:

1. Change the compiled query declared in the class to be tracked explicitly:
``` C#
private static readonly Func<ApplicationDbContext, int, Task<World>> _firstWorldQuery 
    = EF.CompileAsyncQuery((ApplicationDbContext context, int id)  
        => context.World.AsTracking().First(w => w.Id == id)); 
```

2. Then change the code inside the loop to do:
 ``` C#
var id = _random.Next(1, 10001); 
var result = await _firstWorldQuery(_dbContext, id); 
_dbContext.Entry(result).Property("RandomNumber").CurrentValue = _random.Next(1, 10001); 
...
```

mikeharder:
Blocked on https://github.com/aspnet/EntityFramework/issues/7251

anpete:
Fix coming https://github.com/aspnet/EntityFramework/pull/7473

anpete:
Created a new PR, closing.

PR #166, opened 2016-12-17
Fast-path plaintext response headers

benaadams:
If you are fast-pathing `Content-Length` might as well go all in... and really "fast path"!

Even inlines the header setters 😝 

/cc @mikeharder @nathana1 @DamianEdwards


DamianEdwards:
We're not going to use internal APIs in Kestrel for the benchmarks, but appreciate the intent :smile:

PR #168, opened 2017-01-11
Parse and bind to IP's configured via --server.urls

halter73:
- Will revisit when https://github.com/aspnet/KestrelHttpServer/issues/1290
  is addressed

@mikeharder

### DIFF
--- a/src/Benchmarks/Program.cs
+++ b/src/Benchmarks/Program.cs
@@ -171,8 +171,9 @@ namespace Benchmarks
         private static void Listen(KestrelServerOptions options, IConfigurationRoot config, string url)
         {
             var uri = new Uri(url);
+            var endpoint =  CreateIPEndPoint(uri);
             
-            options.Listen(IPAddress.Loopback, uri.Port, listenOptions =>
+            options.Listen(endpoint, listenOptions =>
             {
                 var connectionFilter = GetConnectionFilter(config);
                 if (connectionFilter != null)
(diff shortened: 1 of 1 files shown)


src/Benchmarks/Program.cs
```diff
@@ -186,5 +187,21 @@ private static void Listen(KestrelServerOptions options, IConfigurationRoot conf
...
             });
         }
+
+        private static IPEndPoint CreateIPEndPoint(Uri uri)
+        {
+            IPAddress ip;
+
+            if (string.Equals(uri.Host, "localhost", StringComparison.OrdinalIgnoreCase))
+            {
+                ip = IPAddress.Loopback;
```
mikeharder: Technically I think we should bind to both `IPAddress.Loopback` and `IPAddress.IPv6Loopback`, like we do at https://github.com/aspnet/KestrelHttpServer/blob/2351c1b558fc6f4312af262aa866fd2c5164761f/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs#L230.

halter73: What if the port is also 0? Do we add logic to throw in that case like we do Kestrel? It also complicates things slightly (but not too much) to have multiple IPEndpoints per URL.

In practice IPv4 loopback always seems to work when "localhost" is specified, but I'll change it if it's important

mikeharder: What you have now should be sufficient to unblock most scenarios.  We should revisit this after the final design though.

Issue #165, opened 2016-12-15
Add scenarios for Plaintext and JSON with WebListener


mikeharder:

dougbu:
Dupe of #165

dougbu:
f7b0c33 description mentions this bug. But, #112 contains a more complete description of the work done.

PR #171, opened 2017-01-20
Add `--webHost` options

dougbu:
- #165
- new choice between Kestrel and WebListener server jobs

Note: run BenchmarksServer as Admin
- WebListener scnearios need this to listen anywhere but localhost
- port registration alternative conflicts with dynamic http / https switch
  - e.g. `netsh http add urlacl url=http://garnetc:5000 user=everyone`


src/Benchmarks.ServerJob/WebHost.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Benchmarks.ServerJob
+{
+    public enum WebHost
+    {
+        Kestrel,
+        WebListener,
```
mikeharder: Since `Microsoft.AspNetCore.Server.WebListener` has been renamed to `Mcrosoft.AspNetCore.Server.HttpSys`, I think we should replace `WebListener` with `HttpSys` in your new code and also Benchmarks itself.

Rename announcement: https://github.com/aspnet/Announcements/issues/211

src/BenchmarksDriver/Program.cs
```diff
@@ -87,6 +87,10 @@ public static int Main(string[] args)
                 "Source dependency. Format is 'repo@branchOrCommit'. " +
                 "Repo can be a full URL, or a short name under https://github.com/aspnet.",
                 CommandOptionType.MultipleValue);
+            var webHostOption = app.Option(
```
mikeharder: It looks like the short option `-w` is still free, so I would add it to `--webHost`.

mikeharder:
I've added the new column to the prod SQL database.  Once this change is merged and we have some rows with WebHost=HttpSys, we can update the PowerBI.

dougbu:
:up::date:

@mikeharder I also added fix for a race hit while I was testing. See the last commit if you want it in isolation&hellip;

dougbu:
:up::date: again. Added a few Linux workarounds.

dougbu:
:Up::date:

dougbu:
f7b0c33

Issue #169, opened 2017-01-18
Add response caching middleware scenario that bypass the middleware


JunTaoLuo:
We used to rely on the scenario RequestNoCache to keep track of the performance when the middleware is bypassed. A no-cache on the request shouldn't bypass the middleware, it should just skip the cache look up and has been addressed in https://github.com/aspnet/ResponseCaching/issues/83. 

We need a new scenario that sends a POST request for example.

ferantivero:
aspnet/KestrelHttpServer#1319

PR #172, opened 2017-01-20
Add a scenario where ReponseCaching Middleware no-ops

JunTaoLuo:
#169 

We skip the middleware for unsafe methods so I'm adding a scenario where a DELETE request is sent to the same endpoint that is used by plaintext cached. 

I needed to update the pipeline.lua script to handle specific methods but I think we don't need to update the SQL update query to add Method into the database since the scenario name should be sufficient.

### DIFF
--- a/scripts/pipeline.lua
+++ b/scripts/pipeline.lua
@@ -1,21 +1,27 @@
 local pipelineDepth = 1
 local counter = 0
 local maxRequests = -1
+local method = "GET"
 
 function init(args)
+
    if args[1] ~= nil then
       pipelineDepth = tonumber(args[1])
    end
 
+   if args[2] ~= nil then
+      method = args[2]
+   end
+
    local r = {}
    for i = 1, pipelineDepth, 1 do
-      r[i] = wrk.format(nil)
+      r[i] = wrk.format(method)
    end
 
    print("Pipeline depth: " .. pipelineDepth)
 
-   if args[2] ~= nil then
-      maxRequests = tonumber(args[2])
+   if args[3] ~= nil then
+      maxRequests = tonumber(args[3])
       print("Max requests: " .. maxRequests)
    end
 
--- a/src/BenchmarksClient/Startup.cs
+++ b/src/BenchmarksClient/Startup.cs
@@ -97,7 +97,7 @@ namespace BenchmarkClient
                 if (job != null)
                 {
                     var jobLogText = $"[ID:{job.Id} Connections:{job.Connections} Threads:{job.Threads} " +
-                        $"Duration:{job.Duration} Pipeline:{job.PipelineDepth}";
+                        $"Duration:{job.Duration} Pipeline:{job.PipelineDepth} Method:{job.Method}";
 
                     if (job.Headers != null)
                     {
@@ -161,6 +161,11 @@ namespace BenchmarkClient
             if (job.PipelineDepth > 0)
             {
                 command += $" -- {job.PipelineDepth}";
+
+                if (job.Method != "GET")
+                {
+                    command += $" {job.Method}";
+                }
             }
 
             var process = new Process()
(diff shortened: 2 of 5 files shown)


JunTaoLuo:
ping @mikeharder

JunTaoLuo:
:up::date:

JunTaoLuo:
ping @mikeharder

scripts/pipeline.lua
```diff
@@ -1,21 +1,27 @@
 local pipelineDepth = 1
 local counter = 0
 local maxRequests = -1
+local method = "GET"
```
mikeharder: I think we should create a separate LUA script for different HTTP methods, rather than a single LUA script for both pipelining and different methods.  We currently don't have any scenarios which need both pipelining and a different HTTP method, and I don't forsee a need for this any time soon.  Keeping these scripts separate is simpler and prevents invalidating the previous results with pipeline.lua or no script.

src/BenchmarksClient/Startup.cs
```diff
@@ -151,16 +151,11 @@ private static Process StartProcess(ClientJob job)
                 }
             }
 
-            if (job.PipelineDepth > 0)
-            {
-                command += $" -s scripts/pipeline.lua";
-            }
-
-            command += $" {job.ServerBenchmarkUri}";
+            command += $" -s scripts/pipeline.lua {job.ServerBenchmarkUri} -- {job.Method}";
```
mikeharder: Once we have a separate script for non-GET methods, it should be used instead of `pipeline.lua` but only when `job.Method != "GET"`

scripts/pipeline.lua
```diff
@@ -1,21 +1,27 @@
...
 
 function init(args)
+
```
mikeharder: I think we should create a separate LUA script for different HTTP methods, rather than a single LUA script for both pipelining and different methods. We currently don't have any scenarios which need both pipelining and a different HTTP method, and I don't forsee a need for this any time soon. Keeping these scripts separate is simpler and prevents invalidating the previous results with pipeline.lua or no script.

src/BenchmarksDriver/Program.cs
```diff
@@ -99,6 +99,8 @@ public static int Main(string[] args)
                 "Number of threads used by client", CommandOptionType.SingleValue);
             var headerOption = app.Option("--header",
                 "Header added to request", CommandOptionType.MultipleValue);
+            var methodOption = app.Option("-p|--method",
```
mikeharder: I would remove the short option `-p`.  There are now too many options to choose reasonable single-letter shortcuts, so we might as well stop.

src/BenchmarksDriver/Program.cs
```diff
@@ -363,6 +370,7 @@ public static int Main(string[] args)
                         [Duration] [int] NOT NULL,
                         [PipelineDepth] [int] NULL,
                         [Headers] [nvarchar](max) NULL,
+                        [Method] [nvarchar](max) NULL,
```
mikeharder: `Method` column should be `NOT NULL`.

scripts/pipeline.lua
JunTaoLuo: Spoke offline, the new scenario needs both pipelining and methods. Will keep the changes to the script but will change when the script is used by wrk.

src/BenchmarksClient/Startup.cs
JunTaoLuo: Right now we don't have any scenarios that need a different method and no pipelining so I'm not going to create another lua script for it and rely on the modifications to the pipeline.lua script. If we eventually do have a scenario like this, we'll need to create a different script and tweak this logic.

src/BenchmarksDriver/Program.cs
```diff
@@ -363,6 +370,7 @@ public static int Main(string[] args)
                         [Duration] [int] NOT NULL,
                         [PipelineDepth] [int] NULL,
                         [Headers] [nvarchar](max) NULL,
+                        [Method] [nvarchar](max) NOT NULL,
```
mikeharder: Nit: I would move this above `Headers`, since the method comes before headers in an HTTP request.

JunTaoLuo:
:up::date: rebased on latest dev

PR #174, opened 2017-01-25
Float minor and patch versions of third-party dependencies

mikeharder:


src/Benchmarks/project.json
```diff
@@ -29,8 +29,8 @@
     "Microsoft.Extensions.Configuration.Json": "1.2.0-*",
     "Microsoft.Extensions.Configuration.CommandLine": "1.2.0-*",
     "Microsoft.Extensions.Configuration.EnvironmentVariables": "1.2.0-*",
-    "Npgsql": "3.2.0-*",
-    "Npgsql.EntityFrameworkCore.PostgreSQL": "1.0.2-*"
+    "Npgsql": "3.2.0-beta1",
+    "Npgsql.EntityFrameworkCore.PostgreSQL": "1.1.0"
```
dougbu: Curious: What do `3.2.0-*` and `1.1.0-*` download from NuGet.org? Wondering whether we should continue to let these float.

mikeharder: There are tradeoffs, but I agree it's probably better to float third-party packages as well so we're always testing the latest version.  I'll update this PR to do so.

mikeharder:
@dougbu: updated

PR #177, opened 2017-01-26
Support running driver and server on Linux

dougbu:
Support BenchmarksServer on Linux
- use correct path separator everywhere
- fail HttpSys jobs immediately on Linux
- only use `taskkill.exe` on Windows (`pkill` on Linux)

Enable benchmark-tests with everything on Linux
- allow HttpSys jobs to fail without interrupting `benchmark-tests` target
- also don't make `dotnet run` verbose `--quiet` target used

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -3,6 +3,8 @@ var AUTHORS='Microsoft Open Technologies, Inc.'
 use-standard-lifecycle
 k-standard-goals
 
+default DOTNET_ARGUMENT='-v '
+
 #build-compile target='compile'
   @{
     var projectFiles = Files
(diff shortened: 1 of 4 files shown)


src/BenchmarksServer/Startup.cs
```diff
@@ -112,6 +114,14 @@ public static int Main(string[] args)
                         // TODO: Race condition if DELETE is called during this code
                         try
                         {
+                            if (!_isWindows && job.WebHost != WebHost.Kestrel)
```
dougbu: This part can't go in before #171.

makefile.shade
```diff
@@ -33,8 +38,26 @@ k-standard-goals
       "-n ResponseCachingPlaintextRequestNoCache",
       "-n ResponseCachingPlaintextVaryByCached",
     };
+
+    // Expect failures with HttpSys if BenchmarksServer is running on Linux. Never expect two failures in a row.
```
mikeharder: Ignoring failures might cause us to miss real issues.  Another option would be to add a new `ServerState` named something like `NotSupported`, then update `BenchmarksServer` and `BenchmarksDriver` to flow the `NotSupported` state rather than `Failed`.

dougbu:
:up::date:

dougbu:
3ec24f5

Issue #181, opened 2017-02-02
Upgrade to VS 2017


dougbu:
We need to upgrade our repos to VS2017. That is, we need to migrate from project.json files to MSBuild.

This is part of the https://github.com/orgs/aspnet/projects/1 umbrella. See also related work in aspnet/Universe#469.

PR #182, opened 2017-02-03
Upgrade to VS 2017 (#181)

cesarbs:
#181 

TODO:

- [ ] ~~Fix benchmarking against custom branches. Currently only `-o Benchmarks@<some branch>` works.~~ #185

Everything appears to be working correctly, with the above exception.

@mikeharder @natemcmaster @davidfowl @halter73 @pakrym @dougbu


src/BenchmarksServer/Startup.cs
```diff
@@ -289,7 +273,7 @@ private static void DeleteDir(string path)
         private static Process StartProcess(string hostname, string benchmarksRepo, ServerJob job)
         {
             var filename = "dotnet";
-            var arguments = $"run -c Release -- --scenarios {job.Scenario} --server {job.WebHost} " +
+            var arguments = $"run -c Release -f netcoreapp1.1 -- --scenarios {job.Scenario} --server {job.WebHost} " +
```
cesarbs: We've been using the default picked by CLI. Add option to select framework?

experimental/BasicAsyncHttpServer/BasicAsyncHttpServer.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <TargetFramework>net451</TargetFramework>
+    <!-- TODO remove rid when https://github.com/dotnet/sdk/issues/396 is resolved -->
+    <RuntimeIdentifier Condition="'$(TargetFramework)'!='netcoreapp1.1'">win7-x64</RuntimeIdentifier>
```
natemcmaster: (affects multiple files). As it only targets net451, you can remove the condition.

experimental/BasicAsyncHttpServer/BasicAsyncHttpServer.csproj
```diff
@@ -0,0 +1,13 @@
...
+    <WarningsAsErrors>true</WarningsAsErrors>
```
natemcmaster: (affects all csproj files in PR). Should be TreatWarningsAsErrors. Also, you shouldn't need it. By referencing Internal.AspNetCore.Sdk, this is on by default.

makefile.shade
```diff
@@ -5,16 +5,6 @@ k-standard-goals
 
 default DOTNET_ARGUMENT='-v '
 
-#build-compile target='compile'
```
dougbu: What are the implications of losing this target?

dougbu: Also, why does this need to be removed?

src/Benchmarks/Benchmarks.csproj
```diff
@@ -0,0 +1,45 @@
...
+    <PackageReference Include="Microsoft.AspNetCore.StaticFiles" Version="1.2.0-*" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="1.2.0-*" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="1.2.0-*" />
+    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="1.2.0-*" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.CommandLine" Version="1.2.0-*" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="1.2.0-*" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="1.2.0-*" />
+    <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="1.2.0-*" />
+    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="1.2.0-*" PrivateAssets="All" />
+    <PackageReference Include="Npgsql" Version="3.2.0-*" />
```
dougbu: Copy the comment about Npgsql dependency versions from the project.json file here.

src/BenchmarksServer/Startup.cs
dougbu: This particular change is a workaround for an extra / unused target framework in the Benchmarks project. Don't do this in this PR; either add that option or remove net451 from Benchmarks.csproj.

src/Benchmarks/runtimeconfig.template.json
```diff
@@ -0,0 +1,5 @@
+{
```
dougbu: Shouldn't BenchmarksClient and BenchmarksServer have similar files?

src/BenchmarksServer/Startup.cs
```diff
@@ -211,28 +211,12 @@ private static string CloneAndRestore(string path, ServerJob job)
 
             Debug.Assert(benchmarksDir != null);
 
-            // Modify all global.json to reference source dirs
```
dougbu: Are we :ok: immediately losing the ability to use older branches? E.g. could temporarily make this conditional on the global.json file existing. Remove it a week or three down the road.

halter73: Could we use @pakrym's powershell script to modify the csproj's package references to make them package references?

makefile.shade
cesarbs: Korebuild already builds the projects in `src`, and the stuff under `experimental` is not necessary and can be built by whoever wants to work with it. Also, @natemcmaster is trying to change Korebuild to build the solution file instead of enumerating projects like we do now, which will eventually build the projects in `experimental` as part of the full build again.

We want to remove as much as possible from makefiles.

src/Benchmarks/runtimeconfig.template.json
cesarbs: Those just process jobs - `BenchmarksServer`'s jobs clone the repo and run `Benchmarks`, and `BenchmarkClient`'s jobs invoke `wrk` to generate load. They aren't doing any critical work in terms benchmarking per se.

makefile.shade
dougbu: > We want to remove as much as possible from makefiles.

We do not want experimental projects built by default, ever. The idea here is to provide an _additional_ target that handles those builds when needed. Please restore this target.

src/Benchmarks/runtimeconfig.template.json
dougbu: This isn't about benchmarking. There's no earthly reason to have a long-running web project (which both BenchmarkClients and BenchmarksServer are) use regular GC. That's why we have this specified in web samples.

makefile.shade
cesarbs: Who uses the stuff under `experimental`? One of the projects (`BasicAsyncHttpServer`) is actually empty.

cesarbs: And the `build-compile` target was actually replacing the default target defined in `_k-standard-goals.shade`, meaning the `experimental` projects were being built by default.

src/Benchmarks/Benchmarks.csproj
cesarbs: Done.

src/Benchmarks/Benchmarks.csproj
```diff
@@ -0,0 +1,49 @@
...
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp1.1;net451</TargetFrameworks>
+    <!-- TODO remove rid when https://github.com/dotnet/sdk/issues/396 is resolved -->
+    <RuntimeIdentifier Condition="'$(TargetFramework)'!='netcoreapp1.1'">win7-x64</RuntimeIdentifier>
+    <PreserveCompilationContext>true</PreserveCompilationContext>
+    <OutputType>Exe</OutputType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Content Update="appsettings.json;testCert.pfx;wwwroot\**\*;Views\**\*" CopyToOutputDirectory="PreserveNewest" />
```
mikeharder: I think these should be split into mulitple lines to improve readability and make it easier to add/remove items.

src/Benchmarks/runtimeconfig.template.json
mikeharder: I agree with @dougbu, but I think we should fix this in a separate PR.  Issue: https://github.com/aspnet/benchmarks/issues/184

src/BenchmarksServer/Startup.cs
mikeharder: Created a separate issue to update BenchmarksServer source dependencies feature for MSBuild: https://github.com/aspnet/benchmarks/issues/185

src/BenchmarksServer/Startup.cs
mikeharder: I'm confused.  Is it required to add `netcoreapp1.1` to `dotnet run` as part of the MSBuild conversion?

cesarbs: Yes, if the project targets multiple TFMs. It doesn't pick one up by default anymore.

davidfowl:
👏

Issue #185, opened 2017-02-08
Update BenchmarksServer source dependencies feature for MSBuild


mikeharder:
Change logic to modify `*.csproj` files instead of `global.json`.

PR #188, opened 2017-02-14
Restore source dependencies feature (#185)

cesarbs:
#185

Couldn't make it work reliably by replacing PackageReferences with ProjectReferences (works for direct dependencies but not transitive ones, at least not every time). The solution I found was to build the cloned repos and update `Benchmarks.csproj` to reference the locally built packages. The obvious and unfortunate side-effect of this is that benchmark jobs take longer to run with source dependencies.

Since repos can have references among themselves, each repo build needs to be preceeded by an update to its own `.csproj` files to reference previously built packages. Right now dependency order is resolved by the user i.e. repo builds happen in the order `-o` arguments are passed to `BenchmarksDriver`. Given we're familiar (or it's easy to find someone familiar) with the dependencies among our repos, this should be OK for now, but ideally this shouldn't be dependent on command-line argument order.

Right now package references to locally built packages are added without checking if those packages are actually references by the Benchmarks app. A further improvement would be to do a preliminary restore to figure out which packages really need to be referenced by inspecting `project.assets.json`.

TODO:

- [ ] Make `BenchmarksServer` work on Linux/macOS

### DIFF
--- a/src/BenchmarksServer/Startup.cs
+++ b/src/BenchmarksServer/Startup.cs
@@ -219,7 +219,7 @@ namespace BenchmarkServer
                 UpdateDependencies(path, dir, builtPackages);
 
                 ProcessUtil.Run("dotnet", "restore", workingDirectory: Path.Combine(path, dir));
-                ProcessUtil.Run("cmd", "/c build.cmd build-pack", workingDirectory: Path.Combine(path, dir),
+                ProcessUtil.Run("cmd", "/c build.cmd --for-ci build-pack", workingDirectory: Path.Combine(path, dir),
                     environmentVariables: new Dictionary<string, string>
                     {
                         ["KOREBUILD_SKIP_RUNTIME_INSTALL"] = "1"
--- a/src/BenchmarksServer/Startup.cs
+++ b/src/BenchmarksServer/Startup.cs
@@ -10,12 +10,14 @@ using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Xml;
+using System.Xml.Linq;
 using Benchmarks.ServerJob;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.CommandLineUtils;
 using Microsoft.Extensions.DependencyInjection;
-using Newtonsoft.Json;
+using NuGet.Packaging;
 using Repository;
 
 namespace BenchmarkServer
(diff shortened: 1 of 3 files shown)


dsplaisted:
> Couldn't make it work reliably by replacing PackageReferences with ProjectReferences (works for direct dependencies but not transitive ones, at least not every time).

I don't have all the context here but you shouldn't need to replace the `PackageReferences`, you should be able to just add `ProjectReferences` to to what you want, and if there's a (direct or transitive) package reference to something that you also have a `ProjectReference` for, it will resolve to that project instead of to a NuGet package.

cesarbs:
@dsplaisted That doesn't seem to work well when another package also has a transitive dependency to a package that I'm trying to replace with sources [0]. In those cases I either get a restore error (version conflict) or it resolves to either a project or package dependency (it was very inconsistent when I was playing with it).

@natemcmaster suggested that I try using `ExcludeAssets="All"` to prevent version conflicts. I did a quick test while I was still in the office and that seems to work and would lead to a much better solution to what we're trying to do here. I'm working on an alternative PR to this issue that will use that.

[0] a quick way to see this is to create a new project, clone our https://github.com/aspnet/Configuration repo and add project references to the projects in that repo. That works fine. Then add a package reference to Microsoft.AspNetCore.Mvc (which references Configuration packages) and restore. That will lead to version conflict errors.

src/BenchmarksServer/Startup.cs
```diff
@@ -211,18 +210,126 @@ private static string CloneAndRestore(string path, ServerJob job)
...
 
-            // Restore in each dir
-            foreach (var dir in dirs)
+            foreach (var dir in dirs.Except(new[] { benchmarksDir }))
             {
+                UpdateNuGetConfig(path, dir, builtDirs);
+                UpdateDependencies(path, dir, builtPackages);
+
                 ProcessUtil.Run("dotnet", "restore", workingDirectory: Path.Combine(path, dir));
+                ProcessUtil.Run("cmd", "/c build.cmd build-pack", workingDirectory: Path.Combine(path, dir),
```
halter73: Is this and the backslashes the only things we know of preventing this from working on mac and linux? Is that going to be fixed before merging?

cesarbs: Should be, thought I don't know if we've ever run BenchmarksServer on non-Windows.

cesarbs:
Another example where it doesn't work with `ProjectReference`:

The `Benchmarks` project in this repo depends on Kestrel, which is turns depends on Hosting and HttpAbstractions. If I want to replace just those two transitive dependencies with sources, they work individually but not together i.e. if I add `ProjectReference`s to Hosting projects *only*, those are resolved to project references, and likewise if I add `ProjectReference`s to HttpAbstractions projects *only*, those are also resolved to project references. But if I add `ProjectReference`s to both Hosting and HttpAbstractions projects, the Hosting dependencies are resolved to project references but the HttpAbstractions ones are still resolved as packages.

src/BenchmarksServer/Startup.cs
halter73: This is building the package in debug mode which can hurt perf greatly. 3x in the case of Kestrel. @pranavkm says you can change this to "build.cmd --for-ci build-pack?

cesarbs: Fixed.

halter73:
Close this?

PR #190, opened 2017-02-15
Restore source dependencies feature (#185)

cesarbs:
Better alternative to #188. Uses `ProjectReference` and does not require intermediate builds.

Thanks @dsplaisted for pointing out that `restore /p:VersionSuffix=` does the trick! 😄 

#185

### DIFF
--- a/src/BenchmarksServer/Startup.cs
+++ b/src/BenchmarksServer/Startup.cs
@@ -10,12 +10,13 @@ using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Xml;
+using System.Xml.Linq;
 using Benchmarks.ServerJob;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.CommandLineUtils;
 using Microsoft.Extensions.DependencyInjection;
-using Newtonsoft.Json;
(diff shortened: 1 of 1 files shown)


src/BenchmarksServer/Startup.cs
```diff
@@ -211,16 +212,57 @@ private static string CloneAndRestore(string path, ServerJob job)
 
             Debug.Assert(benchmarksDir != null);
 
-            // TODO: add project references to checked out repos
-            // https://github.com/aspnet/benchmarks/issues/185
+            AddSourceDependencies(path, benchmarksDir, dirs);
 
-            // Restore in each dir
-            foreach (var dir in dirs)
+            ProcessUtil.Run("dotnet", "restore /p:VersionSuffix=zzzzz-99999", workingDirectory: Path.Combine(path, benchmarksDir));
```
cesarbs: The version suffix needs to be higher than whatever package versions we have in our feed. Hence this thing. I tried coming up with something whimsical but decided to go with this boring suffix 😛

dsplaisted:
It makes me a bit uncomfortable to override the `VersionSuffix` for restore but not for build.  I don't think it will affect anything, though, so it's probably fine.

src/BenchmarksServer/Startup.cs
halter73: I think this deserves a comment.

PR #193, opened 2017-02-16
Replace "dotnet run" with "dotnet Benchmarks.dll"

mikeharder:
It is recommended to use "dotnet foo.dll" rather than "dotnet run" for production scenarios.  Should not impact throughput.

### DIFF
--- a/src/BenchmarksServer/Startup.cs
+++ b/src/BenchmarksServer/Startup.cs
@@ -264,7 +264,8 @@ namespace BenchmarkServer
             benchmarksProjectDocument.Root.Add(netFrameworkReferences);
             benchmarksProjectDocument.Root.Add(netCoreReferences);
 
-            using (var writer = XmlWriter.Create(benchmarksProjectPath, new XmlWriterSettings { Indent = true, OmitXmlDeclaration = true }))
+            using (var stream = File.OpenWrite(benchmarksProjectPath))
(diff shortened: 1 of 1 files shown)


src/BenchmarksServer/Startup.cs
```diff
@@ -262,7 +266,8 @@ private static void AddSourceDependencies(string path, string benchmarksDir, IEn
             benchmarksProjectDocument.Root.Add(netFrameworkReferences);
             benchmarksProjectDocument.Root.Add(netCoreReferences);
 
-            using (var writer = XmlWriter.Create(benchmarksProjectPath, new XmlWriterSettings { Indent = true, OmitXmlDeclaration = true }))
+            using (var stream = File.OpenWrite(benchmarksProjectPath))
+            using (var writer = XmlWriter.Create(stream, new XmlWriterSettings { Indent = true, OmitXmlDeclaration = true }))
```
dougbu: Curious: Why did you need to manage the underlying stream separately? E.g. does the other `Create()` overload open the file strangely?

cesarbs: We don't have the former API now that we've reverted to a previous version of NETStandard.Library.

mikeharder: The `Create(string)` overloads only exist in .NET Framework and newer versions of `NETStandard.Library`.  We could go back to the `Create(string)` overload once we move to a newer version of `NETStandard.Library`.

PR #195, opened 2017-02-17
Add option for Kestrel ThreadCount

mikeharder:


src/BenchmarksDriver/Program.cs
```diff
@@ -87,6 +87,9 @@ public static int Main(string[] args)
                 "Source dependency. Format is 'repo@branchOrCommit'. " +
                 "Repo can be a full URL, or a short name under https://github.com/aspnet.",
                 CommandOptionType.MultipleValue);
+            var threadCountOption = app.Option("-t|--threadCount",
+                "Kestrel ThreadCount.",
+                CommandOptionType.SingleValue);
```
dougbu: Bit confusing. Suggest we rename `--threads` to `--clientThreads` and make this one `--serverThreads` or perhaps `-c|--clientthreads` and `-s|--serverThreads`. I'd also rather we use `-t` with `--duration` but we can separate that.

src/BenchmarksDriver/Program.cs
```diff
@@ -446,6 +458,7 @@ public static int Main(string[] args)
                     p.AddWithValue("@ConnectionFilter",
                         string.IsNullOrEmpty(connectionFilter) ? (object)DBNull.Value : connectionFilter);
                     p.AddWithValue("@WebHost", webHost.ToString());
+                    p.AddWithValue("@ThreadCount", (object)threadCount ?? DBNull.Value);
```
dougbu: Life was so much better when .NET Core didn't contain `DBNull` 😈

benchmarks.sln
```diff
@@ -1,12 +1,13 @@
...
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26117.0
+VisualStudioVersion = 15.0.26215.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{995FCFF9-E5F6-4BDD-8E5B-FBDEA21145F9}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{BAF239DE-04E9-428E-A6D6-98D4C6CFB9E3}"
 	ProjectSection(SolutionItems) = preProject
+		makefile.shade = makefile.shade
```
dougbu: I wouldn't bother with this addition since `makefile.shade` shouldn't be long for this world.

src/BenchmarksServer/Startup.cs
```diff
@@ -332,6 +332,11 @@ private static Process StartProcess(string hostname, string benchmarksRepo, Serv
                 arguments += $" --connectionFilter {job.ConnectionFilter}";
             }
 
+            if (job.ThreadCount.HasValue)
+            {
+                arguments += $" --threadCount {job.ThreadCount}";
```
dougbu: Suggest `job.ThreadCount.Value` for consistency with `HasValue()` / `Value()` elsewhere in this class.

src/BenchmarksDriver/Program.cs
cesarbs: I vote for `--serverThreads`.

`-s` and `-c` are already used to inform the server and client URLs.

cesarbs:
NVM I see the option was already there in the Benchmarks app.

src/BenchmarksDriver/Program.cs
mikeharder: Since the option only applies to Kestrel (not HttpSysServer), I'm planning to rename it to `--kestrelThreadCount`.  I will also remove the short version `-t` to avoid confusion.  I'd rather leave the existing `--threads` option for wrk threads alone to minimize churn.

cesarbs: Still think you should rename `--threads` to `--clientThreads`, but that can be addressed in a separate PR.

mikeharder: OK, I will rename `--threads` to `--clientThreads` to prevent confusion.

src/BenchmarksDriver/Program.cs
```diff
@@ -411,6 +421,7 @@ public static int Main(string[] args)
                            ,[Scheme]
                            ,[ConnectionFilter]
                            ,[WebHost]
+                           ,[ThreadCount]
```
halter73: Is the plan just to clear out the old data?

src/BenchmarksDriver/Program.cs
halter73: If you want to be more future-proof with the naming you could call --serverThreads --libuvThreads instead.

src/BenchmarksDriver/Program.cs
mikeharder: Which old data?  The new column `ThreadCount` allowed null, so all existing rows have null.  If I rename the column to `KestrelThreadCount`, existing data in the column will be preserved.

src/BenchmarksDriver/Program.cs
mikeharder: The parameter ultimately maps to `KestrelServerOptions.ThreadCount`, so I think `--kestrelThreadCount` is the best name.

mikeharder:
@dougbu, @cesarbs: Updated

src/BenchmarksDriver/Program.cs
halter73: That's likely to change with https://github.com/aspnet/KestrelHttpServer/issues/828, but we deal with it when we get there.

PR #201, opened 2017-02-22
Write sources parameter to SQL

mikeharder:


cesarbs:
Is this so we can create different graphs for benchmarking different branches?

mikeharder:
@cesarbs: Yes, I am going to add `Kestrel@dev` and `Kestrel@dev-si` to the automation.

Issue #189, opened 2017-02-14
Parameters to wrk should exactly match TechEmpower


mikeharder:
The Benchmarks automation should use exactly the same `wrk` parameters as the official TechEmpower runs.  Specifically, the extra headers in the official TechEmpower reduce throughput by about 10%.

TechEmpower:
```
wrk
  -H 'Host: localhost'
  -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7'
  -H 'Connection: keep-alive'
  --latency
  -d 15
  -c <connections>
  --timeout 8
  -t 32
  http://server-ip:port/plaintext
  -s ~/pipeline.lua
  -- 16
```

Current:
```
wrk
  -d 10
  -c <connections>
  -t 32
  http://server-ip:port/plaintext
  -s ~/pipeline.lua
  -- 16
```

https://github.com/TechEmpower/FrameworkBenchmarks/blob/3a659ed7f1ef907620da9893f21a357fe811a65a/toolset/benchmark/framework_test.py
https://github.com/TechEmpower/FrameworkBenchmarks/blob/e10c3c0f817704ff8fb5d7e2bb0f9a63f97be7b4/toolset/benchmark/test_types/framework_test_type.py

PR #203, opened 2017-02-22
Parameters to wrk should exactly match TechEmpower

mikeharder:
- Addresses #189

### DIFF
--- a/src/BenchmarksClient/Startup.cs
+++ b/src/BenchmarksClient/Startup.cs
@@ -141,7 +141,7 @@ namespace BenchmarkClient
         {
             var tcs = new TaskCompletionSource<bool>();
 
-            var command = $"wrk";
+            var command = "wrk";
 
             if (job.Headers != null)
             {
@@ -151,7 +151,7 @@ namespace BenchmarkClient
                 }
             }
 
-            command += " --latency -d {job.Duration} -c {job.Connections} --timeout 8 -t {job.Threads}";
+            command += $" --latency -d {job.Duration} -c {job.Connections} --timeout 8 -t {job.Threads}";
 
             command += $" {job.ServerBenchmarkUri}";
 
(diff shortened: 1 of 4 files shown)


src/BenchmarksClient/Startup.cs
```diff
@@ -151,23 +151,22 @@ private static Process StartProcess(ClientJob job)
...
-                command += $" -- {job.PipelineDepth}";
+                command += $" -s scripts/pipeline.lua -- {job.PipelineDepth}";
 
                 if (job.Method != "GET")
                 {
                     command += $" {job.Method}";
                 }
             }
 
+            Log(command);
```
cesarbs: Decorate with something like "Running wrk with parameters:".

src/BenchmarksDriver/Program.cs
```diff
@@ -21,40 +21,54 @@ public class Program
...
         private static readonly HttpClient _httpClient = new HttpClient();
 
+        private static readonly IEnumerable<string> _commonHeaders = new string[]
+        {
+            "Host: localhost",
+            "Accept: {0}",
+            "Connection: keep-alive"
+        };
+
+        private static readonly IEnumerable<string> _plaintextHeaders =
```
JunTaoLuo: These look pretty awkward. Why not just include Host and Connection headers in _commonHeaders and make _plaintextHeaders = _commonHeaders.Append(...)

mikeharder: It may not matter, but I wanted the order of headers to match TechEmpower.

src/BenchmarksClient/Startup.cs
mikeharder: I think it's clear enough already:

```
[03:29:24.117] wrk -H "Host: localhost" ...
[03:30:18.970] Running 15s test @ http://server:port/plaintext
```

Issue #111, opened 2016-08-04
Add full .NET Framework


mikeharder:
- [x] Add parameter to Driver, Server, and ServerJob to specify .NET Framework or .NET Core
- [x] Update Server to run Benchmarks on .NET Framework or .NET Core based on parameter
- [x] Add column to SQL
- [x] Add rows to makefile.shade for the following scenarios on .NET Framework
  - Plaintext
  - Plaintext with HTTPS
  - JSON
  - JSON with HTTPS
  - MvcPlaintext
  - MvcJSON
- [x] Update PowerBI report to compare .NET Framework and .NET Core

runxc1:
Is this being worked on or is there any visibility here?   I would love to see the difference between performance of .Net Core and the Full Framework using Asp.Net Core.  Would help with some internal planning.

PR #205, opened 2017-02-23
Add desktop .NET Framework support (#111)

cesarbs:
Task list at #111

### DIFF
--- a/src/Benchmarks.ServerJob/ServerJob.cs
+++ b/src/Benchmarks.ServerJob/ServerJob.cs
@@ -43,5 +43,8 @@ namespace Benchmarks.ServerJob
 
         [JsonConverter(typeof(StringEnumConverter))]
         public WebHost WebHost { get; set; }
+
+        [JsonConverter(typeof(StringEnumConverter))]
+        public Framework Framework { get; set; }
     }
 }
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -111,6 +111,9 @@ namespace BenchmarkDriver
                 "-w|--webHost",
                 "WebHost (Kestrel or HttpSys). Default is Kestrel.",
                 CommandOptionType.SingleValue);
+            var frameworkOption = app.Option("-r|--framework",
+                "Framework (Core or Desktop). Default is Core.",
+                CommandOptionType.SingleValue);
 
             // ClientJob Options
             var connectionsOption = app.Option("--connections",
@@ -142,6 +145,12 @@ namespace BenchmarkDriver
                     webHostValue = "Kestrel";
                 }
 
+                var frameworkValue = frameworkOption.Value();
+                if (string.IsNullOrEmpty(frameworkValue))
+                {
+                    frameworkValue = "Core";
+                }
+
                 var server = serverOption.Value();
                 var client = clientOption.Value();
                 var sqlConnectionString = sqlConnectionStringOption.Value();
@@ -149,9 +158,11 @@ namespace BenchmarkDriver
                 Scheme scheme;
                 Scenario scenario;
                 WebHost webHost;
+                Framework framework;
                 if (!Enum.TryParse(schemeValue, ignoreCase: true, result: out scheme) ||
                     !Enum.TryParse(scenarioOption.Value(), ignoreCase: true, result: out scenario) ||
                     !Enum.TryParse(webHostValue, ignoreCase: true, result: out webHost) ||
+                    !Enum.TryParse(frameworkValue, ignoreCase: true, result: out framework) ||
                     string.IsNullOrWhiteSpace(server) ||
                     string.IsNullOrWhiteSpace(client))
                 {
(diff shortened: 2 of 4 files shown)


src/Benchmarks/Benchmarks.csproj
```diff
@@ -10,6 +10,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Content Include="testCert.pfx" CopyToOutputDirectory="PreserveNewest" />
+  </ItemGroup>
+
+  <ItemGroup>
```
mikeharder: Is this just code cleanup?

src/BenchmarksDriver/Program.cs
```diff
@@ -285,10 +297,21 @@ public static int Main(string[] args)
 
                 if (clientJob.State == ClientState.Completed && !string.IsNullOrWhiteSpace(sqlConnectionString))
                 {
-                    await WriteResultsToSql(sqlConnectionString, scenario, serverJob.Scheme, serverJob.Sources,
-                        serverJob.ConnectionFilter, serverJob.WebHost, serverJob.KestrelThreadCount,
-                        clientJob.Threads, clientJob.Connections, clientJob.Duration, clientJob.PipelineDepth,
-                        clientJob.Method, clientJob.Headers, clientJob.RequestsPerSecond);
+                    await WriteResultsToSql(sqlConnectionString,
```
mikeharder: Move `sqlConnectionString` to named parameter for consistency

src/Benchmarks.ServerJob/ServerJob.cs
```diff
@@ -43,5 +43,8 @@ public IEnumerable<Source> Sources
 
         [JsonConverter(typeof(StringEnumConverter))]
         public WebHost WebHost { get; set; }
+
+        [JsonConverter(typeof(StringEnumConverter))]
+        public TargetFramework TargetFramework { get; set; }
```
mikeharder: I would rename `TargetFramework` to just `Framework`, since it's shorter and it matches the command-line parameter `--framework` .

mikeharder: I would rename both the Enum and the property.

src/Benchmarks/Benchmarks.csproj
cesarbs: I have to revert that.

cesarbs:
Updated.

src/BenchmarksDriver/Program.cs
```diff
@@ -111,6 +111,9 @@ public static int Main(string[] args)
                 "-w|--webHost",
                 "WebHost (Kestrel or HttpSys). Default is Kestrel.",
                 CommandOptionType.SingleValue);
+            var targetFrameworkOption = app.Option("-r|--framework",
```
mikeharder: I think all instances of `targetFramework` should be renamed to `framework`, including the SQL column.

src/BenchmarksDriver/Program.cs
```diff
@@ -413,6 +438,7 @@ public static int Main(string[] args)
                         [Id] [int] IDENTITY(1,1) NOT NULL PRIMARY KEY,
                         [DateTime] [datetimeoffset](7) NOT NULL,
                         [Scenario] [nvarchar](max) NOT NULL,
+                        [TargetFramework] [nvarchar](max) NOT NULL,
```
dougbu: @mikeharder is `NOT NULL` what we want?

src/BenchmarksServer/Startup.cs
```diff
@@ -129,6 +129,8 @@ public static int Main(string[] args)
                             Debug.Assert(tempDir == null);
                             tempDir = GetTempDir();
 
+                            var targetFramework = job.Framework == Framework.Core ? "netcoreapp1.1" : "net451";
```
dougbu: Doesn't this dupe `GetTFM()`?

Suggest `net452` in `GetTFM()` and here (if the code must be duplicated). Same for the Benchmarks.csproj file.

Hmm, why was the Benchmarks project already compiled for desktop framework? 😺

src/BenchmarksDriver/Program.cs
mikeharder: Yes.  Right before this changed is merged to dev, I will do the following:

1. Add `Frameworks` column to existing table with `NOT NULL` and a default value of `Core`.
2. Existing rows are assigned default value.
3. Remove default value, so future inserts will be required to specify `Core` or `Desktop`.

src/BenchmarksServer/Startup.cs
cesarbs: This line can be removed. Why should we bump to `net452`?

cesarbs:
Commenting again because the conversation got hidden by the last push:

@dougbu Why should we bump Benchmarks to `net452`?

mikeharder:
I have added the SQL column `Framework` to our perf DB with default value `Core`.  After this change is merged to `dev`, I will remove the default value.

dougbu:
> Why should we bump Benchmarks to `net452`?

Because Microsoft no longer supports .NET 4.5.1 and it's not clear why we'd benchmark using that old version. Another reasonable option would be to test with the latest desktop version -- .NET 4.6.1. (We may actually want to compare NET 4.5.2, 4.6.0 and 4.6.1 but let's not get ahead of ourselves.)

cesarbs:
@dougbu I've opened https://github.com/aspnet/benchmarks/issues/206 to address that.

PR #210, opened 2017-02-27
Don't append RID to output path on full framework

cesarbs:
Breaks running on full framework. It's easier to suppress the RID than to change the code that builds the path to the executable.

### DIFF
--- a/src/Benchmarks/Benchmarks.csproj
+++ b/src/Benchmarks/Benchmarks.csproj
@@ -4,8 +4,6 @@
 
   <PropertyGroup>
     <TargetFrameworks>netcoreapp1.1;net451</TargetFrameworks>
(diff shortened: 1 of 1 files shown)


src/Benchmarks/Benchmarks.csproj
```diff
@@ -7,6 +7,7 @@
     <!-- TODO remove rid when https://github.com/dotnet/sdk/issues/396 is resolved -->
     <RuntimeIdentifier Condition="'$(TargetFramework)'!='netcoreapp1.1'">win7-x64</RuntimeIdentifier>
```
natemcmaster: Instead, consider deleting this line and the TODO before. This is the bug that was resolved.

PR #212, opened 2017-03-02
Replace makefile.shade with targets file

natemcmaster:
Removes dependency on Sake to run benchmarks


build/repo.targets
```diff
@@ -0,0 +1,60 @@
...
+    </GetDotNetHost>
+    <MSBuild Projects="$(MSBuildProjectFullPath)"
+      Targets="RunBenchmark"
+      Properties="Scenario=%(Scenarios.Identity);DotNetPath=$(DotNetPath)" />
+  </Target>
+
+  <Target Name="RunBenchmark">
+    <PropertyGroup>
+      <DotNetPath Condition="'$(DotNetPath)' == ''">dotnet</DotNetPath>
+      <RunArgs>$(DotNetPath) run -- $(Scenario) </RunArgs>
```
mikeharder: @natemcmaster: It's not related to your change, but it might be slightly faster to change this from `dotnet run` to `dotnet $(OutputPath)/BenchmarksDriver.dll`.  Is this feasible?  Is there an existing variable that would give us the output path from the previous `Compile` task?

Also, it might be more correct for `RunBenchmark` to depend on `Compile` (rather than `TestBenchmarks`).  Since if someone did directly call `RunBenchmark` once, it should ensure `BenchmarksDriver` has been compiled.

I'd suggest merging what you have now if it works, and we can make these improvements in a separate PR.

build.ps1
```diff
@@ -33,4 +33,4 @@ if (!(Test-Path $buildFolder)) {
     }
 }
 
-&"$buildFile" $args
\ No newline at end of file
+&"$buildFile" @args
```
dougbu: Question: Do we require PowerShell 3.0 somewhere? Splatting isn't available earlier.

build/repo.targets
```diff
@@ -0,0 +1,60 @@
...
+      <RunArgs>$(RunArgs) -s $(BENCHMARK_SERVER)</RunArgs>
+      <RunArgs>$(RunArgs) -c $(BENCHMARK_CLIENT)</RunArgs>
+      <RunArgs Condition="'$(BENCHMARK_SQL)' != ''">$(RunArgs) -q $(BENCHMARK_SQL)</RunArgs>
+      <RunArgs Condition="'$(Quiet)' != 'true'">$(RunArgs) -v</RunArgs>
```
natemcmaster: It's optionally set by calling `build.cmd /p:Quiet=true`

build.ps1
```diff
@@ -3,7 +3,7 @@ cd $PSScriptRoot
 $repoFolder = $PSScriptRoot
 $env:REPO_FOLDER = $repoFolder
 
-$koreBuildZip="https://github.com/aspnet/KoreBuild/archive/feature/msbuild.zip"
+$koreBuildZip="https://github.com/aspnet/KoreBuild/archive/dev.zip"
```
natemcmaster: Yes, feature/msbuild was temporary.

> why not do this as part of a pass that makes the same change everywhere

That's exactly what this is.

build/repo.targets
```diff
@@ -0,0 +1,60 @@
...
+    <Scenarios Include="-n MvcJson" />
+    <Scenarios Include="-n MvcJson -r Desktop" />
+    <Scenarios Include="-n MemoryCachePlaintext" />
+    <Scenarios Include="-n MemoryCachePlaintextSetRemove" />
+    <Scenarios Include="-n ResponseCachingPlaintextCached" />
+    <Scenarios Include="-n ResponseCachingPlaintextCached --method DELETE" />
+    <Scenarios Include="-n ResponseCachingPlaintextResponseNoCache" />
+    <Scenarios Include="-n ResponseCachingPlaintextRequestNoCache" />
+    <Scenarios Include="-n ResponseCachingPlaintextVaryByCached" />
+  </ItemGroup>
```
natemcmaster: Not sure we can do anything about that. Include is required.

natemcmaster:
Updated to invoke the DLL directly. Also, made both the "run all" and "run one" scenario depend on compile without triggering multiple builds.

build/repo.targets
```diff
@@ -0,0 +1,71 @@
+<Project>
+  <ItemGroup>
+    <Scenarios Include="-n Plaintext -o KestrelHttpServer@dev --kestrelThreadCount 1" />
+    <Scenarios Include="-n Plaintext -o KestrelHttpServer@feature/dev-si --kestrelThreadCount 2" />
```
natemcmaster: Another idea to make this easier on command line is to give these friendly names like this:
```xml
<Scenario Include="Plaintext">
  <Args>-n Plaintext -o KestrelHttpServer@feature/dev-si --kestrelThreadCount 2</Args>
</Scenario>
```

natemcmaster: usage could then be `build.cmd /t:Benchmark /p:Scenario=Plaintext`

PR #215, opened 2017-03-03
Add parameter for Kestrel ThreadPoolDispatching

mikeharder:


src/BenchmarksDriver/Program.cs
```diff
@@ -107,6 +107,9 @@ public static int Main(string[] args)
             var kestrelThreadCountOption = app.Option("--kestrelThreadCount",
                 "Maps to KestrelServerOptions.ThreadCount.",
                 CommandOptionType.SingleValue);
+            var kestrelThreadPoolDispatchingOption = app.Option("--kestrelThreadPoolDispatching",
+                "Maps to InternalKestrelServerOptions.ThreadPoolDispatching.",
```
dougbu: This should mention `InternalKestrelServerOptions.ThreadPoolDispatching` is a `bool` that defaults to `true`.

build/repo.targets
```diff
@@ -1,9 +1,13 @@
 <Project>
   <ItemGroup>
     <Scenarios Include="-n Plaintext -o KestrelHttpServer@dev --kestrelThreadCount 1" />
+    <Scenarios Include="-n Plaintext -o KestrelHttpServer@dev --kestrelThreadCount 1 --kestrelThreadPoolDispatch false" />
```
dougbu: Curious: does the thread dispatching choice matter when the thread count is `1`?

src/BenchmarksDriver/Program.cs
```diff
@@ -444,6 +454,7 @@ public static int Main(string[] args)
                         [ConnectionFilter] [nvarchar](max) NULL,
                         [WebHost] [nvarchar](max) NOT NULL,
                         [KestrelThreadCount] [int] NULL,
+                        [KestrelThreadPoolDispatching] [bool] NULL,
```
dougbu: Thought we were going to change `[KestrelThreadCount]` to `NOT NULL` after updating the main database?

Either way, might be good to do that for this new column. Would store `true` when not overridden on the command line.

src/BenchmarksDriver/Program.cs
```diff
@@ -444,6 +454,7 @@ public static int Main(string[] args)
                         [ConnectionFilter] [nvarchar](max) NULL,
                         [WebHost] [nvarchar](max) NOT NULL,
                         [KestrelThreadCount] [int] NULL,
+                        [KestrelThreadPoolDispatching] [bit] NULL,
```
dougbu: Surfacing comment that's been buried&hellip;

Thought we were going to change [KestrelThreadCount] to NOT NULL after updating the main database?

Either way, might be good to do that for this new column. Would store true when not overridden on the command line.

build/repo.targets
mikeharder: Yes, these two features are mostly unrelated.  KestrelThreadCount controls the number of libuv threads, while KestrelThreadPoolDispatch controls whether app code is run on the libuv thread or a threadpool thread.

src/BenchmarksDriver/Program.cs
mikeharder: I don't think it's necessary, since anyone setting this should know how the Kestrel feature works.  Also, if the Kestrel default changes someday I don't want to require changes to Benchmarks.

src/BenchmarksDriver/Program.cs
mikeharder: Both `KestrelThreadCount` and `KestrelThreadPoolDispatching` are nullable by design, since `null`, `true`, and `false` all have different meaning.  So the DB column should be nullable as well.

dougbu: `null` and `true` have exactly the same meaning once Kestrel starts up. But :ok:

mikeharder: I understand your point, and these could have been designed either way.  But I think it's more flexible for the benchmarks automation to allow the caller to specify `kestrel default` (`null`) in addition to `true` and `false`.  It matters less for a boolean which must default to either `true` or `false`.  It matters more for a setting like `KestrelThreadCount` where the default is something computed dynamically (`NumCores / 2`).

build/repo.targets
tmds: At the default (dispatching), Kestrel will use all cpu cores (via the ThreadPool). When set to false, Kestrel will only use kestrelThreadCount threads. The best performance when not dispatching is when kestrelThreadCount is set to the number of cpu cores (or is it cpu threads when hyperthreading?).

PR #225, opened 2017-03-14
Restore support for urls like "http://+:5000"

mikeharder:
- System.Uri doesn't support hostnames like "+", but Microsoft.AspNetCore.Server.HttpSys.UrlPrefix does.

### DIFF
--- a/src/Benchmarks/Program.cs
+++ b/src/Benchmarks/Program.cs
@@ -190,7 +190,7 @@ namespace Benchmarks
                     listenOptions.ConnectionAdapters.Add(connectionFilter);
                 }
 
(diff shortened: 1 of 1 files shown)


src/Benchmarks/Program.cs
```diff
@@ -189,27 +190,27 @@ private static void Listen(KestrelServerOptions options, IConfigurationRoot conf
                     listenOptions.ConnectionAdapters.Add(connectionFilter);
                 }
 
-                if (uri.Scheme.Equals("https", StringComparison.OrdinalIgnoreCase))
+                if (urlPrefix.Scheme.Equals("https", StringComparison.OrdinalIgnoreCase))
```
Tratcher: urlPrefix has an IsHttps property for this.

PR #227, opened 2017-03-15
Add LogLevel parameter to help diagnose failures 

mikeharder:
- No-op when LogLevel is not set to avoid any possible performance impact


BrennanConroy:
Looks reasonable

PR #228, opened 2017-03-20
Add `StaticFiles` scenarios

dougbu:
- plumb scenario from `Benchmarks` through to `BenchmarksDriver`
  - add `"/plaintext"` as a supported path for this scenario
  - reorder configuration so that `PlaintextMiddleware` is always enabled after `StaticFilesMiddleware`
- add `--path` option to differentiate potential test URLs and `[Path]` column to save the information

nit: enforce `[ScenarioPath]` usage restrictions in its constructor

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -7,6 +7,7 @@
     <Scenarios Include="-n Json -o KestrelHttpServer@dev" />
     <Scenarios Include="-n Json -o KestrelHttpServer@feature/dev-si" />
     <Scenarios Include="-n Plaintext" />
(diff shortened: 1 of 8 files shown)


dougbu:
> Could we have a scenario where each request hits a different file?

It's probably possible to craft a .lua file that can switch `wrk`'s URL between a small set. But, that's not the purpose of this PR. Please file a separate issue, especially if you'd like to see this scenario tested on an ongoing basis.

dougbu:
52570f7

cesarbs:
@dougbu Filed #230.

PR #229, opened 2017-03-21
Convert to netcoreapp2.0

mikeharder:


davidfowl:
Should we put:

```XML
<RuntimeFrameworkVersion>2.0.0-*</RuntimeFrameworkVersion>
```

In the common props?

mikeharder:
Regarding `RuntimeFrameworkVersion` in common.props, I think this version will be the same for every repo, so maybe this should automatically be added by KoreBuild?

@natemcmaster, @pranavkm, @kichalla

pranavkm:
Putting it in KoreBuild would require running something along the lines of `build initialize` from the commandline before the solution can be opened, something we (or maybe it was just me 😄 ) would prefer to avoid.

src/Benchmarks/Benchmarks.csproj
```diff
@@ -3,7 +3,8 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.1;net451</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;net451</TargetFrameworks>
+    <RuntimeFrameworkVersion>2.0.0-*</RuntimeFrameworkVersion>
```
dougbu: Should move this into build\dependencies.props. Because we don't ship these packages, there's no reason to build against lower framework versions elsewhere in this repo.

src/Benchmarks/Benchmarks.csproj
```diff
@@ -3,7 +3,8 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.1;net451</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;net451</TargetFrameworks>
```
dougbu: Is it time to be a bit more modern and use net46 instead?

src/Benchmarks/Benchmarks.csproj
dougbu: BTW I like dependencies.props better than @davidfowl's suggestion of common.props. Using dependencies.props would be more consistent with the props split in other repos.

dougbu:
> Putting it in KoreBuild would require running something along the lines of build initialize from the commandline before the solution can be opened, something we (or maybe it was just me 😄 ) would prefer to avoid.

My objection to putting the targeted framework library version in BuildTools or KoreBuild is our repos are not all the same and "big bang" upgrades are usually a mess.

But, if we nonetheless wanted it centralized, wouldn't be too hard to include the property in Internal.AspNetCore.BuildTools.Tasks.props. Those files (build and buildMultiTargeting) should be included in every project, even those that don't include the Sdk package.

src/Benchmarks/Benchmarks.csproj
davidfowl: dependencies.props has a list of dependencies doesn't it? This isn't that.

dougbu: No, dependencies.props just contains properties -- versions of dependencies.

davidfowl: And this isn't that so why would this go in there?

natemcmaster: +1 dependencies.props. RuntimeFrameworkVersion is just a bad name for MicrosoftNetCoreAppPackageVersion

davidfowl: You guys are crazy, this makes no sense to me but I don't care enough.

natemcmaster:
We can't centralize common versions to Internal.AspNetCore.Sdk yet. It's properties are imported too late, after NuGet resolves dependency versions. 

> My objection to putting the targeted framework library version in BuildTools or KoreBuild is our repos are not all the same and "big bang" upgrades are usually a mess.

+1. Keep dependency versions in the repo.

src/Benchmarks/Benchmarks.csproj
natemcmaster: > You guys are crazy, this makes no sense to me

@davidfowl  can you elaborate what doesn't make sense? These kinds of jabs make it hard to move the conversation forward.

davidfowl: There is no conversation to move forward. We we're debating common.props vs dependencies.props and decided dependencies.props makes more sense. It doesn't make sense to me because it's not a package version like the other things in `dependencies.props`. You can say it is, but it really isn't.

natemcmaster: Let me add some context the difference between common.props and dependencies.props.

The motivation for `dependencies.props` is to make updates easier through automation. Automatic version replacement on all csproj files was too unconstrained and often replaced versions for things it shouldn't have (like samples that are meant to be pinned). Constraining the automation to only files named dependnecies.props makes these errors less likely.

common.props, on the other hand, is not meant to be edited by automation. We don't always have a common.props file. The EF team, for example, prefers to call it "EntityFramework.props". MusicStore didn't need a common.props file.

Issue #87, opened 2016-05-13
Benchmark Server needs to be able to update dotnet


rynowak:
Our workload project files use wildcarding to specify their version of the SDK/runtime https://github.com/aspnet/benchmarks/blob/dev/src/Benchmarks/project.json#L25

This means that the version our projects need is dependent on our feeds. This also means that the local copy of dotnet on the server can get out of date with what we need to build/run our workload projects.

We need a solution so that every time we take a new CLI it doesn't hose the benchmark server

mikeharder:
I think we can use a method similar to the CI machine.  When running a server job:

1. Use KoreBuild to download the current CLI to a temporary directory.
2. Run dotnet from the temporary directory rather than %ProgramFiles% or %LocalAppData%.
3. Configure the NuGet package cache to use a clean temporary directory.
4. Use a different copy of dotnet.exe to run BenchmarksServer itself.

mspandit:
@rynowak, I think you mean https://github.com/aspnet/benchmarks/blob/c4648450a249a522cf8dc00942a799acaa15b312/src/Benchmarks/project.json#L25

PR #235, opened 2017-03-24
Install latest SDK and runtime before restore and build

mikeharder:
* Increases time of benchmark run from 60 to 75 seconds, but required to avoid manual intervention when SDK or runtime is updated
* Addresses #87 

CC: @cesarbs, @rynowak

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -34,8 +34,11 @@
   <PropertyGroup>
     <BenchmarksDependsOn Condition="'$(NoBuild)' != 'true'">Compile</BenchmarksDependsOn>
   </PropertyGroup>
-  
-  <!-- Used by BenchmarksServer to install latest SDK and runtime but do nothing else -->
+
+  <!--
+    BenchmarksServer calls "build /t:noop" to install the latest SDK and runtime as quickly as possible.
+    The default target would compile benchmarks.sln which takes much longer.
+  -->
   <Target Name="NoOp" />
 
   <Target Name="TestBenchmarks" DependsOnTargets="$(BenchmarksDependsOn)">
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -34,6 +34,9 @@
   <PropertyGroup>
     <BenchmarksDependsOn Condition="'$(NoBuild)' != 'true'">Compile</BenchmarksDependsOn>
   </PropertyGroup>
+  
+  <!-- Used by BenchmarksServer to install latest SDK and runtime but do nothing else -->
+  <Target Name="NoOp" />
 
   <Target Name="TestBenchmarks" DependsOnTargets="$(BenchmarksDependsOn)">
     <Error Text="BENCHMARK_SERVER must be set" Condition="'$(BENCHMARK_SERVER)' == ''" />
--- a/src/BenchmarksServer/Startup.cs
+++ b/src/BenchmarksServer/Startup.cs
@@ -235,7 +235,6 @@ namespace BenchmarkServer
 
             // Build and Restore
             var benchmarksApp = Path.Combine(benchmarksRoot, "src", "Benchmarks");
-
             var dotnetExecutable = Path.Combine(dotnetInstallDir, "dotnet");
 
             // Project versions must be higher than package versions to resolve those dependencies to project ones as expected.
--- a/src/BenchmarksServer/Startup.cs
+++ b/src/BenchmarksServer/Startup.cs
@@ -214,13 +214,18 @@ namespace BenchmarkServer
 
             AddSourceDependencies(path, benchmarksDir, dirs);
 
-            var benchmarksPath = Path.Combine(path, benchmarksDir, "src", "Benchmarks");
+            // Install latest SDK and runtime
+            var benchmarksRoot = Path.Combine(path, benchmarksDir);
+            ProcessUtil.Run("cmd", "/c build.cmd /t:noop", workingDirectory: benchmarksRoot);
+
+            // Build and Restore
+            var benchmarksApp = Path.Combine(benchmarksRoot, "src", "Benchmarks");
 
             // Project versions must be higher than package versions to resolve those dependencies to project ones as expected.
             // Passing VersionSuffix to restore will have it append that to the version of restored projects, making them
             // higher than packages references by the same name.
-            ProcessUtil.Run("dotnet", "restore /p:VersionSuffix=zzzzz-99999", workingDirectory: benchmarksPath);
-            ProcessUtil.Run("dotnet", $"build -c Release -f {GetTFM(job.Framework)}", workingDirectory: benchmarksPath);
+            ProcessUtil.Run("dotnet", "restore /p:VersionSuffix=zzzzz-99999", workingDirectory: benchmarksApp);
+            ProcessUtil.Run("dotnet", $"build -c Release -f {GetTFM(job.Framework)}", workingDirectory: benchmarksApp);
 
             return benchmarksDir;
         }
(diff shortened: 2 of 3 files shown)


build/repo.targets
```diff
@@ -37,6 +37,9 @@
   <PropertyGroup>
     <BenchmarksDependsOn Condition="'$(NoBuild)' != 'true'">Compile</BenchmarksDependsOn>
   </PropertyGroup>
+  
+  <!-- Used by BenchmarksServer to install latest SDK and runtime but do nothing else -->
+  <Target Name="NoOp" />
```
davidfowl: Explain

src/BenchmarksServer/Startup.cs
```diff
@@ -72,6 +77,8 @@ public static int Main(string[] args)
 
         private static async Task<int> Run(string url, string hostname)
         {
+            Console.WriteLine($"DOTNET_INSTALL_DIR: {_dotnetInstallDir}");
```
dougbu: Let's not check in this debug output.

src/BenchmarksServer/Startup.cs
```diff
@@ -32,6 +32,11 @@ public class Startup
         private static readonly IRepository<ServerJob> _jobs = new InMemoryRepository<ServerJob>();
         private static readonly bool _isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
 
+        private static readonly string _dotnetInstallDir = _isWindows ?
+            Path.Combine(Environment.GetEnvironmentVariable("LOCALAPPDATA"), "Microsoft", "dotnet", "BenchmarksServer") :
+            Path.Combine("~", ".dotnet", "BenchmarksServer");
+        private static readonly string _dotnetExecutable = Path.Combine(_dotnetInstallDir, "dotnet");
```
dougbu: Why isn't this just `Path.GetTempFileName()`?

build/repo.targets
mikeharder: `build.cmd` will install the latest SDK and runtime, but it needs **some** target to build.  The default target would build the whole Benchmarks repo which is a waste of time.  The `NoOp` target does nothing so is as fast as possible.

mikeharder: Used here: https://github.com/aspnet/benchmarks/pull/235/files#diff-d3807a7e35147b0c86da6287c24065b7R232

davidfowl: Put that in the comment

src/BenchmarksServer/Startup.cs
```diff
@@ -214,13 +224,22 @@ private static string CloneRestoreAndBuild(string path, ServerJob job)
...
             AddSourceDependencies(path, benchmarksDir, dirs);
 
-            var benchmarksPath = Path.Combine(path, benchmarksDir, "src", "Benchmarks");
+            // Install latest SDK and runtime
+            // * Use custom install dir to avoid changing the default install,  which is impossible if other processes
+            //   are already using it.  Custom install is located under default install, so deleting the default
+            //   install will delete both.
+            var benchmarksRoot = Path.Combine(path, benchmarksDir);
+            ProcessUtil.Run("cmd", "/c build.cmd /t:noop", workingDirectory: benchmarksRoot,
+                environmentVariables: new Dictionary<string, string> { { "DOTNET_INSTALL_DIR", _dotnetInstallDir } });
```
dougbu: Need to handle non-Windows.

build/repo.targets
davidfowl: Ah!

src/BenchmarksServer/Startup.cs
```diff
@@ -214,13 +224,22 @@ private static string CloneRestoreAndBuild(string path, ServerJob job)
 
             AddSourceDependencies(path, benchmarksDir, dirs);
 
-            var benchmarksPath = Path.Combine(path, benchmarksDir, "src", "Benchmarks");
+            // Install latest SDK and runtime
+            // * Use custom install dir to avoid changing the default install,  which is impossible if other processes
+            //   are already using it.  Custom install is located under default install, so deleting the default
+            //   install will delete both.
```
dougbu: Why is it a goal to delete this copy of the `dotnet` folder when the real one is deleted? Suggest instead the usual `Path.GetTempFileName()` approach and clean the folder up when the `BenchmarksServer` process exits. People clear their temporary folder more often than they delete the `dotnet` folder.

mikeharder: The goal was to re-use the custom dotnet folder across runs to save time.  Using a temp dir will make the first run after starting the BenchmarksServer slow, but I agree this is a better solution since the BenchmarksServer is a long-running process.

src/BenchmarksServer/Startup.cs
```diff
@@ -214,13 +226,22 @@ private static string CloneRestoreAndBuild(string path, ServerJob job)
 
             AddSourceDependencies(path, benchmarksDir, dirs);
 
-            var benchmarksPath = Path.Combine(path, benchmarksDir, "src", "Benchmarks");
+            // Install latest SDK and runtime
+            // * Use custom install dir to avoid changing the default install,  which is impossible if other processes
+            //   are already using it.
+            var benchmarksRoot = Path.Combine(path, benchmarksDir);
+            ProcessUtil.Run("cmd", "/c build.cmd /t:noop", workingDirectory: benchmarksRoot,
+                environmentVariables: new Dictionary<string, string> { { "DOTNET_INSTALL_DIR", dotnetInstallDir } });
```
dougbu: Need to handle non-Windows. See uses of `_isWindows` in this file.

src/BenchmarksServer/Startup.cs
```diff
@@ -214,13 +226,22 @@ private static string CloneRestoreAndBuild(string path, ServerJob job)
...
+            var benchmarksApp = Path.Combine(benchmarksRoot, "src", "Benchmarks");
+            var dotnetExecutable = Path.Combine(dotnetInstallDir, "dotnet");
 
             // Project versions must be higher than package versions to resolve those dependencies to project ones as expected.
             // Passing VersionSuffix to restore will have it append that to the version of restored projects, making them
             // higher than packages references by the same name.
-            ProcessUtil.Run("dotnet", "restore /p:VersionSuffix=zzzzz-99999", workingDirectory: benchmarksPath);
-            ProcessUtil.Run("dotnet", $"build -c Release -f {GetTFM(job.Framework)}", workingDirectory: benchmarksPath);
+            ProcessUtil.Run(dotnetExecutable, "restore /p:VersionSuffix=zzzzz-99999", workingDirectory: benchmarksApp);
+            ProcessUtil.Run(dotnetExecutable, $"build -c Release -f {GetTFM(job.Framework)}", workingDirectory: benchmarksApp);
```
dougbu: Is the intention here to use the latest-and-greatest `dotnet` when running the benchmark? Or, should the solution configuration in the `source` commit be able to choose a specific version?

If the second, better to execute a single `build.cmd` / `build.sh` command which installs the necessary `dotnet`, restores one project, builds that project, then runs that project. Need a new target or three in KoreBuild for this.

mikeharder: The highest priority is preventing the benchmarks server from failing every time a newer runtime is targeted.  I agree using a single KoreBuild command would be simpler.  We could have a solution config which restores and builds only `src\Benchmarks\Benchmarks.csproj`.  We would also need to control the parameters to `restore` and `build` like we do currently, and I'm not sure if KoreBuild supports this today.

I'd like to move forward with the current design, and we can simplify to a single KoreBuild command later.

mikeharder:
@dougbu: Up to date and tested on Linux.

PR #249, opened 2017-04-09
Don't hard code the thread count when thread pool dispatching is off

davidfowl:
- Netty uses 2 * numbers of cores by default so that's what the benchmark does
when dispatching is off. On windows at least, picking 2 * cores is recommended for
IOCP usage. It's likely this is the same for linux (at least that's what i've seen in the wild)

### DIFF
--- a/src/Benchmarks/Program.cs
+++ b/src/Benchmarks/Program.cs
@@ -50,7 +50,7 @@ namespace Benchmarks
                     .AddSingleton<Scenarios>()
                 );
 
-            var threadPoolDispatching = true;
+            bool? threadPoolDispatching = null;
             if (String.Equals(Server, "Kestrel", StringComparison.OrdinalIgnoreCase))
             {
                 var kestrelThreadPoolDispatchingValue = config["KestrelThreadPoolDispatching"];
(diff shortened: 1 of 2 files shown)


src/Benchmarks/Program.cs
```diff
@@ -50,8 +50,15 @@ public static void Main(string[] args)
                     .AddSingleton<Scenarios>()
                 );
 
+            var threadPoolDispatching = true;
```
mikeharder: This should be a nullable bool to distinguish between `true`, `false`, and `Not Set`.  If the value is `Not Set`, then neither `options.ThreadCount` nor `internalOptions.ThreadPoolDispatching` should be modified.

This is a principle I've tried to maintain throughout Benchmarks.  If a Benchmarks config is not set, then framework defaults should not be changed.

davidfowl:
🆙 📅

src/Benchmarks/Program.cs
davidfowl: Made the change, am going to merge it.

tmds:
@davidfowl I've seen best result for such a benchmark on Linux when threadcount = processor cores (machine with no hyperthreading). So `ThreadCount = Environment.ProcessorCount` or `ThreadCount = Environment.ProcessorCount / 2` will probably do better than `ThreadCount = Environment.ProcessorCount * 2`.

PR #253, opened 2017-04-15
Fix bugs in KestrelTransport

mikeharder:
- Use default transport if KestrelTransport not specified
- Use Libuv if ThreadPoolDispatching is disabled


mikeharder:
CC @davidfowl

Issue #255, opened 2017-04-20
Benchmark Linux specific Transport


tmds:
@mikeharder I've been working on a Linux specific Transport. I'd like to include it here to get some benchmark results. I am doing some benchmarks of my own, but this is small compared to what you have set up. It would be great if I can reuse your efforts.

I closely follow the changes made in the upstream Transport abstraction so everything keeps working.

The Transport code is at: https://github.com/tmds/Tmds.Kestrel.Linux

If ok for you, I can send a PR to add this. I will follow the same patterns added recently to benchmark the Sockets transport.

cc @davidfowl @geoffkizer

DamianEdwards:
Relies on aspnet/Performance#226

tmds:
@davidfowl @DamianEdwards is the benchmark setup now capable of running the Linux specific Transport?

PR #256, opened 2017-04-20
Add Linux Transport benchmark

tmds:
Per https://github.com/aspnet/benchmarks/issues/255

@davidfowl

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -1,9 +1,11 @@
 <Project>
   <ItemGroup>
     <!-- Optimal ThreadCount for current hardware -->
-    <Scenarios Include="-n Plaintext --kestrelThreadCount 2" />
-    
+    <Scenarios Include="-n Plaintext --kestrelTransport Libuv --kestrelThreadCount 2" />
+    <Scenarios Include="-n Plaintext --kestrelTransport RHLinux --kestrelThreadCount 2" />
+
     <Scenarios Include="-n Plaintext --kestrelTransport Sockets" />
+    <Scenarios Include="-n Plaintext --kestrelTransport RHLinux" />
     <Scenarios Include="-n Plaintext -r Desktop" />
     <Scenarios Include="-n Plaintext --webHost HttpSys" />
     <Scenarios Include="-n Plaintext -f Benchmarks.PassthroughConnectionFilter" />
--- a/src/Benchmarks/Program.cs
+++ b/src/Benchmarks/Program.cs
@@ -111,7 +111,10 @@ namespace Benchmarks
                         {
                             options.ThreadCount = threadCount;
                         }
-                        options.SetThreadAffinity = true;
+                        else
+                        {
+                            options.SetThreadAffinity = true;
+                        }
                     });
                 }
                 else if (string.IsNullOrEmpty(kestrelTransport))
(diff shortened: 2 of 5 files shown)


build/repo.targets
```diff
@@ -1,9 +1,11 @@
 <Project>
   <ItemGroup>
     <!-- Optimal ThreadCount for current hardware -->
-    <Scenarios Include="-n Plaintext --kestrelThreadCount 2" />
-    
+    <Scenarios Include="-n Plaintext --kestrelTransport Libuv --kestrelThreadCount 2" />
```
tmds: Before setting kestrelThreadCount implied libuv. That is no longer the case. Made libuv explicit here.

build/repo.targets
```diff
@@ -1,9 +1,11 @@
...
+    <Scenarios Include="-n Plaintext --kestrelTransport RHLinux --kestrelThreadCount 2" />
```
tmds: I added one scenario with kestrelThreadCount and one without.
The one with is for the threadpool dispatch, the one without is for the inline dispatch.

src/Benchmarks/Program.cs
```diff
@@ -75,13 +77,13 @@ public static void Main(string[] args)
                         // Dispatching to the thread pool means we don't want to use the transport thread
                         options.UseTransportThread = !bool.Parse(kestrelThreadPoolDispatchingValue);
                     }
+                    threadPoolDispatching = !options.UseTransportThread;
```
tmds: threadPoolDispatching was not used. So I assigned it here.

src/Benchmarks/Program.cs
```diff
@@ -97,10 +99,28 @@ public static void Main(string[] args)
                         }
                     });
                 }
-                else if (string.Equals(kestrelTransport, "Sockets", StringComparison.OrdinalIgnoreCase))
+                else if (string.Equals(kestrelTransport, "Sockets", StringComparison.OrdinalIgnoreCase)
+                         && threadPoolDispatching != true)
```
tmds: This avoids benchmarking the double dispatch.

src/Benchmarks/Program.cs
```diff
@@ -97,10 +99,28 @@ public static void Main(string[] args)
...
                 {
                     webHostBuilder.UseSockets();
                 }
+                else if (string.Equals(kestrelTransport, "RHLinux", StringComparison.OrdinalIgnoreCase))
+                {
+                    if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                    {
+                        throw new PlatformNotSupportedException("RHLinux Transport requires Linux");
+                    }
+#if NETCOREAPP2_0
```
davidfowl: Why does this require netcoreapp2.0?

tmds: The app targets `<TargetFrameworks>netcoreapp2.0;net46</TargetFrameworks>`.
And the library targets `netstandard2.0`.
The netstandard2.0 library doesn't work on the net46 target (net461 needed for netstandard2.0).
I think I can make this work on netstandard1.3. Then this condition (+ matching one in csproj) can go.

tmds:
hi @davidfowl I still like to add this to the benchmark repo. I believe this wasn't merged due to a dependency on infrastructure changes (https://github.com/aspnet/Performance/issues/226). I see that issue still isn't resolved. Perhaps we are no longer depending on it? Let me know when I can bring this PR up to date so it can be merged.

sebastienros:
Done on TE

PR #261, opened 2017-05-05
Enable Copy scenario in automation

cesarbs:
- Add `ClientJob` parameter to specify client script
- Tweak `post.lua` script to read request data from a file with random data


src/Benchmarks/Benchmarks.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.0;net46</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
```
dougbu: This change should be in #262.

src/BenchmarksClient/Startup.cs
```diff
@@ -96,8 +96,13 @@ public static int Main(string[] args)
                 var job = allJobs.FirstOrDefault();
                 if (job != null)
                 {
-                    var jobLogText = $"[ID:{job.Id} Connections:{job.Connections} Threads:{job.Threads} " +
-                        $"Duration:{job.Duration} Pipeline:{job.PipelineDepth} Method:{job.Method}";
+                    var jobLogText =
+                        $"[ID:{job.Id} Connections:{job.Connections} Threads:{job.Threads} Duration:{job.Duration} Method:{job.Method}";
+
+                    if (!string.IsNullOrEmpty(job.ScriptName) && job.PipelineDepth > 0)
```
dougbu: Allow scripts when the pipeline depth is `0`. At most, `Debug.Assert(!string.IsNullOrEmpty(job.ScriptName)` when `PipelineDepth > 0`.

scripts/post.lua
```diff
@@ -1,17 +1,24 @@
...
+   f:close()
+   return data
+end
+
 function init(args)
    wrk.method = "POST"
-   wrk.body   = "foo=bar&baz=quux"
-   wrk.headers["Content-Type"] = "application/x-www-form-urlencoded"
+   wrk.body = readData("scripts/data.txt")
+   wrk.headers["Content-Type"] = "text/plain"
```
dougbu: Please confirm this file is presently unused. Might otherwise affect existing scenarios.

src/Benchmarks.ServerJob/Scenario.cs
```diff
@@ -13,6 +13,7 @@ public enum Scenario
     {
         Plaintext,
         Json,
+        Copy,
```
dougbu: Where did this name come from? Suggest `Post`.

src/BenchmarksClient/Startup.cs
```diff
@@ -155,9 +160,9 @@ private static Process StartProcess(ClientJob job)
 
             command += $" {job.ServerBenchmarkUri}";
 
-            if (job.PipelineDepth > 0)
+            if (!string.IsNullOrEmpty(job.ScriptName) && job.PipelineDepth > 0)
```
dougbu: Separate `ScriptName` and `PipelineDepth` handling.

scripts/post.lua
```diff
@@ -1,17 +1,24 @@
...
       pipelineDepth = args[1]
    end
 
    local r = {}
    for i = 1, pipelineDepth, 1 do
-      r[i] = wrk.format(nil)
+      r[i] = wrk.format()
    end
 
    print("Pipeline depth: " .. pipelineDepth)
```
dougbu: Do we need this? Because it's repeated for every client, always seems like noise.

src/Benchmarks/Benchmarks.csproj
cesarbs: Yes. I had this here so I could test it.

src/Benchmarks.ServerJob/Scenario.cs
cesarbs: It's testing the `CopyToAsync` method in the request stream.

dougbu: Then `CopyToAsync` would be a better name. At least make the names consistent; don't use `post` some of the time.

Also, update `Scenarios` to move the `Copy` property above the `Scenarios below this point are not currently automated.` comment.

cesarbs: I'll rename the scenario to `CopyToAsync`, but will keep the script name `post.lua` - all the script is going is posting data. From the script's perspective there's nothing about `CopyToAsync` going on.

scripts/post.lua
cesarbs: @mikeharder Thoughts?

We have to either remove it altogether or leave it there. There's no hook in the scripts that allows us to do something per wrk run - it's always per thread.

src/BenchmarksClient/Startup.cs
cesarbs: Why? Pipeline depth is a script argument, and if it's 0 that will generate 0 data.

I could make it so if pipeline depth <= 0 then default to 1.

dougbu: _Current_ scripts accept (but do not require) a pipeline depth argument. Even with that optionality, the restriction you're enforcing here doesn't make sense in general. What about future scripts with new arguments?

My proposed assertion would check that a non-default pipeline depth has a corresponding script to do something with it. Not a requirement from my perspective but somewhat in line with what you've written.

> I could make it so if pipeline depth <= 0 then default to 1.

Both existing scripts already do that.

src/Benchmarks.ServerJob/Scenario.cs
dougbu: A reasonable way forward.

src/BenchmarksClient/Startup.cs
cesarbs: > Both existing scripts already do that.

The scripts test that `args[1] ~= nil`. In Lua 0 != nil, so passing 0 would set the pipeline depth to 0.

scripts/post.lua
cesarbs: No scenarios are using that script. `pipeline.lua` is currently hardcoded for scenarios that use a script.

src/Benchmarks/Middleware/CopyToAsyncMiddleware.cs
```diff
@@ -20,17 +21,22 @@ public CopyToAsyncMiddleware(RequestDelegate next)
...
         {
             if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal))
             {
                 httpContext.Response.StatusCode = 200;
                 httpContext.Response.ContentType = "text/plain";
 
-                return httpContext.Request.Body.CopyToAsync(httpContext.Response.Body);
+                using (var ms = new MemoryStream())
+                {
+                    await httpContext.Request.Body.CopyToAsync(new MemoryStream());
```
cesarbs: Oops. Fixed.

cesarbs:
🆙 📅

scripts/post.lua
dougbu: Excellent

src/Benchmarks/Middleware/CopyToAsyncMiddleware.cs
```diff
@@ -20,17 +21,25 @@ public CopyToAsyncMiddleware(RequestDelegate next)
...
+                    // the time to copy the body and not copying plus writing
+                    // the response.
+                    await httpContext.Request.Body.CopyToAsync(ms);
+                }
+
+                await httpContext.Response.WriteAsync("Hello World!");
             }
 
-            return _next(httpContext);
+            await _next(httpContext);
```
dougbu: Very weird to call `_next` after writing to the `Response`. This doesn't match what the test did before.

cesarbs: Just fixed it.

PR #262, opened 2017-05-05
netcoreapp2.0

cesarbs:
- Target netcoreapp2.0 only.
- Remove long abandoned stuff under 'experimental'
- Remove framework switch from automation


build/dependencies.props
```diff
@@ -1,16 +1,25 @@
 <Project>
   <PropertyGroup>
-    <NetStandardImplicitPackageVersion>1.6.1</NetStandardImplicitPackageVersion>
+    <AspNetCoreVersion>2.0.0-*</AspNetCoreVersion>
+    <CoreFxVersion>4.3.0</CoreFxVersion>
+    <DapperVersion>1.*</DapperVersion>
+    <InternalAspNetCoreSdkVersion>2.0.0-*</InternalAspNetCoreSdkVersion>
+    <JsonNetVersion>9.0.1</JsonNetVersion>
+    <NETStandardImplicitPackageVersion>$(BundledNETStandardPackageVersion)</NETStandardImplicitPackageVersion>
     <RuntimeFrameworkVersion>2.0.0-*</RuntimeFrameworkVersion>
```
cesarbs: @natemcmaster Is this used by KoreBuild anywhere? I don't see it in other repos. Wondering if I can remove it.

src/BenchmarksDriver/Program.cs
```diff
@@ -592,7 +579,7 @@ public static int Main(string[] args)
                     var p = command.Parameters;
                     p.AddWithValue("@DateTime", DateTimeOffset.UtcNow);
                     p.AddWithValue("@Scenario", scenario.ToString());
-                    p.AddWithValue("@Framework", framework.ToString());
+                    p.AddWithValue("@Framework", "Core");
```
cesarbs: @mikeharder Do we want to keep this around at all?

mikeharder: We should preserve the historical data, which means we cannot delete the column from SQL.  Your change is good for now.  After your change is merged, I will update the column in the live database to include a default value, and remove the column from the insert statement.

src/Repository/Repository.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <Description>Repository to hold objects which can be identified by an integer.</Description>
-    <TargetFramework>netstandard1.0</TargetFramework>
```
mikeharder: I think `TargetFramework` should remain `netstandard1.0` since no dependencies require higher than this.

src/Benchmarks/Benchmarks.csproj
```diff
@@ -3,42 +3,39 @@
...
+    <PackageReference Include="Microsoft.AspNetCore.StaticFiles" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.CommandLine" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
```
davidfowl: You should be able to kill this dependency

src/Repository/Repository.csproj
davidfowl: Nah, lets up bump it 😄

benchmarks.sln
```diff
@@ -12,14 +12,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "experimental", "experimental", "{C2EB60CC-CA5A-4E49-A3F8-4432FB117869}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BasicAsyncHttpServer", "experimental\BasicAsyncHttpServer\BasicAsyncHttpServer.csproj", "{07773C38-B3F3-4D6C-B318-29C88F016AA9}"
```
dougbu: +1 and what's special about experimental\HttpBenchmark\HttpBenchmark.csproj that let's it live?

build/dependencies.props
```diff
@@ -1,16 +1,25 @@
 <Project>
   <PropertyGroup>
-    <NetStandardImplicitPackageVersion>1.6.1</NetStandardImplicitPackageVersion>
+    <AspNetCoreVersion>2.0.0-*</AspNetCoreVersion>
+    <CoreFxVersion>4.3.0</CoreFxVersion>
+    <DapperVersion>1.*</DapperVersion>
+    <InternalAspNetCoreSdkVersion>2.0.0-*</InternalAspNetCoreSdkVersion>
+    <JsonNetVersion>9.0.1</JsonNetVersion>
```
dougbu: Why are we using a lower Json.NET version here than in any other repo? Should be `10.0.1`.

src/Benchmarks/Benchmarks.csproj
natemcmaster: Tangent: we looked into removing from runtime projects too, but TaskCache.CompletedCache contains this comment:

>  We're caching this in a static readonly field to make it more inlinable and avoid the volatile lookup done by Task.CompletedTask.

cc @pakrym Any idea if this still applies?

build/dependencies.props
natemcmaster: Can be removed. Our other repos have removed this. The SDK now appropriate sets this value to BundledNETCoreAppPackageVersion so that the package version and runtime always line up.

benchmarks.sln
cesarbs: Checked with @DamianEdwards a while back and he said it was fine. I've kept HttpBenchmark per @mikeharder's request at the same time.

src/Benchmarks.ClientJob/Benchmarks.ClientJob.csproj
```diff
@@ -4,12 +4,12 @@
 
   <PropertyGroup>
     <Description>Class to transfer data to/from the benchmark client.</Description>
-    <TargetFramework>netstandard1.0</TargetFramework>
```
cesarbs: TBH I'm confused about this because I'm seeing everything being lifted to `netcoreapp2.0` in other repos, even things that were `netstandard1.*`. @Eilon what is the guidance here?

build/dependencies.props
cesarbs: I just kept what was in the csproj.

dougbu: Don't preserve that mistake.

src/Benchmarks.ClientJob/Benchmarks.ClientJob.csproj
Eilon: I think it depends on where this is used. But, is there any reason *not* to use NCA2.0 here?

mikeharder: This is an internal DLL shared by different components of BenchmarkAutomation.  So it doesn't really matter what we choose here.

However, following the spirit of the changes, I do think this should stay `netstandard1.0` since it doesn't depend on anything higher.  But, nothing should break if this is moved to `netcoreapp2.0` either.

Eilon: Flip a coin.

src/Benchmarks/Benchmarks.csproj
pakrym: Not sure

cesarbs: Just checked and there's a single place where this is used (`JsonMiddleware`). I don't think we need this dependency just for that, and we can use `Task.CompletedTask` there.

Any objections?

cesarbs:
📢

src/Benchmarks/Benchmarks.csproj
dougbu: Unfortunately the 🏏s emoji is about the game 😸

Issue #263, opened 2017-05-05
Upgrade Npgsql from "3.2.0-*" to "3.*"


mikeharder:
Now that `Npgsql 3.2.0` has shipped, the dependency version should be changed from `3.2.0-*` to `3.*`.  We only needed `3.2.0-*` to pickup `3.2.0-beta`, since `3.*` doesn't match pre-release versions.

dougbu:
Taking this as I'm doing related cleanup / alignment w/ TechEmpower repo.

dougbu:
8fa6759

Issue #251, opened 2017-04-14
Update benchmarks to use versions in dependencies.props


davidfowl:
Today the projects in benchmarks have a dependencies.props but we still hard code versions in the project files themselves.

https://github.com/aspnet/benchmarks/blob/ae6af537186f8493d34f8ac97f77d6b878aa573b/src/Benchmarks/Benchmarks.csproj#L11-L32

We should replace these with properties specified in `dependencies.props`

dougbu:
Taking this as I'm doing related cleanup / alignment w/ TechEmpower repo.

davidfowl:
👍

dougbu:
8fa6759

PR #269, opened 2017-06-07
Bring latest Benchmarks code over from TechEmpower repo

dougbu:
- see individual commits for details
- mostly alignment with TechEmpower code i.e. recent fixes there

addresses:
- #251, Update benchmarks to use versions in dependencies.prop
- #263, Upgrade Npgsql from "3.2.0-*" to "3.*"


cesarbs:
Can you change the title of the PR to reflect what the changes are about?

dougbu:
@cesarbs yes, I can 😈

dougbu:
:up::date: to match final TechEmpower code

build/dependencies.props
```diff
@@ -14,6 +14,8 @@
     <NpgsqlVersion>3.2.0-*</NpgsqlVersion>
     <NpgsqlEntityFrameworkCorePostgreSQLVersion>1.*</NpgsqlEntityFrameworkCorePostgreSQLVersion>
 
+    <RuntimeFrameworkVersion Condition="'$(TargetFramework)'=='netcoreapp2.0'">2.0.0-*</RuntimeFrameworkVersion>
```
cesarbs: Why do we need this?

src/Benchmarks/Benchmarks.csproj
```diff
@@ -1,34 +1,33 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Web">
-
-  <Import Project="..\..\build\common.props" />
-
+<Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
     <TargetFramework>netcoreapp2.0</TargetFramework>
     <OutputType>Exe</OutputType>
+    <MvcRazorCompileOnPublish>true</MvcRazorCompileOnPublish>
+    <RuntimeFrameworkVersion>2.0.0-*</RuntimeFrameworkVersion>
```
cesarbs: Don't we get this from `dependencies.props`?

src/Benchmarks/Benchmarks.csproj
```diff
@@ -1,34 +1,33 @@
...
-    -->
-    <PackageReference Include="Npgsql" Version="$(NpgsqlVersion)" />
-    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="$(NpgsqlEntityFrameworkCorePostgreSQLVersion)" />
-    <PackageReference Include="System.Runtime.Serialization.Primitives" Version="$(CoreFxVersion)" />
+    <None Update="wwwroot/**" CopyToOutputDirectory="PreserveNewest" />
+    <None Include="appsettings.json" CopyToOutputDirectory="PreserveNewest" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Dapper" Version="1.*" />
```
cesarbs: Why are versions hardcoded again?

davidfowl:
TechEmpower doesn't use floating versions does it?

build/dependencies.props
dougbu: To avoid relying on dependencies.props. See #251

src/Benchmarks/Benchmarks.csproj
dougbu: We don't import dependencies.props. See #251

src/Benchmarks/Benchmarks.csproj
dougbu: To avoid relying on dependencies.props. See #251

dougbu:
> TechEmpower doesn't use floating versions does it?

No, it doesn't. Floating versions here and fewer supported scenarios there are the main differences between here and the TechEmpower repo.

build/dependencies.props
dougbu: Oops, wrong answer. This is here to match what we do in other repos. I bet it's there to avoid version mismatches and associated warnings.

src/Benchmarks/Benchmarks.csproj
```diff
@@ -1,34 +1,33 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Web">
-
-  <Import Project="..\..\build\common.props" />
-
+<Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
     <TargetFramework>netcoreapp2.0</TargetFramework>
     <OutputType>Exe</OutputType>
+    <MvcRazorCompileOnPublish>true</MvcRazorCompileOnPublish>
```
davidfowl: Why is this here

build/dependencies.props
cesarbs: Didn't realize it at first. Can you group it together with the properties at the top of the file? That's where it is in other `dependencies.props` files.

src/Benchmarks/Benchmarks.csproj
cesarbs: Why don't we import it?

src/Benchmarks/Benchmarks.csproj
cesarbs: Why?

src/Benchmarks/Benchmarks.csproj
dougbu: @davidfowl prefers that we don't

src/Benchmarks/Benchmarks.csproj
dougbu: To enable pre-compilation and to be explicit about it. Means we need fewer changes when using 1.1.x because pre-compilation wasn't the default with those versions.

build/dependencies.props
dougbu: Huh? The property settings in this file are in alphabetic order.

dougbu: But, just realized we don't need to set `$(DapperVersion)`, `$(NpgsqlVersion)`, or `$(NpgsqlEntityFrameworkCorePostgreSQLVersion)` at all. Removing the last two will bring everything together.

src/Benchmarks/Benchmarks.csproj
davidfowl: I do?

src/Benchmarks/Benchmarks.csproj
davidfowl: What are we precompiling in benchmarks?

dougbu: TechEmpower scenarios include a number with MVC views.

src/Benchmarks/Benchmarks.csproj
dougbu: This file previously used only properties defined in dependencies.props. So, if what you wanted was that (and I misread the issue completely), #251 was fixed before you filed it.

cesarbs: I think at the time @davidfowl filed the issue, the versions were hardcoded in the csproj. But now they are using the versions defined in `dependencies.props`, which is what he wanted. #251 had already been fixed (by me actually, but I wasn't aware there was an open issue for it).

build/dependencies.props
cesarbs: 👍

src/Benchmarks/Benchmarks.csproj
davidfowl: Yep

dougbu:
:up::date:

@davidfowl Benchmarks.csproj now uses separate properties from dependencies.props -- as we discussed.

dougbu:
@cesarbs @davidfowl any additional comments?

src/Benchmarks/Middleware/DebugInfoPageMiddleware.cs
```diff
@@ -42,6 +42,7 @@ public DebugInfoPageMiddleware(RequestDelegate next, IServerAddressesFeature ser
         public async Task Invoke(HttpContext httpContext)
         {
             httpContext.Response.ContentType = "text/html";
+            httpContext.Response.StatusCode = StatusCodes.Status404NotFound;
```
davidfowl: Why is the status 404?

src/Benchmarks/Program.cs
```diff
@@ -34,6 +33,7 @@ public static void Main(string[] args)
             Console.WriteLine("-----------------------");
 
             Console.WriteLine($"Current directory: {Directory.GetCurrentDirectory()}");
+            Console.WriteLine($"WebHostBuilder loading from: {typeof(WebHostBuilder).GetTypeInfo().Assembly.Location}");
```
davidfowl: Why is this relevant? To show if it's using the package store or not?

src/Benchmarks/Program.cs
```diff
@@ -59,7 +59,9 @@ public static void Main(string[] args)
                     .AddSingleton(new ConsoleArgs(args))
                     .AddSingleton<IScenariosConfiguration, ConsoleHostScenariosConfiguration>()
                     .AddSingleton<Scenarios>()
-                );
+                )
+                .UseDefaultServiceProvider(
```
davidfowl: Why did we add this to the benchmarks?

src/Benchmarks/Startup.cs
```diff
@@ -138,8 +123,6 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
             {
                 services.AddResponseCaching();
             }
-
-            return services.BuildServiceProvider(validateScopes: true);
```
davidfowl: I see why

src/Benchmarks/Program.cs
davidfowl: Never mind.

build/dependencies.props
```diff
@@ -2,23 +2,29 @@
...
 
     <!--
       * Use 4.4.0-* instead of $(CoreFxVersion) to address "SqlClient fails with netcoreapp2.0 on Win7/Server2008"
         * https://github.com/dotnet/corefx/issues/18406
       * Revert if and when $(CoreFxVersion) is upgraded to 4.4.0-*
     -->
     <SqlClientVersion>4.4.0-*</SqlClientVersion>
+
+    <!--
+      Following properties are specific to src/Benchmarks/Benchmarks.csproj.
```
davidfowl: Why do we need these?

src/Benchmarks/Middleware/DebugInfoPageMiddleware.cs
dougbu: This middleware is only used when none of the others match i.e. nothing was found.

Besides that, I mistakenly did a manual `wrk` run with an incorrect URL. Took much longer than it should have to realize why I was seeing higher-than-MVC numbers. This prevents future mistakes like that and browsers still show all the info.

src/Benchmarks/Program.cs
dougbu: Yup

build/dependencies.props
dougbu: So that we can test previous releases without having to undo tracking changes in all of the web projects. This repo is far easier to automate than our fork of the TechEmpower repo.

I plan to port my branches for the existing releases from that fork here.

src/Benchmarks/Middleware/DebugInfoPageMiddleware.cs
davidfowl: Not that it actually mattes but it seems wrong, this is an actual valid URL.

dougbu: I'd agree if this middleware only responded to specific URLs.

dougbu:
8fa6759

PR #270, opened 2017-06-15
Always reference latest NETStandard.Library

dougbu:
- aspnet/Coherence-Signed#565
- continues work @kichalla did recently


dougbu:
fd3b08e

PR #273, opened 2017-07-26
Update bootstrappers to install the compiled version of KoreBuild

natemcmaster:
Part of https://github.com/aspnet/KoreBuild/issues/258


.gitignore
```diff
@@ -280,3 +280,4 @@ __pycache__/
 # Cake - Uncomment if you are using it
 # tools/**
 # !tools/packages.config
+korebuild-lock.txt
```
mikeharder: Would a better name be something like `.korebuildlock`?

version.xml
```diff
@@ -0,0 +1,8 @@
+<!-- This file may be overwritten by automation. -->
```
mikeharder: Why rename from `.props` to `.xml`?  I don't have a preference but I'm curious why you did it.

natemcmaster: The file may be interpreted as xml. e.g. build.sh/ps1 read the file to get values

.gitignore
natemcmaster: I think the file name is a matter of preference. I prefer to have the file extension .txt. In the event we decide to add more to the lockfile or change its format, it could become korebuild-lock.{json,xml,csv,yml}.

PR #278, opened 2017-08-08
Add `systemd` configuration file for BenchmarksServer

dougbu:
- configuration used on asp-perf-linux

### DIFF
--- a/scripts/aspnet.benchmarks.service
+++ b/scripts/aspnet.benchmarks.service
@@ -1,0 +1,20 @@
+[Unit]
+Description=ASP.NET Benchmarks Server
+
+# Start after network has at least a bound IP address; stop before network is stopped
+After=network.target
+After=network-online.target
+
+[Service]
+Environment=NUGET_PACKAGES=/home/aspnet/.nuget/packages
+Environment=NUGET_HTTP_CACHE_PATH=/home/aspnet/.local/share/NuGet/v3-cache
+ExecStart=/home/aspnet/.dotnet/dotnet bin/Debug/netcoreapp2.0/BenchmarksServer.dll --hostname 10.0.0.102
+
+Restart=on-failure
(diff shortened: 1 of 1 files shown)


scripts/aspnet.benchmarks.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=ASP.NET Benchmarks Server
+
+# Start after network has at least a bound IP address; stop before network is stopped
+After=network.target
+After=network-online.target
```
mikeharder: I think we should remove `network-online.target`.  From https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/:

> It is strongly recommended not to pull in this target too liberally: for example network server software should generally not pull this in (since server software generally is happy to accept local connections even before any routable network interface is up), it's primary purpose is network client software that cannot operate without network.

dougbu: This line does not "pull in" the `network-online.target`. It controls only ordering; `Requires` and `Wants` are completely orthogonal.

mikeharder: I think the point is, why does a server need to wait for `network-online.target`?  What's the harm in starting a server before the network is up?

mikeharder: As an example, nginx only depends on `network.target`:

https://www.nginx.com/resources/wiki/start/topics/examples/systemd/

dougbu: Discussed offline and I'll try a few reboots without `After=network-online.target` to confirm we don't need it. (If there's something ASP.NET Core-specific here, it's not currently documented.)

dougbu:
Files updated to address comments

dougbu:
That was FYI. f7915f5 now in `dev`.

PR #280, opened 2017-08-24
Benchmarks and PostgreSQL in Docker

mikeharder:


docker/postgres-techempower-fortunesbench/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM postgres
```
anurse: Do we want to float on `latest` or should we pin a version?

docker/postgres-techempower-fortunesbench/run.sh
```diff
@@ -0,0 +1,23 @@
...
+
+host=$1
+threads=$(nproc --all)
+connections=$(($threads * 8))
+
+#echo on
+set -x
+
+docker run \
+    -it \
```
anurse: The others all daemonize, is there a reason this one doesn't or was it just missed?

docker/benchmarks/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+        git \
+        procps \
+    && rm -rf /var/lib/apt/lists/*
+
+# Runtimes prior to 2.0.0 are not supported on debian:stretch (or later), but they should
+# not be required for Benchmarks.
+ENV KOREBUILD_SKIP_OLD_RUNTIME_INSTALL 1
+
+WORKDIR /benchmarks
+COPY . .
```
anurse: A slightly more efficient layering (though it's not a big thing here) would be to do a single `COPY` that pulls in the `sln` and `csproj`, then `RUN dotnet restore` to restore NuGet packages into the container, then do `COPY   . .` to copy the rest of the files. Then, the nuget packages came be cached and that layer is only rebuilt when the dependencies of one of the projects changes.

docker/README.md
```diff
@@ -0,0 +1,33 @@
+# Benchmarks in Docker
```
natemcmaster: Doesn't need to be in this PR, but have you considered adding a docker-compose.yml file to make it easier to start combinations of containers?

docker/benchmarks/run-client.sh
```diff
@@ -0,0 +1,15 @@
...
+docker run \
+    -d \
+    --log-opt max-size=10m \
+    --log-opt max-file=3 \
+    --name benchmarks-client \
+    --network host \
+    --restart always \
+    benchmarks \
+    /root/.dotnet/dotnet \
+    /benchmarks/src/BenchmarksClient/bin/Debug/netcoreapp2.0/BenchmarksClient.dll
```
natemcmaster: Should this be configured to Release instead of Debug?

docker/postgres-techempower-fortunesbench/Dockerfile
natemcmaster: Pin, definitely pin.

docker/postgres-techempower/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM postgres
+
+ADD max_connections.sh /docker-entrypoint-initdb.d/
+
+ADD https://raw.githubusercontent.com/TechEmpower/FrameworkBenchmarks/master/toolset/setup/linux/databases/postgresql/create-postgres.sql /docker-entrypoint-initdb.d/
```
natemcmaster: Would be good to use a permalink, or make a local copy of this file.

docker/README.md
mikeharder: For perf testing, each container needs to be run on a separate VM.  We talked about using Docker Swarm to start the whole cluster, but it didn't seem necessary yet.  It's easy enough to build and run the images on each host VM as needed.

docker/benchmarks/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+        libicu57 \
+        liblttng-ust0 \
+        libssl1.0.2 \
+        libstdc++6 \
+        libunwind8 \
+        libuuid1 \
+        zlib1g \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN apt-get update \
```
mikeharder: The idea was the .NET Core dependencies are temporary, so I just copied the whole `apt-get` statement from the .NET Core Dockerfile.  I'll move these as a section in the other `apt-get` statement, it will be just as easy to remove later.

docker/benchmarks/Dockerfile
mikeharder: Is there a wildcard to copy all csproj, or would each need to be listed explicitly (and updated when they change)?

docker/benchmarks/run-client.sh
mikeharder: The Client and Server are fine to run in Debug, they are just web APIs which manage the client and server jobs.  The app we use to measure RPS is always run in Release.

docker/postgres-techempower-fortunesbench/run.sh
mikeharder: The `postgres-techempower-fortunesbench` is a client app which runs a raw SQL benchmark against postges, simulating the queries run by the `fortunes` web benchmark.  It's designed to run in a container, print the benchmark results, and exit.

docker/benchmarks/Dockerfile
anurse: I believe `src\*\*.csproj` may work (there's no recursive wildcard, it uses Go's [filepath.Match rules](https://golang.org/pkg/path/filepath/#Match)). Also, if the `sln` is present and the `csproj` is added to the `sln`, restore would fail if you added a `csproj` to the `sln` but not to the Dockerfile. So it would be a clearer failure.

Anyway, it's just optimizing build times, so it's not essential.

docker/postgres-techempower-fortunesbench/run.sh
anurse: Perhaps a comment to that effect would help.

anurse: Though the people looking at this most of the time will probably know it already ;)

anurse:
Also `sysctl` and `ulimit` calls are tricky in Docker since they control kernel options. Many will need to be run on the host rather than the container (and some will also need to be propagated as `--ulimit` switches in `docker run` because docker configures special resource limits in the kernel, because that's how the whole container thing works in the first place :)).

docker/benchmarks/Dockerfile
mikeharder: `COPY src/*/*.csproj ./` doesn't really work, since it doesn't preserve the directory structure.  It copies all the `.csproj` files to the root of the destination.  A single copy is good enough for now.

PR #282, opened 2017-08-24
Record Hardware and OperatingSystem

mikeharder:


build/repo.targets
```diff
@@ -1,50 +1,57 @@
...
     <BenchmarksDependsOn Condition="'$(NoBuild)' != 'true'">Compile</BenchmarksDependsOn>
   </PropertyGroup>
 
   <Target Name="TestBenchmarks" DependsOnTargets="$(BenchmarksDependsOn)">
-    <Error Text="BENCHMARK_SERVER must be set" Condition="'$(BENCHMARK_SERVER)' == ''" />
-    <Error Text="BENCHMARK_CLIENT must be set" Condition="'$(BENCHMARK_CLIENT)' == ''" />
+    <Error Text="BENCHMARK_SERVER_PHYSICAL must be set" Condition="'$(BENCHMARK_SERVER_PHYSICAL)' == ''" />
+    <Error Text="BENCHMARK_CLIENT_PHYSICAL must be set" Condition="'$(BENCHMARK_CLIENT_PHYSICAL)' == ''" />
+    <Error Text="BENCHMARK_SERVER_CLOUD must be set" Condition="'$(BENCHMARK_SERVER_CLOUD)' == ''" />
+    <Error Text="BENCHMARK_CLIENT_CLOUD must be set" Condition="'$(BENCHMARK_CLIENT_CLOUD)' == ''" />
```
dougbu: Should we make one or both run types conditional? E.g. fail if none of the cloud or physical variables are set or if (say) the cloud server variable is set but not the cloud client variable. Then, the `<MsBuild>` tasks below would be conditional on the necessary variables.
The alternative is to assume anyone subsetting the tests would edit this file temporarily (including removing `<Error>` tasks as needed.

docker/benchmarks/run-server.sh
```diff
@@ -5,6 +5,13 @@ set -x
 
 server_ip=$(ip route get 1 | awk '{print $NF;exit}')
 
+if [ -e /var/log/waagent.log ]
```
dougbu: Why not look for this file in the `BenchmarkServer.Startup` constructor? Seems inconsistent and extra complicated to require the value to be passed in.

src/BenchmarksServer/Startup.cs
```diff
@@ -30,7 +32,25 @@ public class Startup
         private static readonly string _defaultHostname = Environment.MachineName.ToLowerInvariant();
 
         private static readonly IRepository<ServerJob> _jobs = new InMemoryRepository<ServerJob>();
-        private static readonly bool _isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+
+        public static OperatingSystem OperatingSystem { get; private set; }
```
dougbu: Don't need the `private set` because only the constructor sets it.

build/repo.targets
mikeharder: This target is mainly used by the CI where both will be set.  I don't want to put more effort into refactoring this file unless people actually use it for ad-hoc testing.   @natemcmaster said he does occasionally, but I think most people run `BenchmarksDriver` manually for ad-hoc testing.

docker/benchmarks/run-server.sh
mikeharder: `BenchmarkServer` is running inside a container, so it can't detect whether the host is `Cloud` or `Physical`.  I couldn't find a better way than requiring a command-line parameter.

dougbu: Ick. Guess the option to conditionally mount this directory in the `docker run` command is just as ugly -- more complex here balanced against only slight `BenchmarkServer.Startup` cleanup.

build/repo.targets
mikeharder: @dougbu: Based on my own ad-hoc usage, I've changed my mind and updated the PR.  I think it's sufficient to make the tasks conditional on the env vars without any error messages.

dougbu: 👍 Looks good and matches my ad-hoc usage as well.

PR #283, opened 2017-08-25
Refactor ClientJobs to reduce duplication

mikeharder:


src/BenchmarksDriver/Program.cs
```diff
@@ -25,60 +25,52 @@ public class Program
...
+                { Scenario.MvcJson, new ClientJob(_jsonJob) },
+                { Scenario.MemoryCachePlaintext, new ClientJob(_plaintextJob) },
+                { Scenario.MemoryCachePlaintextSetRemove, new ClientJob(_plaintextJob) },
+                { Scenario.ResponseCachingPlaintextCached, new ClientJob(_plaintextJob) },
+                { Scenario.ResponseCachingPlaintextResponseNoCache, new ClientJob(_plaintextJob) },
+                { Scenario.ResponseCachingPlaintextRequestNoCache, new ClientJob(_plaintextJob) {
+                    Headers = _plaintextJob.Headers.Append("Cache-Control: no-cache")
                 } },
+                { Scenario.ResponseCachingPlaintextVaryByCached, new ClientJob(_plaintextJob) },
+                { Scenario.StaticFiles, new ClientJob(_plaintextJob) },
```
dougbu: For the many entries where the `CleintJob` is exactly `_plaintextJob` or `_jsonJob`, why make copies? Changes made later are either done after creating a copy or applied to all entries in this dictionary. This is minor from a performance perspective but makes the code harder to read.

mikeharder: Good suggestion.  Fixed.

PR #284, opened 2017-08-28
Serialize a JObject

benaadams:
Seems to go a bit better

/cc @JamesNK

### DIFF
--- a/src/Benchmarks/Middleware/JsonMiddleware.cs
+++ b/src/Benchmarks/Middleware/JsonMiddleware.cs
@@ -9,6 +9,7 @@ using Benchmarks.Configuration;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace Benchmarks.Middleware
 {
--- a/src/Benchmarks/Middleware/JsonMiddleware.cs
+++ b/src/Benchmarks/Middleware/JsonMiddleware.cs
@@ -15,7 +15,6 @@ namespace Benchmarks.Middleware
     public class JsonMiddleware
     {
         private static readonly PathString _path = new PathString(Scenarios.GetPath(s => s.Json));
-        private static readonly JsonSerializer _json = new JsonSerializer();
         private static readonly UTF8Encoding _encoding = new UTF8Encoding(false);
         private const int _bufferSize = 27;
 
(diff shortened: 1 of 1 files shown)


mikeharder:
@benaadams: Do you have any numbers showing the perf impact of this change?

JamesNK:
I believe the rules of the test are that the type is serialized using a generic mechanism rather than is hard coded. Do any of the JSON tests from others do something like this?

benaadams:
> 1. For each request, an object mapping the key message to Hello, World! must be instantiated.
> 2. The recommended URI is /json.
> 3. A JSON serializer must be used to convert the object to JSON.
> 4. The response text must be {"message":"Hello, World!"}, but white-space variations are acceptable.
> 5. The response content length should be approximately 28 bytes.
> ...

Is still doing 1 & 3 ?

mikeharder:
@benaadams: This doesn't compile for me.  I think you're missing a using directive.

```
Middleware\JsonMiddleware.cs(39,35): error CS0246: The type or namespace name 'JObject' could not be found (are you missing a using directive or an assembly reference?)
```

mikeharder:
Assuming `JObject.WriteTo()` is faster than `JsonSerializer.Serialize()`, my next question would be why can't the implementation of `JsonSerializer.Serialize()` use the same code as this PR?  I don't know much about Json.NET so this may be a bad question.

benaadams:
Added missing using

mikeharder:
I only collected 3 samples, but I don't see a measurable perf benefit on our physical hardware:

Branch | RPS | Average Deviation
------------ | ------------- | -------------
dev | 282,448 | 2,096
benaadams:jobject | 282,278 | 1756

benaadams:
@JamesNK isn't giving me easy wins... 😺

benaadams:
Though `StreamWriter` why you allocate so much! O_o but that's corefx...

benaadams:
Main allocations in Json test is from `StreamWriter` so I've added a PR for that https://github.com/dotnet/corefx/pull/23727

davidfowl:
We have an HttpResponseStreamWriter built in that we could use. It does the pooling using the array pool.

jkotalik:
See https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.WebUtilities/HttpResponseStreamWriter.cs

benaadams:
@davidfowl sounds like a target to beat 😉

PR #285, opened 2017-08-28
Add Fortunes to Benchmarks automation

mikeharder:


src/BenchmarksServer/Startup.cs
```diff
@@ -96,13 +100,29 @@ public static int Main(string[] args)
...
+
+
+                return Run(url, hostname, database, sqlConnectionStringOption.Value()).Result;
             });
 
             return app.Execute(args);
         }
 
-        private static async Task<int> Run(string url, string hostname)
+        private static async Task<int> Run(string url, string hostname, Database? database, string sqlConnectionString)
```
dougbu: Should this method complain / throw / return non-`0` if `database` is `null` but `sqlConnectionString` is not (or visa versa)?

mikeharder: I considered that as well.  However, the Benchmarks app itself doesn't perform this check, so I decided to make BenchmarksServer match for now.  So everything allowed by Benchmarks is allowed by BenchmarksServer, even if it doesn't necessarily make sense.

https://github.com/aspnet/benchmarks/blob/dev/src/Benchmarks/Startup.cs#L56-L71

Issue #288, opened 2017-08-30
Can run benchmarks with a repo that uses a custom version of roslyn


natemcmaster:
i.e. KestrelHttpServer uses an experimental version of rosyn, which means we can't compile Benchmarks.csproj with a ProjectReference to it.

https://github.com/aspnet/KestrelHttpServer/blob/284367db9ffca9ecc3eb734a068e52f682f569fa/Directory.Build.props#L17-L25

The error:

> 02:26:26.966] [/tmp/tmpkrngEo.tmp/benchmarks/src/Benchmarks] [/tmp/tmp56HflU.tmp/dotnet build -c Release] /tmp/tmp56HflU.tmp/sdk/2.1.0-preview1-007042/Roslyn/Microsoft.CSharp.Core.targets(84,5): error MSB4062: The "Microsoft.CodeAnalysis.BuildTasks.Csc" task could not be loaded from the assembly /tmp/tmp56HflU.tmp/sdk/2.1.0-preview1-007042/Roslyn/Microsoft.Build.Tasks.CodeAnalysis.dll. Assembly with same name is already loaded Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask. [/tmp/tmpkrngEo.tmp/benchmarks/src/Benchmarks/Benchmarks.csproj]

cc @pakrym @mikeharder

PR #289, opened 2017-09-05
Reference custom compiler package from benchmarks

pakrym:
Fixes https://github.com/aspnet/benchmarks/issues/288


src/Benchmarks/Benchmarks.csproj
```diff
@@ -1,5 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
-  <Import Project="..\..\build\dependencies.props" />
```
pakrym: Was referencing `dependencies.props` instead of `common.prop` intentional?

PR #292, opened 2017-09-05
Install runtime on /t:Noop and pass DOTNET_HOME in PATH

pakrym:
Checking that docker image builds right now


mikeharder:
@pakrym: I'm going to move my `Dockerfile` changes to a separate PR, primarily so it's easier to revert the 1.1 runtime changes in isolation.

PR #293, opened 2017-09-06
Change base image to 2.1.0-preview1-runtime-deps-jessie

mikeharder:
Required to support the customer compiler, which requires the 1.1 runtime, which is only supported on Debian Jessie.


dougbu:
> the 1.1 runtime

Do you mean the 2.1 runtime? And what is the "customer compiler"?

pakrym:
We use custom branch of compiler to build kestrel and need to also use it in benchmarks when referencing a branch. Compiler requires 1.1 runtime to run.

PR #301, opened 2017-09-08
Add remaining DB scenarios

mikeharder:


src/BenchmarksClient/Startup.cs
```diff
@@ -163,7 +163,7 @@ private static Process StartProcess(ClientJob job)
                 }
             }
 
-            command += $" --latency -d {job.Duration} -c {job.Connections} --timeout 8 -t {job.Threads}  {job.ServerBenchmarkUri}";
+            command += $" --latency -d {job.Duration} -c {job.Connections} --timeout 8 -t {job.Threads}  {job.ServerBenchmarkUri}{job.Query}";
```
dougbu: Is this missing a space before `{job.Query}`?

dougbu: Is this command missing a space before `{job.Query}`?

mikeharder: No.  `{job.Query}` is a query string, like `?queries=20`, so it needs to be appended to the URL without a space.

PR #304, opened 2017-09-18
Update compiler and remove 1.1 runtime installation support

pakrym:
https://github.com/aspnet/KestrelHttpServer/pull/2052/files


pakrym:
Updated hardware server too.

PR #306, opened 2017-09-20
Added variant of JSON benchmarks using Jil

DamianEdwards:
Middleware and MVC versions of JSON scenario using Jil

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -5,8 +5,8 @@
     <CloudScenarios Include="-n Plaintext" />
     <CloudScenarios Include="-n MvcPlaintext" />
 
-    <CloudScenarios Include="-n Jil" />
     <CloudScenarios Include="-n Json" />
+    <CloudScenarios Include="-n Jil" />
     <CloudScenarios Include="-n MvcJson" />
     <CloudScenarios Include="-n MvcJil" />
 
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -5,8 +5,10 @@
     <CloudScenarios Include="-n Plaintext" />
     <CloudScenarios Include="-n MvcPlaintext" />
 
+    <CloudScenarios Include="-n Jil" />
     <CloudScenarios Include="-n Json" />
     <CloudScenarios Include="-n MvcJson" />
+    <CloudScenarios Include="-n MvcJil" />
 
     <CloudScenarios Include="-n DbSingleQueryRaw" />
     <CloudScenarios Include="-n DbSingleQueryDapper" />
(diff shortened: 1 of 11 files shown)


build/repo.targets
```diff
@@ -5,8 +5,10 @@
     <CloudScenarios Include="-n Plaintext" />
     <CloudScenarios Include="-n MvcPlaintext" />
 
+    <CloudScenarios Include="-n Jil" />
```
mikeharder: Move below `Json` for consistency.

build/repo.targets
```diff
@@ -60,12 +62,24 @@
     <PhysicalScenarios Include="-n Json -m https --kestrelTransport Sockets" />
     <PhysicalScenarios Include="-n Json -m https --webHost HttpSys" />
 
+    <PhysicalScenarios Include="-n Jil" />
+
+    <PhysicalScenarios Include="-n Jil --kestrelTransport Sockets" />
```
mikeharder: Would it be sufficient to just test `Jil` and compare to `Json`?  I don't think the other variants of this scenario are worth the cost.

src/Benchmarks/Configuration/Scenarios.cs
```diff
@@ -41,12 +41,18 @@ public Scenarios(IScenariosConfiguration scenariosConfiguration)
         [ScenarioPath("/json")]
         public bool Json { get; set; }
 
+        [ScenarioPath("/jil")]
```
mikeharder: New scenarios must also be added to `src/Benchmarks.ServerJob/Scenario.cs` and `src/BenchmarksDriver/Program.cs`.  See https://github.com/aspnet/benchmarks/pull/301/files.

mikeharder:
@DamianEdwards: Should the `Jil` scenarios also be added to the TechEmpower repo?

We will also need to update our PowerBI dashboards to include `Jil` in the `Json` charts.

src/Benchmarks/Benchmarks.csproj
```diff
@@ -17,6 +17,7 @@
 
   <ItemGroup>
     <PackageReference Include="Dapper" Version="$(DapperVersion)" />
+    <PackageReference Include="Jil" Version="2.15.4" />
```
mikeharder: Move version number to `build/dependencies.props` and reference with a variable.

src/Benchmarks/Controllers/HomeController.cs
```diff
@@ -18,12 +19,19 @@ public IActionResult Plaintext()
...
-        [Produces("application/json")]
+        [Produces("application/json", Type = typeof(JsonOutputFormatter))]
         public object Json()
         {
             return new { message = "Hello, World!" };
         }
-        
+
+        [HttpGet("jil")]
+        [Produces("application/json", Type = typeof(JilOutputFormatter))]
```
rynowak: This doesn't do what you want. The `Type` is there for API Explorer

My suggestion:
1. DI MVC options
2. Get the formatters from options
3. `return new ObjectResult(....) { Formatters { myformatter } };`

build/repo.targets
DamianEdwards: Fair enough, we can always add it back later.

DamianEdwards:
@mikeharder let's run them here first before we decide to add them to official benchmarks.

src/Benchmarks/Controllers/HomeController.cs
DamianEdwards: Hmmm, little concerned how much farther away from "idiomatic" this takes us. Is there not a way to do this outside of the controller?

src/Benchmarks/Configuration/Scenarios.cs
DamianEdwards: Done.

src/BenchmarksDriver/Program.cs
```diff
@@ -71,11 +80,13 @@ public class Program
             {
                 { Scenario.Plaintext, _plaintextJob },
                 { Scenario.Json, _jsonJob },
+                { Scenario.Jil, _jilJob },
```
mikeharder: If there are no differences between `_jsonJob` and `_jilJob` (I didn't see any), you can delete `_jilJob` and just use `_jsonJob`.

build/repo.targets
mikeharder: @DamianEdwards or @sebastienros: Please remove the extra scenarios before merging.  Let's just run `Jil` and `MvcJil` on Physical.

src/Benchmarks/Controllers/HomeController.cs
rynowak: You can use a filter, but you need to do the same dance. The JSON formatter isn't trivially constructable.

It's not really *idomatic* to use two separate JSON formatters in an app. We originally had a feature like this but we cut it for simplification purposes.

DamianEdwards: Fair point, hmmm. Perhaps I'll make it replace the JSON formatter instead.

DamianEdwards:
@sebastienros are we getting this one in soon?

PR #307, opened 2017-09-21
Adding more metrics

sebastienros:


build/dependencies.props
```diff
@@ -5,7 +5,7 @@
     <InternalAspNetCoreSdkVersion>2.1.1-*</InternalAspNetCoreSdkVersion>
     <JsonNetVersion>10.0.1</JsonNetVersion>
     <NETStandardImplicitPackageVersion>2.0.0-*</NETStandardImplicitPackageVersion>
-    <RuntimeFrameworkVersion Condition="'$(TargetFramework)'=='netcoreapp2.0'">2.0.0-*</RuntimeFrameworkVersion>
+    <RuntimeFrameworkVersion>2.0.0-*</RuntimeFrameworkVersion>
```
mikeharder: Why did you change this?

mikeharder: Why did you change this?

src/Benchmarks.ClientJob/Latency.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Benchmarks.ClientJob
+{
+    public class Latency
+    {
+        public double Average { get; set; }
```
mikeharder: I would use `TimeSpan` unless there's a good reason to use `double`.

src/BenchmarksClient/Startup.cs
```diff
@@ -34,6 +35,13 @@ public void ConfigureServices(IServiceCollection services)
         public void Configure(IApplicationBuilder app)
         {
             app.UseMvc();
+
+            app.Map("", builder =>
```
mikeharder: Why was this added?

build/dependencies.props
sebastienros: I assumed now we are only having `netcoreapp2.0`.

src/BenchmarksClient/Startup.cs
sebastienros: Diagnistics page, just to be sure the app is up and also to be sure the url is valid when configuring docker.

src/Benchmarks.ClientJob/Latency.cs
sebastienros: I was averaging values, and had to convert to ms before. I can return a TimeSpan back, ok.

build/dependencies.props
mikeharder: I had forgotten, but you are correct we only target `netcoreapp2.0` as of https://github.com/aspnet/benchmarks/pull/262.  So it's fine to clean this up.

src/BenchmarksClient/Startup.cs
mikeharder: Please add a comment explaining this.

src/BenchmarksClient/Startup.cs
```diff
@@ -209,13 +217,58 @@ private static Process StartProcess(ClientJob job)
...
+
+                double p99 = -1;
+                var p99Match = new Regex(@"99%\s*([\d\.]*)").Match(job.Output, latencyDistributionIndex);
+                if (p99Match.Success && p99Match.Groups.Count == 2)
+                {
+                    double.TryParse(p99Match.Groups[1].Value, out p99);
+                }
+                job.Latency.P99 = p99;
+
+
```
mikeharder: NIT: Remove extra whitespace line.

src/BenchmarksServer/Startup.cs
```diff
@@ -217,6 +221,28 @@ public static int Main(string[] args)
                                 Debug.Assert(process == null);
                                 process = StartProcess(hostname, database, sqlConnectionString, Path.Combine(tempDir, benchmarksDir),
                                     job, dotnetHome);
+                                
+                                if (timer != null)
```
mikeharder: I don't think this is necessary since the timer should be disposed an set to null when deleting the job.

src/BenchmarksServer/Startup.cs
```diff
@@ -217,6 +221,28 @@ public static int Main(string[] args)
...
+                                }
+
+                                DateTime lastMonitorTime = DateTime.UtcNow;
+                                var oldCPUTime = new TimeSpan(0);
+
+                                timer = new Timer(_ => 
+                                {
+
+                                    var newCPUTime = process.TotalProcessorTime;
+                                    var ellapsed = DateTime.UtcNow.Subtract(lastMonitorTime).TotalMilliseconds;
```
mikeharder: Typo `ellapsed` -> `elapsed`

src/BenchmarksServer/Startup.cs
```diff
@@ -217,6 +221,28 @@ public static int Main(string[] args)
                                 Debug.Assert(process == null);
                                 process = StartProcess(hostname, database, sqlConnectionString, Path.Combine(tempDir, benchmarksDir),
                                     job, dotnetHome);
+                                
+                                if (timer != null)
+                                {
+                                    timer.Dispose();
+                                }
+
+                                DateTime lastMonitorTime = DateTime.UtcNow;
```
mikeharder: `DateTime` -> `var`

src/BenchmarksServer/Startup.cs
```diff
@@ -217,6 +221,28 @@ public static int Main(string[] args)
...
+                                var oldCPUTime = new TimeSpan(0);
+
+                                timer = new Timer(_ => 
+                                {
+
+                                    var newCPUTime = process.TotalProcessorTime;
+                                    var ellapsed = DateTime.UtcNow.Subtract(lastMonitorTime).TotalMilliseconds;
+                                    var cpu =  Math.Round((newCPUTime - oldCPUTime).TotalMilliseconds / (Environment.ProcessorCount *  ellapsed) * 100);
+                                    lastMonitorTime = DateTime.UtcNow;
```
mikeharder: Technically I think you should only call DateTime.UtcNow once and assign it to a local var, like you do with `newCPUTime`.  Otherwise you could miss some elapsed time between the two calls to DateTime.UtcNow.

src/BenchmarksServer/Startup.cs
mikeharder: NIT: I would either change the name to `elapsedMilliseconds`, or else keep the type as `TimeSpan` and only convert to milliseconds when computing `cpu`, like you do with `newCPUTime` and `oldCPUTime`.

src/Benchmarks.ServerJob/ServerJob.cs
```diff
@@ -33,7 +34,14 @@ public class ServerJob : IIdentifiable
...
         public Scheme Scheme { get; set; }
-
+        public string AspNetCoreVersion { get; set; }
+        public string RuntimeFrameworkVersion { get; set; }
+        public string Session { get; set; }
+        public string Description { get; set; }
+        public TimeSpan Startup { get; set; }
+        public TimeSpan Latency { get; set; }
+        public List<long> WorkingSets { get; set; } = new List<long>();
+        public List<double> Cpus { get; set; } = new List<double>();
```
mikeharder: I think it would be slightly better to associate the elapsed time with each `WorkingSet` and `CPU` data point, so it would be possible to say create a graph of them over time.  Even if we don't store this in SQL today.  So I would change these to something like `List<Tuple<TimeSpan, long>>` or create a custom type.

src/BenchmarksServer/Startup.cs
```diff
@@ -319,10 +351,28 @@ private static string CloneRestoreAndBuild(string path, ServerJob job, string do
                 ["PATH"] = dotnetExeLocation + Path.PathSeparator + Environment.GetEnvironmentVariable("PATH")
             };
 
+            // Update the package and framework version in the props file
+
+            var dependenciesProps = Path.Combine(path, "benchmarks", "build", "dependencies.props");
```
mikeharder: First, I think it's cleaner to modify MSBuild properties by passing them on the command line to `build.cmd|sh`, rather than modifying the `dependencies.props` file.  Second, I think you should modify the `Benchmarks*` properties that were explicitly added to make it easier to test other versions (https://github.com/aspnet/benchmarks/commit/8fa6759ea7ab145640a65b5739b8e6dd3df76bfd#diff-30dcc9789f8ea3c7632e4f054e60c072R12).  @dougbu Also has experience with this.

src/BenchmarksServer/Startup.cs
```diff
@@ -560,13 +619,40 @@ private static string GetDotNetExecutable(string dotnetHome)
 
                     if (job.State == ServerState.Starting && e.Data.Contains("Application started"))
                     {
-                        job.State = ServerState.Running;
-                        job.Url = ComputeServerUrl(hostname, job.Scheme, job.Scenario);
                         Log.WriteLine($"Running job '{job.Id}' with scenario '{job.Scenario}'");
+                        job.Url = ComputeServerUrl(hostname, job.Scheme, job.Scenario);
+                        Log.WriteLine("Measuring startup time");
+                        
+                        using(var response = _httpClient.GetAsync(job.Url).GetAwaiter().GetResult())
```
mikeharder: Is it correct to wait until `OutputDataReceived` before executing the request?  Is it possible the server can respond to requests before it writes to stdout?

src/BenchmarksServer/Startup.cs
```diff
@@ -560,13 +619,40 @@ private static string GetDotNetExecutable(string dotnetHome)
...
+                        }
+
+                        // Measure a low-load latency by getting 5 samples synchronously
+                        // This could be done during the Client job but we are already measuring the Startup time here.
+                        const int LatencySamples = 10;
+                        var latencies = new List<TimeSpan>();
+                        for (var i = 0; i < LatencySamples; i++)
+                        {
+                            stopwatch.Restart();
+                            using(var response = _httpClient.GetAsync(job.Url).GetAwaiter().GetResult())
```
mikeharder: Could `Task.GetAwaiter().GetResult()` be replaced by just `Task.Result`?

mikeharder: Even better, `ProcessJobs()` is already async, so I think `StartProcess()` should also be made async, then you can just `await` rather than `Task.Result`.

src/BenchmarksServer/Startup.cs
mikeharder: Actually, I think the startup and latency code should be moved out of `StartProcess()` completely since they are not related.  I would move the code into `ProcessJobs()` after the call to `StartProcess()`.

src/BenchmarksServer/Startup.cs
benaadams: `Task.Result` returns exception wrapped in an AggregateException, `Task.GetAwaiter().GetResult()` returns actual exception

src/BenchmarksServer/Startup.cs
```diff
@@ -560,13 +619,40 @@ private static string GetDotNetExecutable(string dotnetHome)
...
+                            {
+                                var responseContent = response.Content.ReadAsStringAsync().GetAwaiter().GetResult();
+                                latencies.Add(stopwatch.Elapsed);
+                            }
+                        }
+
+                        // Ignore highest and lowest values
```
mikeharder: I would return all latency data from Server to Driver, and let the Driver decide how to process them and write the results to SQL.  I would definitely not discard the high and low values, since high latency should be measured rather than ignored.

In order of complexity:

1. Driver averages all data points and writes a single column to SQL.
2. Driver writes min, max, and average to SQL in 3 columns.
3. Client uses a more sophisticated latency tool with thousands of requests over many seconds, and sends back enough data to create a latency chart showing latency by percentile.

src/Benchmarks.ServerJob/ServerJob.cs
```diff
@@ -33,7 +34,14 @@ public class ServerJob : IIdentifiable
 
         [JsonConverter(typeof(StringEnumConverter))]
         public Scheme Scheme { get; set; }
-
+        public string AspNetCoreVersion { get; set; }
+        public string RuntimeFrameworkVersion { get; set; }
+        public string Session { get; set; }
```
mikeharder: Does the `ServerJob` need to know the `Session` and `Description`, or could these just live in the `Driver`?

src/Benchmarks.ServerJob/ServerJob.cs
sebastienros: Was too lazy to create a dedicated class (better than a tuple for JSon), but will do then.

src/BenchmarksClient/Startup.cs
```diff
@@ -209,13 +217,58 @@ private static Process StartProcess(ClientJob job)
...
+                // Start Latency Distribution pattent matching after a specific index as 
+                // previous results could render a .75% result for instance
+                var latencyDistributionIndex = job.Output.IndexOf("Latency Distribution");
+
+                double p50 = -1;
+                var p50Match = new Regex(@"50%\s*([\d\.]*)").Match(job.Output, latencyDistributionIndex);
+                if (p50Match.Success && p50Match.Groups.Count == 2)
+                {
+                    double.TryParse(p50Match.Groups[1].Value, out p50);
```
mikeharder: If `TryParse()` fails, the client will silently set latency to `-1`.  If parsing should never fail, I think it would be better for the client to throw and shutdown.  This should prevent the driver from writing invalid latency data to SQL.

mikeharder: And yes, we should do the same for RPS.  This was a bug before your changes.

src/BenchmarksDriver/Program.cs
```diff
@@ -156,6 +156,14 @@ public static int Main(string[] args)
                 "-w|--webHost",
                 "WebHost (Kestrel or HttpSys). Default is Kestrel.",
                 CommandOptionType.SingleValue);
+            var aspnetCoreVersionOption = app.Option("--aspnetCoreVersion",
+                "ASP.NET Core version (2.0.0 or 2.1.0-*).  Default is 2.1.0-*.", CommandOptionType.SingleValue);
+            var runtimeFrameworkVersionOption = app.Option("--runtimeFrameworkVersion",
```
mikeharder: I think this can be condensed down to a single version with values like "2.0.0", "2.0.1", and "latest".  If the version is "latest", we just use whatever is set in the `Benchmarks` branch we clone and build.  If the version is say "2.0.0", we should set the environment variables `KOREBUILD_DOTNET_SHARED_RUNTIME_VERSION` and `KOREBUILD_DOTNET_VERSION` to control the runtime and CLI installed by KoreBuild, and also the MSBuild variables `BenchmarksAspNetCoreVersion`, `BenchmarksRuntimeFrameworkVersion` (and possibly `BenchmarksNETStandardImplicitPackageVersion` and `BenchmarksNETStandardLibraryNETFrameworkVersion` if they're still relevant).

src/BenchmarksDriver/Program.cs
```diff
@@ -191,6 +199,26 @@ public static int Main(string[] args)
...
+                var runtimeFrameworkVersion = runtimeFrameworkVersionOption.Value();
+                if (string.IsNullOrEmpty(runtimeFrameworkVersion))
+                {
+                    runtimeFrameworkVersion = "2.0.0";
+                }
+
+                var session = sessionOption.Value();
+                if (String.IsNullOrEmpty(session)) 
+                {
+                    session = Guid.NewGuid().ToString("n");
```
mikeharder: Is there any point to using `Guid.NewGuid()` for session if not specified?  Would it be better to leave it as `null`?

src/BenchmarksDriver/Program.cs
```diff
@@ -402,26 +434,71 @@ public static int Main(string[] args)
...
+
+                    serverJob = JsonConvert.DeserializeObject<ServerJob>(responseContent);
+
+                    Log("Writing results to SQL...");
+
+                    await WriteJobsToSql(
+                        serverJob: serverJob, 
+                        clientJob: clientJob,
                         connectionString: sqlConnectionString,
                         scenario: scenario,
```
mikeharder: `scenario` is also available as `serverJob.Scenario`.  I would just remove the local variable entirely and replace it with `serverJob.Scenario`.

src/BenchmarksDriver/Program.cs
mikeharder: I realized it's probably to correlate the different metrics for a given job.

build/dependencies.props
dougbu: Yeah, the only thing that's interesting about pre-2.0.0 releases is running the Benchmarks site against 1.x bits. We can change `$(BenchmarksRuntimeFrameworkVersion)` to do that.

src/BenchmarksDriver/Program.cs
```diff
@@ -539,17 +648,21 @@ public static int Main(string[] args)
...
         {
-            Log("Writing results to SQL...");
-
             const string createCmd =
                 @"
                 IF OBJECT_ID(N'dbo.AspNetBenchmarks', N'U') IS NULL
                 BEGIN
                     CREATE TABLE [dbo].[AspNetBenchmarks](
                         [Id] [int] IDENTITY(1,1) NOT NULL PRIMARY KEY,
+                        [Excluded] [bit] NOT NULL,
```
mikeharder: Should we change this column to `DEFAULT 0`, rather than inserting 0 every time?  It would only be set to `1` manually as needed to cleanup bad data.

src/Benchmarks.ServerJob/Framework.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Benchmarks.ServerJob
+{
+    public enum AspNetCoreVersion
```
dougbu: Why be this specific? The version chosen when running Benchmarks is already controllable using `ServerJob.Sources` and changes to `dependencies.props` in the named branches.

src/Benchmarks.ClientJob/Latency.cs
```diff
@@ -0,0 +1,14 @@
...
+        public double P50 { get; set; }
+        public double P75 { get; set; }
+        public double P90 { get; set; }
+        public double P99 { get; set; }
```
dougbu: I know `wrk` is this terse. But, suggest expanding these names to something like `Within50thPercentile` &hellip; to avoid future confusion.

src/BenchmarksServer/Startup.cs
dougbu: Yes, this (as well as branches for testing Benchmarks against earlier runtimes) is why we use the separate projects in `Benchmarks.csproj`.

src/BenchmarksServer/Startup.cs
```diff
@@ -217,6 +221,28 @@ public static int Main(string[] args)
...
+
+                                DateTime lastMonitorTime = DateTime.UtcNow;
+                                var oldCPUTime = new TimeSpan(0);
+
+                                timer = new Timer(_ => 
+                                {
+
+                                    var newCPUTime = process.TotalProcessorTime;
+                                    var ellapsed = DateTime.UtcNow.Subtract(lastMonitorTime).TotalMilliseconds;
+                                    var cpu =  Math.Round((newCPUTime - oldCPUTime).TotalMilliseconds / (Environment.ProcessorCount *  ellapsed) * 100);
```
dougbu: So, this is actually a percentage? Suggest making the name more specific.

src/BenchmarksServer/Startup.cs
```diff
@@ -217,6 +221,28 @@ public static int Main(string[] args)
...
+                                    lastMonitorTime = DateTime.UtcNow;
+                                    oldCPUTime = newCPUTime;
+
+                                    job.WorkingSets.Add(process.WorkingSet64);
+                                    job.Cpus.Add(cpu);
+
+                                }, null, TimeSpan.FromTicks(0), TimeSpan.FromSeconds(1));
```
dougbu: Is this accurate enough? The lambda will execute after at least a second, perhaps throwing off averages of `WorkingSets` and `Cpus` values.

src/BenchmarksServer/Startup.cs
```diff
@@ -629,23 +629,20 @@ private static string GetDotNetExecutable(string dotnetHome)
...
-                        for (var i = 0; i < LatencySamples; i++)
+                        for (var i = 0; i < 10; i++)
                         {
                             stopwatch.Restart();
                             using(var response = _httpClient.GetAsync(job.Url).GetAwaiter().GetResult())
                             {
                                 var responseContent = response.Content.ReadAsStringAsync().GetAwaiter().GetResult();
-                                latencies.Add(stopwatch.Elapsed);
+
+                                // We keep the last measy to simulate a warmup phase.
```
mikeharder: Typo `measy`

src/Benchmarks.ServerJob/ServerCounter.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Benchmarks.ServerJob
+{
+    public class ServerCounter
+    {
+        public long WorkingSet { get; set; }
+        public double CpuPercentage { get; set; }
```
mikeharder: I would add a `TimeSpan Elapsed` property to indicate how soon after process start the metric was obtained.

src/Benchmarks.ServerJob/Framework.cs
mikeharder: Is this used anymore?  If not please remove it.

sebastienros:
updated

src/Benchmarks.ServerJob/ServerCounter.cs
```diff
@@ -1,10 +1,13 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+
 namespace Benchmarks.ServerJob
 {
     public class ServerCounter
     {
+        public TimeSpan Ellapsed { get; set; }
```
mikeharder: Typo

sebastienros: I guess I don't learn from my mistakes

PR #309, opened 2017-09-25
Don't use runtime store by default

sebastienros:


src/BenchmarksServer/Startup.cs
```diff
@@ -421,8 +422,18 @@ private static string CloneRestoreAndBuild(string path, ServerJob job, string do
...
+
+            if (job.EnableRuntimeStore)
+            {
+                ProcessUtil.Run(dotnetExecutable, $"build -c Release", throwOnError: false, workingDirectory: benchmarksApp, environmentVariables: env);
+            }
+            else
+            {
+                // Publishing the application will prevent the runtime store from being used as it copies the assemblies in the published folder
+                ProcessUtil.Run(dotnetExecutable, $"publish -c Release -o {Path.Combine(benchmarksApp, "published")}", throwOnError: false, workingDirectory: benchmarksApp, environmentVariables: env);
+            }
```
mikeharder: @JunTaoLuo: Is there a more explicit way to enable/disable runtime store?  Relying on the different behavior of `dotnet build` and `dotnet publish` seems fragile.

JunTaoLuo: Not really. The binaries are loaded given a hierarchy. If the store is installed (beside the dotnet.exe or in the global installed location), it will be used. The only way to bypass it is to deploy app local which takes precedence over the installed runtime stores.

mikeharder: OK, sounds like using `build` and `publish` is currently the best way to toggle between enabling and disabling the store.

src/BenchmarksDriver/Program.cs
```diff
@@ -162,6 +162,8 @@ public static int Main(string[] args)
                 "A logical identifier to group related jobs.", CommandOptionType.SingleValue);
             var descriptionOption = app.Option("--description",
                 "The description of the job.", CommandOptionType.SingleValue);
+            var enableRuntimeStoreOption = app.Option("--enableRuntimeStore",
+                "Runs the benchmarks using the runtime store is available.", CommandOptionType.NoValue);
```
mikeharder: Typo `is` instead of `if`?

mikeharder:
@sebastienros: Are you planning to use this switch for only ad-hoc runs, or also automated runs?  If the latter, we will need to add this column to SQL as well.

src/BenchmarksServer/Startup.cs
```diff
@@ -397,6 +397,7 @@ private static string CloneRestoreAndBuild(string path, ServerJob job, string do
             }
 
             // Source dependencies are always built using KoreBuild
+            // Source dependencies are always built using KoreBuild
```
JunTaoLuo: Why the duplicated comment here?

sebastienros:
Merging this change will make jobs run without runtime store by default. Because we were actually using it until now, this should have a visible impact on the graphs. 

As long as we agree we'll always run the jobs without the runtime store and that there is nothing to detect automatically by comparing both then ad-hoc scenarios only is fine and we don't need a flag.

PR #311, opened 2017-09-25
Change column types to allow indices

sebastienros:
With the increase of data points for each new job, and the fact that graphs will need more filters, adding indices should have a significant impact on perf. The columns that need it will require to have more specific limits and get away from the TEXT(nvarchar(max)) type.


sebastienros:
updated

src/BenchmarksDriver/Program.cs
```diff
@@ -724,9 +724,20 @@ private static Task WriteJobsToSql(ServerJob serverJob, ClientJob clientJob, str
...
                         [Value] [float] NOT NULL
                     )
+
+                    CREATE INDEX IDX_Scenario_Hardware_OS_Description_Dimension
+                    ON AspNetBenchmarks
+                    (
+                        [Scenario],
+                        [Hardware],
+                        [OperatingSystem],
+                        [Description],
```
mikeharder: Why index `Description`?

mikeharder:
@sebastienros: Do we know if these indexes improve PowerBI perf?  We should be able to identify the queries used by PowerBI, either via SQL Azure Auditing or another mechanism, and then test the queries before and after the indexes.

sebastienros:
Ok, I will get some numbers.

src/BenchmarksDriver/Program.cs
sebastienros: This is to get only the data points that are run with the CI job. If we get more automatic data points like 2.0/2.1 comparisons, we don't want these to be in the graphs.

sebastienros: BTW I have another branch (not yet PR'ed) to set this description as part of the CI job.

PR #314, opened 2017-10-02
Adding --useRuntimeStore option

sebastienros:
By default it's not using the runtime store anymore

### DIFF
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -763,7 +763,8 @@ namespace BenchmarksDriver
                         method: clientJob.Method,
                         headers: clientJob.Headers,
                         dimension: dimension,
-                        value: value);
+                        value: value,
+                        runtimeStore: serverJob.UseRuntimeStore);
         }
         private static async Task WriteResultsToSql(
             string connectionString,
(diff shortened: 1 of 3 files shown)


mikeharder:
@sebastienros: I think we should somehow indicate in our DB that older tests had the runtime store enabled.

sebastienros:
Use the description, add "/RuntimeStore" to all existing rows. This will let us use them or not in the graphs.

sebastienros:
Unless you prefer a new field. As we have a new argument that would make sense.

mikeharder:
Wouldn't it be cleaner to use a dedicated column rather than a text description?

sebastienros:
Doing it, we agree.

src/BenchmarksDriver/Program.cs
```diff
@@ -799,6 +807,7 @@ private static Task WriteJobsToSql(ServerJob serverJob, ClientJob clientJob, str
                         [Hardware] [nvarchar](max) NOT NULL,
                         [OperatingSystem] [nvarchar](max) NOT NULL,
                         [Framework] [nvarchar](max) NOT NULL,
+                        [RuntimeStore] [bit] DEFAULT 1,
```
mikeharder: I think this column should be `NOT NULL` rather than `DEFAULT 1`.  It's dangerous to default to either 1 or 0.  When creating the column in the prod DB, you can set a default value to convert all existing rows, then remove the default so new rows must specify.

PR #327, opened 2017-10-04
Update compiler

pakrym:
https://github.com/aspnet/KestrelHttpServer/pull/2090/files


pakrym:
Merging them simultaneously

sebastienros:
The issue with "simultaneously" is that it's only if it gets deployed on the Server ;) ... which I will do now I assume

PR #329, opened 2017-10-05
Progress towards this fully working on osx

davidfowl:
- OSX has some issues getting the total process time and killing a process tree.
- Did some other general code cleanup


sebastienros:
I'll merge when approved, and redeploy the server.

PR #330, opened 2017-10-06
Use RunAsync in the client

davidfowl:
Using new APIs


davidfowl:
I'm going to yolo this

PR #331, opened 2017-10-06
Initial docker support

davidfowl:
- Support passing a docker file as the source of a benchmark
- The steps we do with this change are:
 - Clone the repo
 - Build a random image name
 - Start the container with an exposed port
 - Use docker container stats to get the status of the container we started
- Clean up when finished
- Also fixed some race conditions with the timer callback.

I only tested locally which means the numbers were all out of wack as client/server and driver were all on the same machine. I'll make sure i didn't break any mainline scenarios as well. The code generally needs some refactoring as it's getting to be lots of spaghetti in Startup.

PS: Args need to be added, they don't work as yet.

### DIFF
--- a/src/BenchmarksServer/Startup.cs
+++ b/src/BenchmarksServer/Startup.cs
@@ -264,7 +264,7 @@ namespace BenchmarkServer
 
                                     try
                                     {
-                                        if (disposed || Debugger.IsAttached)
+                                        if (disposed)
                                         {
                                             return;
                                         }
--- a/src/BenchmarksServer/Startup.cs
+++ b/src/BenchmarksServer/Startup.cs
@@ -402,7 +402,7 @@ namespace BenchmarkServer
         {
             var source = job.Source;
             // Docker image names must be lowercase
-            var imageName = $"{Path.GetDirectoryName(source.Project)}_{Guid.NewGuid()}".ToLowerInvariant();
+            var imageName = $"benchmarks_{Path.GetDirectoryName(source.Project)}".ToLowerInvariant();
             var cloneDir = Path.Combine(path, Git.Clone(path, source.Repository));
 
             if (!string.IsNullOrEmpty(source.BranchOrCommit))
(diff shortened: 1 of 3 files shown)


src/BenchmarksServer/Startup.cs
```diff
@@ -230,34 +234,94 @@ public static int Main(string[] args)
...
-                                        Elapsed = now - startMonitorTime,
-                                        WorkingSet = process.WorkingSet64,
-                                        CpuPercentage = cpu
-                                    });
+                                        return;
+                                    }
 
+                                    try
+                                    {
+                                        if (disposed || Debugger.IsAttached)
```
mikeharder: Why `Debugger.IsAttached`?

src/BenchmarksServer/Startup.cs
```diff
@@ -328,6 +398,38 @@ void CleanJob()
...
+            var cloneDir = Path.Combine(path, Git.Clone(path, source.Repository));
+
+            if (!string.IsNullOrEmpty(source.BranchOrCommit))
+            {
+                Git.Checkout(cloneDir, source.BranchOrCommit);
+            }
+
+            ProcessUtil.Run("docker", $"build -t {imageName} -f {source.DockerFile} .", cloneDir);
+
+            var result = ProcessUtil.Run("docker", $"run -d -p {job.Port}:{job.Port} {imageName}");
```
mikeharder: This should use `--network host` instead of `-p port:port` for improved performance.  Also, if you add `--rm`, then the container will be automatically removed after it's stopped, so you will no longer need `docker container rm`.

src/BenchmarksServer/Startup.cs
davidfowl: Debugging this code is impossible, but I'm not sure this check is 100% required now that we have the lock.

davidfowl:
🆙 📅

src/BenchmarksServer/Startup.cs
```diff
@@ -412,7 +412,12 @@ private static (string containerId, string imageName) DockerBuildAndRun(string p
 
             ProcessUtil.Run("docker", $"build -t {imageName} -f {source.DockerFile} .", cloneDir);
 
-            var result = ProcessUtil.Run("docker", $"run -d -p {job.Port}:{job.Port} {imageName}");
+            // OSX doesn't support host networking
+            var useHostNetworking = OperatingSystem != OperatingSystem.OSX;
```
mikeharder: Change this to `useHostNetworking = OperatingSystem == OperatingSystem.Linux`.

PR #334, opened 2017-10-06
Adding support for MySql and SqlServer

sebastienros:


docker/benchmarks/run-server.sh
```diff
@@ -14,27 +14,34 @@ fi
...
+    /benchmarks/src/BenchmarksServer/bin/Debug/netcoreapp2.0/BenchmarksServer.dll \
+    -n $server_ip \
+    --hardware $hardware \
+    $database \
+    $sql \
+    $@"
 else
     echo DBHOST needs to be defined
-fi
+fi
```
mikeharder: Missing newline

docker/benchmarks/run-server.sh
```diff
@@ -14,27 +14,34 @@ fi
...
-        -e POSTGRES_CONN \
-        -e MYSQL_CONN \
-        -e MSSQL_CONN \
-        benchmarks \
-        bash -c \
-        "/root/.dotnet/dotnet \
-        /benchmarks/src/BenchmarksServer/bin/Debug/netcoreapp2.0/BenchmarksServer.dll \
-        -n $server_ip \
-        --hardware $hardware"
+POSTGRES_CONN = 'Server=$DBHOST;Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=1024;NoResetOnClose=true'
```
mikeharder: Missing indentation

docker/benchmarks/run-server.sh
```diff
@@ -14,27 +14,34 @@ fi
...
+MYSQL_CONN = 'Server=$DBHOST;Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass'
+MSSQL_CONN = 'Server=$DBHOST;Database=hello_world;User Id=sa;Password=Benchmarkdbp@55'
```
mikeharder: Is there a reason we can't use the same username and password for ms sql?

docker/benchmarks/run-server.sh
```diff
@@ -38,3 +42,6 @@ docker run \
     $database \
```
mikeharder: If the `database` var has been removed, this usage should be removed as well.

docker/benchmarks/run-server.sh
```diff
@@ -38,3 +42,6 @@ docker run \
...
     $sql \
     $@"
+else
+    echo DBHOST needs to be defined
+fi
```
mikeharder: Missing newline.

docker/benchmarks/run-server.sh
```diff
@@ -14,9 +14,10 @@ fi
 
 if [[ -v DBHOST ]]
 then
-    database="--database PostgreSql"
-    sql="--sql \"Server=$DBHOST;Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=1024;NoResetOnClose=true\""
-fi
+
+POSTGRES_CONN = 'Server=$DBHOST;Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=1024;NoResetOnClose=true'
```
mikeharder: Missing indentation.

docker/mssql-techempower/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM microsoft/mssql-server-linux:2017-latest
```
mikeharder: I think we should use `2017-GA` instead of `2017-latest`.  It's generally better to use specific versions for consistency and test perf when upgrading.  Can you also please change `postgres:9` to `postgres:9.6.5`?

docker/mssql-techempower/entrypoint.sh
```diff
@@ -0,0 +1,2 @@
+#start SQL Server, start the script to create the DB and import the data
+/opt/mssql/bin/sqlservr & /usr/src/app/import-data.sh
```
mikeharder: Missing newline.

docker/mssql-techempower/create.sql
```diff
@@ -0,0 +1,47 @@
...
+INSERT INTO [fortune] ([message]) VALUES ('A bad random number generator: 1, 1, 1, 1, 1, 4.33e+67, 1, 1, 1');
+INSERT INTO [fortune] ([message]) VALUES ('A computer program does what you tell it to do, not what you want it to do.');
+INSERT INTO [fortune] ([message]) VALUES ('Emacs is a nice operating system, but I prefer UNIX. — Tom Christaensen');
+INSERT INTO [fortune] ([message]) VALUES ('Any program that runs right is obsolete.');
+INSERT INTO [fortune] ([message]) VALUES ('A list is only as strong as its weakest link. — Donald Knuth');
+INSERT INTO [fortune] ([message]) VALUES ('Feature: A bug with seniority.');
+INSERT INTO [fortune] ([message]) VALUES ('Computers make very fast, very accurate mistakes.');
+INSERT INTO [fortune] ([message]) VALUES ('<script>alert("This should not be displayed in a browser alert box.");</script>');
+INSERT INTO [fortune] ([message]) VALUES ('フレームワークのベンチマーク');
+GO
```
mikeharder: Missing newline.

docker/mssql-techempower/import-data.sh
```diff
@@ -0,0 +1,5 @@
+#wait for the SQL Server to come up
+sleep 90s
```
mikeharder: I would try to find a way to avoid this 90s delay since it makes building the container quite slow, but it's OK for now.

docker/mysql-techempower/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM mysql:5
```
mikeharder: `FROM mysql:5.7.19`

docker/mysql-techempower/Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+
+ADD create.sql /docker-entrypoint-initdb.d/
+ADD my.cnf /etc/mysql/conf.d/
+RUN chmod 755 /docker-entrypoint-initdb.d/create.sql
+
+ENV MYSQL_USER=benchmarkdbuser \
+    MYSQL_PASSWORD=benchmarkdbpass \
+    MYSQL_ROOT_PASSWORD=benchmarkdbpass \
+	MYSQL_DATABASE=hello_world
```
mikeharder: Fix indentation

src/Benchmarks.ServerJob/ServerJob.cs
```diff
@@ -35,7 +35,8 @@ public class ServerJob : IIdentifiable
         public int Port { get; set; } = 5000;
         public string Path { get; set; } = "/";
         public string AspNetCoreVersion { get; set; } = "2.1.0-*";
-
+        public Database Database { get; set; } = Database.PostgreSql;
```
mikeharder: Would it be better to default this to `None`?  It's the default in `Benchmarks/Configuration/AppSettings.cs`.

src/BenchmarksServer/Startup.cs
```diff
@@ -834,17 +816,29 @@ private static string GetDotNetExecutable(string dotnetHome)
...
+                string connectionString = null;
 
-            if (!string.IsNullOrEmpty(sqlConnectionString))
-            {
-                process.StartInfo.Environment.Add("ConnectionString", sqlConnectionString);
-            }
+                switch (job.Database)
+                {
+                    case Database.PostgreSql:
+                        connectionString = Environment.GetEnvironmentVariable("POSTGRES_CONN");
```
mikeharder: I would pass these in as command-line parameters rather than environment variables for consistency and discoverability.

docker/mssql-techempower/import-data.sh
sebastienros: This is from the image documentation, we can go smaller but this is what they use.

src/Benchmarks.ServerJob/ServerJob.cs
sebastienros: I didn't want to break any current job definition as the new parameter is not specified for now. And I assumed you'd preferred defaults.

mikeharder: If we're going to run all 3 servers in our Cloud environment, there's no reason one server should be the default over the others.  So I would default to `None` and add `PostgreSql` to jobs we want to run on `PostgreSql`, like say the jobs in `build/repo.targets`.

src/BenchmarksServer/Startup.cs
sebastienros: We can't pass the values in the server command-line as each job can use a different one. So I assume you don't mean that.

We shouldn't pass it to the benchmarked application as command-line as we might have encoding issues while generating the command-line (double quotes) and we can load the database configuration more easily with `new ConfigurationBuilder().AddEnvironmentVariables()`
But if you really prefer this way then I'll do the changes and parse the command line, which we don't do at all right now in benchmarks.

mikeharder: I think the BenchmarksServer should have 3 command-line parameters like `postgresqlConnectionString`, `mysqlConnectionString`, and `sqlserverConnectionString`.  Currently, BenchmarksServer takes all input as command-line parameters rather than environment variables.  Command-line parameters are more discoverable since they are listed in `--help` documentation, easier to find in the source code, etc.

No change to how it should be passed to the benchmarked application.

PR #338, opened 2017-10-08
Various docker fixes

davidfowl:
- Wait for the docker image to listen on the port. We were marking the job as running too early and which means the client would try to connect before things were properly listening. This fixes that by trying to do a simple get request 5 times (with a fixed delay) before marking the job as running.
- Parsing the memory usage was incorrect for handling GiB.

### DIFF
--- a/src/BenchmarksServer/Startup.cs
+++ b/src/BenchmarksServer/Startup.cs
@@ -313,11 +313,15 @@ namespace BenchmarkServer
                                             // Format is {value}%
                                             var cpuPercentRaw = data[0];
 
-                                            // Format is {used}MiB/{total}MiB
+                                            // Format is {used}M/GiB/{total}M/GiB
                                             var workingSetRaw = data[1];
-                                            var usedMemoryRaw = workingSetRaw.Split('/')[0];
+                                            var usedMemoryRaw = workingSetRaw.Split('/')[0].Trim();
                                             var cpu = double.Parse(cpuPercentRaw.Trim('%'));
-                                            var workingSet = (long)(double.Parse(usedMemoryRaw.Substring(0, usedMemoryRaw.IndexOf('M'))) * 1048576);
+
+                                            // MiB or GiB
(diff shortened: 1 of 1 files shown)


src/BenchmarksServer/Startup.cs
```diff
@@ -424,11 +429,32 @@ private static (string containerId, string imageName) DockerBuildAndRun(string p
...
-            var command = useHostNetworking ? $"run -d --rm --network host {imageName}" : 
+            var command = useHostNetworking ? $"run -d --rm --network host {imageName}" :
                                               $"run -d --rm -p {job.Port}:{job.Port} {imageName}";
             var result = ProcessUtil.Run("docker", $"{command} {job.Arguments}");
             var containerId = result.StandardOutput.Trim();
 
+            // Wait until the service is reachable to avoid races where the container started but isn't
+            // listening yet. We only try 5 times, if it keeps failing we ignore it. If the port
+            // is unreachable then clients will fail to connect and the job will be cleaned up properly
+            using (var client = new HttpClient())
```
sebastienros: We already have a static `_httpClient` that you can reuse. The server is always on we should reuse the same instance.

src/BenchmarksServer/Startup.cs
```diff
@@ -424,11 +429,32 @@ private static (string containerId, string imageName) DockerBuildAndRun(string p
...
+            {
+                const int maxRetries = 5;
+                for (var i = 0; i < maxRetries; ++i)
+                {
+                    try
+                    {
+                        // We don't care if it's a 404, it just needs to not fail
+                        await client.GetAsync($"http://{hostname}:{job.Port}");
```
sebastienros: Maybe use `job.Url` as it's already computing what url to benchmark, and it contains the custom path value that can be set on the driver.

src/BenchmarksServer/Startup.cs
sebastienros: Not an issue actually, just taste.

PR #340, opened 2017-10-13
Batching db updates

sebastienros:
Fixing DbMultiUpdatesEf scenario by forcibly batching updates.

Before: Timeouts 
After: 200 rps

### DIFF
--- a/src/Benchmarks/Data/EfDb.cs
+++ b/src/Benchmarks/Data/EfDb.cs
@@ -16,13 +16,11 @@ namespace Benchmarks.Data
     {
         private readonly IRandom _random;
         private readonly ApplicationDbContext _dbContext;
-        private readonly bool _useBatchUpdate;
 
         public EfDb(IRandom random, ApplicationDbContext dbContext, IOptions<AppSettings> appSettings)
         {
             _random = random;
             _dbContext = dbContext;
-            _useBatchUpdate = appSettings.Value.Database == DatabaseServer.SqlServer;
         }
 
         private static readonly Func<ApplicationDbContext, int, Task<World>> _firstWorldQuery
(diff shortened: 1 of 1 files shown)


mikeharder:
@anpete added the `_useBatchUpdate` in https://github.com/aspnet/benchmarks/commit/dc37a6833591e6415f2994a68bc2f57d7c3fef3e.  @anpete, what do you think of this PR?

anpete:
I just moved it. I think disabling batching was there due to the Postgres deadlock issue. Has that gone away?

sebastienros:
What has changed since then is that this benchmark without batching now timeouts all requests under load, whereas the batched one works, not as good as the Raw test which can order the batch queries, but still better than failing the jobs.

anpete:
Do we know why?

mikeharder:
The original postgres code was added here: https://github.com/aspnet/benchmarks/commit/a7a2448fe2282969034f51d66b823a4558197934#diff-19e1c5a206d8c95a29b182c558128f40.  I don't know if there's been any change to the Postgres deadlock issue.

In EF, is there any way to sort the rows before performing the batch update?  Should EF do this by default, to prevent customers from hitting the same issue if they enable batching with Postgres?

anpete:
We may be sorting already: cc @AndriySvyryd for confirmation

AndriySvyryd:
We do sort the rows. That benchmark might have been written before this was in place.

mikeharder:
@sebastienros: Let's enable batching for EF then, and we'll see if we get any errors.

anpete:
>What has changed since then is that this benchmark without batching now timeouts all requests under load.

We should still try and understand why this started happening though, right?

PR #342, opened 2017-10-19
Reacting to Sockets changes

sebastienros:
These changes are intended to be compatible with the current database


src/Benchmarks.ServerJob/WebHost.cs
```diff
@@ -5,7 +5,8 @@ namespace Benchmarks.ServerJob
 {
     public enum WebHost
     {
-        Kestrel,
+        Sockets,
+        Libuv,
```
mikeharder: Lets rename to `KestrelSockets` and `KestrelLibuv` for clarity

PR #343, opened 2017-10-19
Adding a timeout on jobs

sebastienros:
In case the application throws an exception, or the driver gets killed and doesn't update the job.

### DIFF
--- a/src/Benchmarks.ServerJob/ServerJob.cs
+++ b/src/Benchmarks.ServerJob/ServerJob.cs
@@ -62,5 +62,6 @@ namespace Benchmarks.ServerJob
         public WebHost WebHost { get; set; }
 
         public bool UseRuntimeStore { get; set; }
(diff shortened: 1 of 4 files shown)


src/Benchmarks.ServerJob/ServerJob.cs
```diff
@@ -62,5 +62,6 @@ public class ServerJob : IIdentifiable
         public WebHost WebHost { get; set; }
 
         public bool UseRuntimeStore { get; set; }
+        public TimeSpan Timeout { get; set; } = TimeSpan.FromMinutes(5);
```
mikeharder: Should this be 2 minutes instead of 5?  Everything else says the default is 2 minutes.

mikeharder:
Do we need a similar timeout in BenchmarksClient?  I know `wrk` can hang indefinitely in certain cases.

PR #346, opened 2017-10-23
Update compiler

pakrym:
https://github.com/aspnet/KestrelHttpServer/pull/2121


sebastienros:
GTG as long as you merge Kestrel simultaneously (as usual)

PR #352, opened 2017-10-27
Making services more resilient

### LINKED ISSUE #345: Add a timeout for client jobs
wrk can get stuck sometimes

sebastienros:
Fixes #345

### DIFF
--- a/src/BenchmarksClient/Startup.cs
+++ b/src/BenchmarksClient/Startup.cs
@@ -158,7 +158,7 @@ namespace BenchmarkClient
                         {
                             // We can't wait for ever as the driver is informed that 
                             // the job is pending deletion, and the client would be blocked otherwise
-                            if (!process.WaitForExit((job.Duration + 5) * 1000))
+                            if (!process.WaitForExit((int)TimeSpan.FromSeconds(job.Duration + 5).TotalMilliseconds))
                             {
                                 process.Kill();
                             }
--- a/src/BenchmarksClient/Startup.cs
+++ b/src/BenchmarksClient/Startup.cs
@@ -7,7 +7,6 @@ using System.IO;
 using System.Linq;
 using System.Net.Http;
 using System.Reflection;
-using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
--- a/src/BenchmarksClient/Startup.cs
+++ b/src/BenchmarksClient/Startup.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Linq;
 using System.Net.Http;
 using System.Reflection;
+using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
@@ -154,9 +155,20 @@ namespace BenchmarkClient
 
                         Debug.Assert(process != null);
 
-                        process.WaitForExit();
-                        process.Dispose();
-                        process = null;
+                        try
+                        {
+                            // We can't wait for ever as the driver is informed that 
+                            // the job is pending deletion, and the client would be blocked otherwise
+                            if (!process.WaitForExit((job.Duration + 5) * 1000))
+                            {
+                                process.Kill();
+                            }
+                        }
+                        finally
+                        {
+                            process.Dispose();
+                            process = null;
+                        }
 
                         _jobs.Remove(job.Id);
                     }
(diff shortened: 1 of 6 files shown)


sebastienros:
Based on different failures that happened this week.

GitHub clones would fail
Server job timeout, but the driver just displaying a 404
Client job timeout, which would just block indefinitely

src/BenchmarksServer/Git.cs
```diff
@@ -27,12 +29,34 @@ public static string Clone(string path, string repository, string branch = null)
 
         public static void Checkout(string path, string branchOrCommit)
         {
-            RunGitCommand(path, $"checkout {branchOrCommit}");
+            RunGitCommand(path, $"checkout {branchOrCommit}", retries: 5);
+        }
+
+        private static ProcessResult RunGitCommand(string path, string command, bool throwOnError = true, int retries = 0)
+        {
+            return RetryOnException(retries, () => ProcessUtil.Run("git", command, workingDirectory: path, throwOnError: throwOnError, timeout: CommandTimeout));
```
sebastienros: Maybe this timeout should be one level up as each command might require a different timeout. Clone could be the only one to use it.

src/BenchmarksClient/Startup.cs
```diff
@@ -154,9 +154,20 @@ public static int Main(string[] args)
...
                         Debug.Assert(process != null);
 
-                        process.WaitForExit();
-                        process.Dispose();
-                        process = null;
+                        try
+                        {
+                            // We can't wait for ever as the driver is informed that 
+                            // the job is pending deletion, and the client would be blocked otherwise
+                            if (!process.WaitForExit((job.Duration + 5) * 1000))
```
mikeharder: NIT: Clearer to use `TimeSpan.FromSeconds(job.Duration + 5).TotalMilliseconds`.

src/BenchmarksClient/Startup.cs
```diff
@@ -154,9 +154,20 @@ public static int Main(string[] args)
 
                         Debug.Assert(process != null);
 
-                        process.WaitForExit();
-                        process.Dispose();
-                        process = null;
+                        try
```
mikeharder: What is the overall scenario where the client job does not exit?  Is it when `wrk` hangs because it can't contact the server?  In this scenario, when will the driver set the job to `deleting`?

src/BenchmarksServer/ProcessUtil.cs
```diff
@@ -12,7 +12,7 @@ namespace BenchmarkServer
     public static class ProcessUtil
     {
         public static ProcessResult Run(string filename, string arguments, string workingDirectory = null,
-            bool throwOnError = true, IDictionary<string, string> environmentVariables = null)
+            bool throwOnError = true, IDictionary<string, string> environmentVariables = null, int timeout = int.MaxValue)
```
mikeharder: I would use `TimeSpan? timeout = null` instead of `int timeout = int.MaxValue`.  If `timeout` is null, call `WaitForExit()`, else call `WaitForExit(timeout.TotalMilliseconds)`.

src/BenchmarksServer/Git.cs
```diff
@@ -8,6 +8,8 @@ namespace BenchmarkServer
 {
     public static class Git
     {
+        private const int CommandTimeout = 5000;
```
mikeharder: Change to `TimeSpan`.  Also, cloning a large repo may take a fairly long time.  Would 30 seconds be a better timeout?

src/BenchmarksClient/Startup.cs
sebastienros: If the CI stops the driver because it times out, the client job is not deleted and then block all upcoming ones.

mikeharder: How does this code change fix that scenario?

sebastienros: Right, that can't be it. I know when I wrote the code that the status of the job was "deleting" but it kept this way indefinitely. I assumed if the wrk process would not quit by itself it could explain this. I will try to find the CI job and the trace that had me to think this way.

sebastienros: In this case I should probably also add the case when the job has been in Running state for more than the same duration. Which would be when the driver gets killed from the CI, and can't send the Delete message.

sebastienros: I added this case. Please review.

PR #355, opened 2017-11-22
Adding MongoDb scenarios to graphs

sebastienros:


src/Benchmarks/Data/ApplicationDbSeeder.cs
```diff
@@ -1,97 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
```
mikeharder: Why was this file removed?

sebastienros: All the docker images we have for the databases are using custom scripts which are already seeded. And because it was showing error messages I decided to remove it.

PR #357, opened 2017-12-08
Adding HardwareVersion column

sebastienros:
Mandatory option on BenchmarkServer.
Will add the column once the change is approved, and set the correct historical values.
Also update the onenote.


src/BenchmarksDriver/Program.cs
```diff
@@ -949,6 +959,8 @@ private static Task WriteJobsToSql(ServerJob serverJob, ClientJob clientJob, str
                     p.AddWithValue("@AspNetCoreVersion", aspnetCoreVersion);
                     p.AddWithValue("@Scenario", scenario.ToString());
                     p.AddWithValue("@Hardware", hardware.ToString());
+                    p.AddWithValue("@HardwareVersion",
```
mikeharder: Do you want to allow `null` HardwareVersion or not?  The column definition says `NOT NULL`, but this line is trying to insert `DBNull.Value`.

PR #360, opened 2017-12-18
Installing sdk/runtime defined by Universe

sebastienros:


sebastienros:
Fixes the current regressions due to the wrong shared runtime used.
Uses the official way to install specific SDKs and runtimes.
Loads the runtime and sdk versions specified in Coherence Universe repos, when targeting 2.1-*

src/BenchmarksServer/Startup.cs
```diff
@@ -578,15 +599,37 @@ private static void DockerCleanUp(string containerId, string imageName)
...
             else
             {
-                ProcessUtil.Run("/usr/bin/env", "bash build.sh /t:noop",
+                // Install latest stable 2.0 SDK version (and associated runtime)
+                ProcessUtil.Run("/usr/bin/env", $"bash dotnet-install.sh --channel 2.0",
+                    workingDirectory: buildToolsPath,
+                    environmentVariables: env);
+
+                // Install latest SDK version (and associated runtime)
+                ProcessUtil.Run("/usr/bin/env", $"bash dotnet-install.sh --version {sdkVersion}",
```
pakrym: Why do we need two sdks?

sebastienros: Benchmarks can be run on the latest released SDK (2.0.3) or the one Universe is currently using (2.1.x). But writing that I assume your question is more about now that we know which one we use, we can only download one ... That would make sense, and this is not something we could do before with KoreBuild.

pakrym: Benchmark run on runtimes, not on SDK's we might need to have multiple runtimes installed but one SDK should be sufficient.

PR #372, opened 2018-01-26
Batch

sebastienros:


src/BenchmarksDriver/Program.cs
```diff
@@ -1147,6 +1168,7 @@ private static Task WriteJobsToSql(ServerJob serverJob, ClientJob clientJob, str
                         [Excluded] [bit] DEFAULT 0,
                         [DateTime] [datetimeoffset](7) NOT NULL,
                         [Session] [nvarchar](max) NOT NULL,
+                        [Batch] [nvarchar](128) NOT NULL,
```
mikeharder: Should `NULL` be allowed for cases when `Batch` is not specified?

PR #373, opened 2018-01-26
Custom client support

BrennanConroy:
@sebastienros

### DIFF
--- a/src/Benchmarks.ClientJob/ClientJob.cs
+++ b/src/Benchmarks.ClientJob/ClientJob.cs
@@ -18,11 +18,10 @@ namespace Benchmarks.ClientJob
         public ClientJob(ClientJob clientJob)
         {
             Id = clientJob.Id;
-            Threads = clientJob.Threads;
             Connections = clientJob.Connections;
             Duration = clientJob.Duration;
-            ScriptName = clientJob.ScriptName;
-            PipelineDepth = clientJob.PipelineDepth;
+            ClientProperties = clientJob.ClientProperties;
+            ClientName = clientJob.ClientName;
             Headers = clientJob.Headers;
             ServerBenchmarkUri = clientJob.ServerBenchmarkUri;
             Query = clientJob.Query;
@@ -36,16 +35,14 @@ namespace Benchmarks.ClientJob
 
         public int Id { get; set; }
 
-        public int Threads { get; set; } = 32;
+        public string ClientName { get; set; }
 
         public int Connections { get; set; } = 256;
 
         public int Duration { get; set; } = 15;
         public int Warmup { get; set; } = 15;
 
-        public string ScriptName { get; set; }
-
-        public int PipelineDepth { get; set; }
+        public Dictionary<string, string> ClientProperties { get; set; } = new Dictionary<string, string>();
 
         public Dictionary<string, string> Headers { get; set; } = new Dictionary<string, string>();
 
@@ -70,7 +67,7 @@ namespace Benchmarks.ClientJob
         public string Method { get; set; } = "GET";
         public bool SkipStartupLatencies { get; set; }
 
-        // Latency of  first request
+        // Latency of first request
         public TimeSpan LatencyFirstRequest { get; set; }
 
         // Latency with a single connection
--- a/src/BenchmarksClient/Startup.cs
+++ b/src/BenchmarksClient/Startup.cs
@@ -3,29 +3,24 @@
 
 using System;
 using System.Diagnostics;
-using System.IO;
 using System.Linq;
 using System.Net.Http;
-using System.Reflection;
-using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Benchmarks.ClientJob;
+using BenchmarksClient;
+using BenchmarksClient.Workers;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.CommandLineUtils;
 using Microsoft.Extensions.DependencyInjection;
-using Newtonsoft.Json;
 using Repository;
 
 namespace BenchmarkClient
 {
     public class Startup
     {
-        private static HttpClient _httpClient;
-        private static HttpClientHandler _httpClientHandler;
-
         private const string _defaultUrl = "http://*:5002";
 
         private static readonly IRepository<ClientJob> _jobs = new InMemoryRepository<ClientJob>();
@@ -67,11 +62,6 @@ namespace BenchmarkClient
                 return Run(url).Result;
             });
 
-            // Configuring the http client to trust the self-signed certificate
-            _httpClientHandler = new HttpClientHandler();
-            _httpClientHandler.ServerCertificateCustomValidationCallback = HttpClientHandler.DangerousAcceptAnyServerCertificateValidator;
-            _httpClient = new HttpClient(_httpClientHandler);
-
             return app.Execute(args);
         }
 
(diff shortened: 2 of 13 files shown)


BrennanConroy:
@sebastienros could you take a look?
I'd like to iterate on it

src/Benchmarks.ClientJob/ClientJob.cs
```diff
@@ -18,11 +18,11 @@ public ClientJob()
         public ClientJob(ClientJob clientJob)
         {
             Id = clientJob.Id;
-            Threads = clientJob.Threads;
             Connections = clientJob.Connections;
             Duration = clientJob.Duration;
-            ScriptName = clientJob.ScriptName;
-            PipelineDepth = clientJob.PipelineDepth;
+            MaxDuration = clientJob.MaxDuration;
```
sebastienros: We don't need it anymore

src/Benchmarks.ClientJob/ClientJob.cs
```diff
@@ -35,16 +35,16 @@ public ClientJob(ClientJob clientJob)
...
-        public int Threads { get; set; } = 32;
+        public string ClientName { get; set; }
 
         public int Connections { get; set; } = 256;
 
         public int Duration { get; set; } = 15;
         public int Warmup { get; set; } = 15;
 
-        public string ScriptName { get; set; }
+        public int MaxDuration { get; set; } = 20;
```
sebastienros: Remove

src/BenchmarksClient/Startup.cs
```diff
@@ -103,58 +100,42 @@ public static int Main(string[] args)
...
                         try
                         {
-                            Debug.Assert(process == null);
-
-                            await MeasureFirstRequestLatencyAsync(job);
+                            if (WorkerFactory.TryCreate(job, _httpClient, out worker, out var error) == false)
+                            {
+                                Console.WriteLine(error);
+                                // Worker failed to start
+                                job.State = ClientState.Completed;
```
sebastienros: `job.State = ClientState.Deleting;`

src/BenchmarksClient/Startup.cs
```diff
@@ -103,58 +100,42 @@ public static int Main(string[] args)
...
-                            job.State = ClientState.Running;
-                            _jobs.Update(job);
+                                await worker.StartAsync();
 
-                            job.LastDriverCommunicationUtc = DateTime.UtcNow;
+                                Log($"Running job {worker.JobLogText}");
+                            }
 
-                            process = StartProcess(job);
+                            job.LastDriverCommunicationUtc = DateTime.Now;
```
sebastienros: `UtcNow`

src/BenchmarksClient/Startup.cs
```diff
@@ -167,8 +148,7 @@ public static int Main(string[] args)
                     {
                         var now = DateTime.UtcNow;
 
-                        // Clean the job in case the driver is not running
-                        if (now - job.LastDriverCommunicationUtc > TimeSpan.FromSeconds(30))
+                        if (now - job.LastDriverCommunicationUtc > TimeSpan.FromSeconds(job.MaxDuration))
```
sebastienros: Revert to 30, this actually has nothing to do with `MaxDuration`. Later I'll update it to a much smaller value as the driver is actually communicating every second.

src/BenchmarksClient/Startup.cs
```diff
@@ -177,21 +157,19 @@ public static int Main(string[] args)
...
 
                         try
                         {
-                            if (!process.HasExited)
-                            {
-                                process.Kill();
-                            }
+                            // Wait 5 seconds before killing worker
+                            var stopTask = worker.StopAsync();
+                            await Task.WhenAny(stopTask, Task.Delay((int)TimeSpan.FromSeconds(5).TotalMilliseconds));
```
sebastienros: We should really kill the worker. So `StopAsync` should itself be responsible for killing the process if it's not stopped after some delay. So no need to wait for anything else than `StopAsync`

src/BenchmarksClient/WorkerFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System.Net.Http;
+using Benchmarks.ClientJob;
+using BenchmarksClient.Workers;
+
+namespace BenchmarksClient
+{
+    static public class WorkerFactory
+    {
+        static public bool TryCreate(ClientJob clientJob, HttpClient httpClient, out IWorker worker, out string errors)
```
sebastienros: Do we really need the `HttpClient` ? This can be local to each worker.

src/BenchmarksServer/Startup.cs
```diff
@@ -851,7 +851,9 @@ private static async Task<(string benchmarkDir, string dotnetDir)> CloneRestoreA
 
                 var outputFolder = Path.Combine(benchmarkedApp, "published");
 
-                ProcessUtil.Run(dotnetExecutable, $"publish -c Release -o {outputFolder} {buildParameters}",
+                var rid = OperatingSystem == OperatingSystem.Windows ? "win-x64" : "linux-x64";
```
sebastienros: Revert this change, I reverted it in the meantime, sorry.

src/BenchmarksDriver/Program.cs
```diff
@@ -1193,7 +1211,7 @@ private static Task WriteJobsToSql(ServerJob serverJob, ClientJob clientJob, str
             string connectionFilter,
             WebHost webHost,
             int? kestrelThreadCount,
-            int clientThreads,
+            int? clientThreads,
```
sebastienros: Don't change that, just initialize to zero. We really want a value.

BrennanConroy:
🆙 📅

src/BenchmarksDriver/Program.cs
```diff
@@ -1325,7 +1337,7 @@ private static Task WriteJobsToSql(ServerJob serverJob, ClientJob clientJob, str
                         string.IsNullOrEmpty(connectionFilter) ? (object)DBNull.Value : connectionFilter);
                     p.AddWithValue("@WebHost", webHost.ToString());
                     p.AddWithValue("@KestrelThreadCount", (object)kestrelThreadCount ?? DBNull.Value);
-                    p.AddWithValue("@ClientThreads", clientThreads);
+                    p.AddWithValue("@ClientThreads", (object)clientThreads);
```
sebastienros: I assume that anything that is specific to workers can't be mapped to specific database columns.
But we should have a generic column that records the parameters in case we want to analyze it. Then we can just add an NVARCHAR(max) column that would contain a CSV with the worker properties. CSV would be easier to manipulate than Json in a db column.

src/BenchmarksDriver/Program.cs
```diff
@@ -508,16 +519,17 @@ public static int Main(string[] args)
                 }
                 if (pipelineDepthOption.HasValue())
                 {
-                    _clientJob.PipelineDepth = int.Parse(pipelineDepthOption.Value());
+                    var pipelineDepth = int.Parse(pipelineDepthOption.Value());
+                    _clientJob.WorkerProperties["PipelineDepth"] = pipelineDepth;
```
sebastienros: This could be handled in a generic way were we pass client specific arguments to the worker properties.
Something like `--client wrk --properties thread=16;connections=32;pipelinedepth=16` or similar

BrennanConroy: Do we want to get rid of the `--clientThreads` `--script` and `--pipelineDepth` options

BrennanConroy:
Rebased and :up: 📅

davidfowl:
@sebastienros can you review this please? We're trying to get performance runs for SignalR going ASAP.

davidfowl:
:clap:

PR #384, opened 2018-02-21
Adding support for RSP files

sebastienros:
Switching to Nate's library that supports @ parameters
Also supports urls in filenames

Example:
`@local-linux.rsp @kestrel-plaintext.rsp --duration 10`

### DIFF
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -13,7 +13,7 @@ using System.Threading.Tasks;
 using System.Web;
 using Benchmarks.ClientJob;
 using Benchmarks.ServerJob;
-using Microsoft.Extensions.CommandLineUtils;
+using McMaster.Extensions.CommandLineUtils;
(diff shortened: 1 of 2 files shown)


src/BenchmarksDriver/Program.cs
```diff
@@ -611,6 +612,27 @@ public static int Main(string[] args)
                 return Run(new Uri(server), new Uri(client), sqlConnectionString, serverJob, session, description, iterations, exclude, shutdownOption.Value(), span, downloadFilesOption.Values, collectR2RLogOption.HasValue()).Result;
             });
 
+            // Resolve reponse files from urls
+
+            for (var i = 0; i < args.Length; i++)
+            {
+                if (args[i].StartsWith("@http", StringComparison.OrdinalIgnoreCase))
```
pakrym: Interesting idea 👍

davidfowl:
Damn thats baller

Issue #375, opened 2018-02-06
Make the benchmark driver a global tool


davidfowl:
I shouldn't have to build this repository to run a benchmark.

sebastienros:
The driver should also verify that it's compatible with the servers before starting a job. And suggest to update if not. It means adding an API version on the servers.

PR #385, opened 2018-02-21
Dockerfile and scripts to run BenchmarksDriver on Windows

mikeharder:
CC: @davidfowl


sebastienros:
This also might be as useful https://github.com/aspnet/benchmarks/issues/375

davidfowl:
nice!

PR #388, opened 2018-02-22
Support for --trace-output

sebastienros:


src/BenchmarksDriver/Program.cs
```diff
@@ -883,9 +900,14 @@ public static int Main(string[] args)
                                 var counter = 1;
                                 while (File.Exists(filename))
                                 {
-                                    filename = $"trace({counter++}).etl.zip";
+                                    filename = $"trace ({counter++}).etl.zip";
                                 }
 
+                                if (!String.IsNullOrEmpty(traceDestination))
```
mikeharder: I think you need to prepend the `traceDestination` before the `File.Exists()` check.

sebastienros:
Updated

PR #389, opened 2018-02-22
Improved attachments

sebastienros:


sebastienros:
Note you can't test it until the servers are up to date as there is a new state for the server job.

src/BenchmarksDriver/Program.cs
```diff
@@ -1231,6 +1245,44 @@ public static int Main(string[] args)
...
+                    return 8;
+                }
+
+                var destinationFilename = outputFileSegments.Length > 1
+                    ? outputFileSegments[1]
+                    : Path.GetFileName(attachmentFilename);
+
+                var requestContent = new MultipartFormDataContent();
+
+                var fileContent = attachmentFilename.StartsWith("http", StringComparison.OrdinalIgnoreCase)
```
davidfowl: var fileContent = attachmentFilename.StartsWith("http", StringComparison.OrdinalIgnoreCase)
                    ? new StreamContent(await _httpClient.GetByteArrayAsync(attachmentFilename))
                    : new StreamContent(new FileStream(attachmentFilename,FileMode.Open,FileAccess.Read,FileShare.ReadWrite,1,FileOptions.Asynchronous | FileOptions.SequentialScan));

sebastienros: Is `1` for the buffer size intentional? default is 4K

benaadams: `0` throws, `1` says don't use internal buffer

benaadams: Came up in https://github.com/aspnet/KestrelHttpServer/pull/2049#discussion_r165830218

image

PR #392, opened 2018-02-28
Support for custom client

sebastienros:
Without any breaking change in the Server and Client protocols.
This way it can be merged without affecting the current running processes and configurations, then we can update the clients one by one, and finally remove support for the current clients.


src/BenchmarksClient/Startup.cs
```diff
@@ -102,64 +103,39 @@ public static int Main(string[] args)
...
                 var allJobs = _jobs.GetAll();
                 var job = allJobs.FirstOrDefault();
                 if (job != null)
                 {
-                    var jobLogText =
-                        $"[ID:{job.Id} Connections:{job.Connections} Threads:{job.Threads} Duration:{job.Duration} Method:{job.Method} ServerUrl:{job.ServerBenchmarkUri}";
-
-                    Debug.Assert(job.PipelineDepth <= 0 || job.ScriptName != null, "A script name must be present when the pipeline depth is larger than 0.");
-
-                    if (!string.IsNullOrEmpty(job.ScriptName))
```
BrennanConroy: These aren't needed for any scripts we're running still?

sebastienros: The currently deployed clients have this code. When I deploy the new code it will use the properties.

PR #396, opened 2018-03-01
Add option "--trace-arguments"

mikeharder:
- Increase default BufferSize from 256 to 1024, which should be sufficient for capturing a trace of a default Plaintext run
- Set default BufferSize in Driver rather than Server, so it can be changed without updating servers.

### DIFF
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -110,7 +110,10 @@ namespace BenchmarksDriver
                 "\"--runtimeFile c:\\build\\System.Net.Security.dll\"",
                 CommandOptionType.MultipleValue);
             var collectTraceOption = app.Option("--collect-trace",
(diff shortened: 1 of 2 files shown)


src/BenchmarksDriver/Program.cs
```diff
@@ -382,13 +385,8 @@ public static int Main(string[] args)
...
 
-                    serverJob.CollectArguments = collectTraceOption.Value();
-
-                    // Clear the arguments if the value is "on" as this is a marker for NoValue on the command parser
-                    if (serverJob.CollectArguments == "on")
-                    {
-                        serverJob.CollectArguments = "";
-                    }
+                    // Default to arguments which should be sufficient for collecting trace of default Plaintext run
+                    serverJob.CollectArguments = traceArgumentsOption.Value() ?? "BufferSize=1024";
```
sebastienros: I think it should be appended, this way even if you set other options you get the default BufferSize.

sebastienros: And put "BufferSize=1024" first so that `traceArgumentOption` can override the value.

PR #398, opened 2018-03-02
Add SignalR Client and scenario

BrennanConroy:
Lets go!

I'd prefer getting this in quickly so we can iterate on it.
It shouldn't cause any issues with our automation and I can spin up a couple test runs on our machines to see what kind of things need work.

### DIFF
--- a/src/Benchmarks/Middleware/SignalRMiddleware.cs
+++ b/src/Benchmarks/Middleware/SignalRMiddleware.cs
@@ -1,0 +1,55 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Benchmarks.Configuration;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.SignalR;
+
+namespace Benchmarks.Middleware
+{
+    public static class SignalRMiddlewareExtensions
+    {
+        public static IApplicationBuilder UseSignalRMiddleware(this IApplicationBuilder builder)
+        {
+            return builder.UseSignalR(route =>
+            {
+                route.MapHub<EchoHub>(Scenarios.GetPath(s => s.SignalRBroadcast) + "/default");
+            });
+        }
+    }
+
+    public class EchoHub : Hub
+    {
+        public override Task OnConnectedAsync()
+        {
+            return Task.CompletedTask;
+        }
+
+        public override Task OnDisconnectedAsync(Exception exception)
+        {
+            return Task.CompletedTask;
+        }
+
+        public async Task Echo(int duration)
+        {
+            try
+            {
+                var t = new CancellationTokenSource();
+                t.CancelAfter(TimeSpan.FromSeconds(duration));
+                while (!t.IsCancellationRequested)
+                {
+                    await Clients.All.SendAsync("echo", DateTime.UtcNow);
+                }
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine(e);
+            }
+
+            Console.WriteLine("Echo exited");
+        }
+    }
+}
--- a/src/Benchmarks/Middleware/SignalRMiddleware.cs
+++ b/src/Benchmarks/Middleware/SignalRMiddleware.cs
@@ -23,23 +23,13 @@ namespace Benchmarks.Middleware
 
     public class EchoHub : Hub
     {
-        public override Task OnConnectedAsync()
-        {
-            return Task.CompletedTask;
-        }
-
-        public override Task OnDisconnectedAsync(Exception exception)
-        {
-            return Task.CompletedTask;
-        }
-
         public async Task Echo(int duration)
         {
             try
             {
                 var t = new CancellationTokenSource();
                 t.CancelAfter(TimeSpan.FromSeconds(duration));
-                while (!t.IsCancellationRequested)
+                while (!t.IsCancellationRequested && !Context.Connection.ConnectionAbortedToken.IsCancellationRequested)
                 {
                     await Clients.All.SendAsync("echo", DateTime.UtcNow);
                 }
--- a/src/Benchmarks/Startup.cs
+++ b/src/Benchmarks/Startup.cs
@@ -154,6 +154,12 @@ namespace Benchmarks
             {
                 services.AddResponseCaching();
             }
+
+            if (Scenarios.Any("SignalR"))
+            {
+                services.AddSignalR()
+                    .AddMessagePackProtocol();
+            }
         }
 
         public void Configure(IApplicationBuilder app, IHostingEnvironment env)
@@ -305,6 +311,11 @@ namespace Benchmarks
             if (Scenarios.ResponseCachingPlaintextVaryByCached)
             {
                 app.UseResponseCachingPlaintextVaryByCached();
+            }
+
+            if (Scenarios.SignalRBroadcast)
+            {
+                app.UseSignalRMiddleware();
             }
 
             app.RunDebugInfoPage();
--- a/src/Benchmarks/signalr.json
+++ b/src/Benchmarks/signalr.json
@@ -1,0 +1,13 @@
+{
+  "Default": {
+    "ClientName": "signalr",
+    "Source": {
+      "Repository": "https://github.com/aspnet/benchmarks.git",
+      "BranchOrCommit": "brecon/signalrClientNew",
+      "Project": "src/Benchmarks/Benchmarks.csproj"
+    }
+  },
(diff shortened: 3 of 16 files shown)


src/Benchmarks/signalr.json
```diff
@@ -0,0 +1,13 @@
+{
+  "Default": {
+    "ClientName": "signalr",
+    "Source": {
+      "Repository": "https://github.com/aspnet/benchmarks.git",
+      "BranchOrCommit": "brecon/signalrClientNew",
```
BrennanConroy: Change this before merging

src/BenchmarksWorkers/Workers/SignalRWorker.cs
```diff
@@ -0,0 +1,292 @@
...
+                    max = _requestsPerConnection[i];
+                }
+                if (_requestsPerConnection[i] < min)
+                {
+                    min = _requestsPerConnection[i];
+                }
+            }
+            // Review: This could be interesting information, see the gap between most active and least active connection
+            // Ideally they should be within a couple percent of each other, but if they aren't something could be wrong
+            Log($"Least Requests to Connection: {min}");
```
anurse: ~~to~~ **per**

src/Benchmarks/Middleware/SignalRMiddleware.cs
```diff
@@ -0,0 +1,55 @@
...
+            return builder.UseSignalR(route =>
+            {
+                route.MapHub<EchoHub>(Scenarios.GetPath(s => s.SignalRBroadcast) + "/default");
+            });
+        }
+    }
+
+    public class EchoHub : Hub
+    {
+        public override Task OnConnectedAsync()
```
davidfowl: Remove these overrides?

src/Benchmarks/Middleware/SignalRMiddleware.cs
```diff
@@ -0,0 +1,55 @@
...
+            return Task.CompletedTask;
+        }
+
+        public async Task Echo(int duration)
+        {
+            try
+            {
+                var t = new CancellationTokenSource();
+                t.CancelAfter(TimeSpan.FromSeconds(duration));
+                while (!t.IsCancellationRequested)
```
davidfowl: Wait on client disconnected as well?

BrennanConroy: Ok, I'll keep track of how many connections are on the server and when it hits 0 exit the loop

davidfowl: You don't have to. You can just make sure you're not sending after this client is gone by checking this token and the aborted token.

src/BenchmarksWorkers/Workers/SignalRSerializer.cs
```diff
@@ -0,0 +1,289 @@
...
+            {
+                await WriteJobResultToSqlAsync(serverJob, clientJob, connectionString, tableName, path, session, description, statistics, longRunning, "RequestsPerSecond", statistics.RequestsPerSecond);
+            });
+
+            await RetryOnExceptionAsync(5, async () =>
+            {
+                await WriteJobResultToSqlAsync(serverJob, clientJob, connectionString, tableName, path, session, description, statistics, longRunning, "CPU", statistics.Cpu);
+            });
+
+            await RetryOnExceptionAsync(5, async () =>
```
davidfowl: I see working set here twice? Copy paste error?

src/BenchmarksWorkers/Workers/SignalRWorker.cs
```diff
@@ -0,0 +1,292 @@
...
+        private void CreateConnections(TransportType transportType = TransportType.WebSockets)
+        {
+            _connections = new List<HubConnection>(_job.Connections);
+            _requestsPerConnection = new List<int>(_job.Connections);
+            _latencyPerConnection = new List<List<double>>(_job.Connections);
+
+            var hubConnectionBuilder = new HubConnectionBuilder()
+                .WithUrl(_job.ServerBenchmarkUri)
+                .WithMessageHandler(_httpClientHandler)
+                //.WithConsoleLogger(Microsoft.Extensions.Logging.LogLevel.Trace) TODO: Support logging on client for debugging purposes
```
davidfowl: File an issue for this

src/BenchmarksWorkers/Workers/SignalRWorker.cs
```diff
@@ -0,0 +1,292 @@
...
+            {
+                _job.State = ClientState.Completed;
+            }
+        }
+
+        public async Task StopAsync()
+        {
+            if (_timer != null)
+            {
+                _timer?.Dispose();
```
davidfowl: We know the timer isn't null here. Do we need a lock?

src/BenchmarksWorkers/Workers/SignalRWorker.cs
```diff
@@ -0,0 +1,292 @@
...
+
+            await Task.WhenAll(tasks);
+
+            _job.State = ClientState.Running;
+            _job.LastDriverCommunicationUtc = DateTime.UtcNow;
+
+            // SendAsync will return as soon as the request has been sent (non-blocking)
+            await _connections[0].SendAsync("Echo", _job.Duration + 1);
+            _workTimer.Start();
+            _timer = new Timer(StopClients, null, TimeSpan.FromSeconds(_job.Duration), Timeout.InfiniteTimeSpan);
```
davidfowl: Wait, we have a timer that runs once?

src/BenchmarksWorkers/Workers/SignalRSerializer.cs
BrennanConroy: Yes

src/BenchmarksWorkers/Workers/SignalRWorker.cs
BrennanConroy: We should lock

src/BenchmarksWorkers/Workers/SignalRWorker.cs
BrennanConroy: A new worker is created for every job, and the timer is to stop the worker when the job is done

src/BenchmarksWorkers/Workers/SignalRSerializer.cs
```diff
@@ -0,0 +1,289 @@
...
+                        [RuntimeVersion] [nvarchar](50) NOT NULL,
+                        [Scenario] [nvarchar](50) NOT NULL,
+                        [Hardware] [nvarchar](50) NOT NULL,
+                        [HardwareVersion] [nvarchar](128) NOT NULL,
+                        [OperatingSystem] [nvarchar](50) NOT NULL,
+                        [Framework] [nvarchar](50) NOT NULL,
+                        [RuntimeStore] [bit] NOT NULL,
+                        [Scheme] [nvarchar](50) NOT NULL,
+                        [Sources] [nvarchar](50) NULL,
+                        [ConnectionFilter] [nvarchar](50) NULL,
```
sebastienros: Do you need that?

src/BenchmarksWorkers/Workers/SignalRSerializer.cs
```diff
@@ -0,0 +1,289 @@
...
+                        [WebHost] [nvarchar](50) NOT NULL,
+                        [Transport] [nvarchar](50) NOT NULL,
+                        [HubProtocol] [nvarchar](50) NOT NULL,
+                        [Connections] [int] NOT NULL,
+                        [Duration] [int] NOT NULL,
+                        [Path] [nvarchar](200) NULL,
+                        [Method] [nvarchar](50) NOT NULL,
+                        [Headers] [nvarchar](max) NULL,
+                        [Dimension] [nvarchar](50) NOT NULL,
```
sebastienros: I would suggest to go for a static set of columns instead, unless you feel confident with PBI.

src/Benchmarks/Startup.cs
```diff
@@ -154,6 +154,12 @@ public void ConfigureServices(IServiceCollection services)
             {
                 services.AddResponseCaching();
             }
+
+            if (Scenarios.Any("SignalR"))
```
sebastienros: I am fine having them in Benchmarks, but you might prefer a simpler custom application to have more flexibility. Even in this repository.

src/BenchmarksWorkers/Workers/SignalRSerializer.cs
BrennanConroy: Removed a couple items, ConnectionFilter and Method

src/BenchmarksWorkers/Workers/SignalRSerializer.cs
BrennanConroy: Seems fairly straightforward, we can talk offline

PR #409, opened 2018-03-13
Added platform level plaintext benchmark

davidfowl:
- Added very basic plain text benchmark
- Machinery is hidden behind an HttpConnection base class.
It encapsulates the Kestrel HttpParser and control flow
and has exposes virtual methods for handling request headers.
- Writing is manually writing headers to the output buffer.
We can no longer use the kestrel headers directly because of
some API changes...
- We may need to add a bit more API to make sure this doesn't
count as a stripped implementation.

/cc @benaadams @KrzysztofCwalina

### DIFF
--- a/src/PlatformBenchmarks/AsciiString.cs
+++ b/src/PlatformBenchmarks/AsciiString.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
 using System.Text;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 
--- a/src/PlatformBenchmarks/AsciiString.cs
+++ b/src/PlatformBenchmarks/AsciiString.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Text;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
+
+namespace PlatformBenchmarks
+{
+    public struct AsciiString : IEquatable<AsciiString>
+    {
+        private readonly byte[] _data;
+
+        public AsciiString(string s) => _data = Encoding.ASCII.GetBytes(s);
+
+        public int Length => _data.Length;
+
+        public ReadOnlySpan<byte> AsSpan() => _data;
+
+        public static implicit operator ReadOnlySpan<byte>(AsciiString str) => str._data;
+        public static implicit operator byte[] (AsciiString str) => str._data;
+
+        public static implicit operator AsciiString(string str) => new AsciiString(str);
+
+        public override string ToString() => HttpUtilities.GetAsciiStringNonNullCharacters(_data);
+        public static explicit operator string(AsciiString str) => str.ToString();
+
+        public bool Equals(AsciiString other) => ReferenceEquals(_data, other._data) || SequenceEqual(_data, other._data);
+        private bool SequenceEqual(byte[] data1, byte[] data2) => new Span<byte>(data1).SequenceEqual(data2);
+
+        public static bool operator ==(AsciiString a, AsciiString b) => a.Equals(b);
+        public static bool operator !=(AsciiString a, AsciiString b) => !a.Equals(b);
+        public override bool Equals(object other) => (other is AsciiString) && Equals((AsciiString)other);
+
+        public override int GetHashCode()
+        {
+            // Copied from x64 version of string.GetLegacyNonRandomizedHashCode()
+            // https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/String.Comparison.cs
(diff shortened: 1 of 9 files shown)


src/PlatformBenchmarks/PlatformBenchmarks.csproj
```diff
@@ -0,0 +1,33 @@
...
+    <DefineConstants>$(DefineConstants);DOTNET210</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="$(BenchmarksAspNetCoreVersion.StartsWith('2.0'))">
+    <DefineConstants>$(DefineConstants);DOTNET200</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Jil" Version="$(JilVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Diagnostics" Version="$(BenchmarksAspNetCoreVersion)" />
```
davidfowl: I can remove most of these

src/PlatformBenchmarks/BenchmarkApplication.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            // Server headers
+            writer.Write(_headerServer);
+
+            // Date header
+            writer.Write(_dateHeaderValueManager.GetDateHeaderValues().Bytes);
+            writer.Write(_crlf);
+
+            // Content-Length 0
+            writer.Write(_headerContentLengthZero);
```
KrzysztofCwalina: This constant could have CRLF at the end so you dont have to write twice.

src/PlatformBenchmarks/BenchmarkApplication.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            // Server headers
+            writer.Write(_headerServer);
+
+            // Date header
+            writer.Write(_dateHeaderValueManager.GetDateHeaderValues().Bytes);
+            writer.Write(_crlf);
+
+            // Content-Type header
+            writer.Write(_headerContentTypeText);
```
KrzysztofCwalina: same

KrzysztofCwalina:
Nice! Merge and get the results :-)

sebastienros:
@KrzysztofCwalina we don't need to merge to get the results 😎

src/PlatformBenchmarks/benchmarks.json
```diff
@@ -0,0 +1,30 @@
...
+      "ScriptName": "pipeline",
+      "PipelineDepth": 16
+    },
+    "ScriptName": "pipeline",
+    "PipelineDepth": 16,
+    "PresetHeaders": "Plaintext",
+    "Source": {
+      "Repository": "https://github.com/aspnet/benchmarks.git",
+      "BranchOrCommit": "dev",
+      "Project": "src/Benchmarks/PlatformBenchmarks.csproj"
```
sebastienros: `src\PlatformBenchmarks\PlatformBenchmarks.csproj`

I am fixing it

src/PlatformBenchmarks/Program.cs
```diff
@@ -0,0 +1,26 @@
...
+        public static void Main(string[] args)
+        {
+            BuildWebHost(args).Run();
+        }
+
+        public static IWebHost BuildWebHost(string[] args) =>
+            new WebHostBuilder()
+                .UseKestrel(options =>
+                {
+                    options.Listen(IPAddress.Loopback, 8080, builder =>
```
sebastienros: It needs to use the address passed in `--server.urls` so I'd suggest to use `UseConfiguration(config)` and get the args and ENV added. I will do a PR to this PR to test it works.

src/PlatformBenchmarks/BenchmarkApplication.cs
davidfowl: Nice catch!

src/PlatformBenchmarks/Program.cs
davidfowl: Lame 😄

sebastienros:
Latest CI results from TE are here: https://www.techempower.com/benchmarks/#section=test&runid=1e249f57-895c-459a-aecb-2b2b6981a56c

This PR reaches 6,148,558 so 3rd position being two rapidoid entries.
FYI the aspnetcore in this list will be 2.6 M in their next CI run.
We use the same hardware so numbers are directly comparable.

davidfowl:
@sebastienros updated

src/PlatformBenchmarks/AsciiString.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
```
JamesNK: Missing file headers for all cs files

sebastienros:
6,219,555 RPS

KrzysztofCwalina:
> 6,219,555 RPS

How does it compare to the official TechEmpower ASP.NET entry?

KrzysztofCwalina:
Very nice results!

PR #412, opened 2018-03-13
Platform level: Reduce interface dispatch and trys in main loop

benaadams:
/cc @davidfowl @KrzysztofCwalina @sebastienros

### DIFF
--- a/src/PlatformBenchmarks/HttpApplication.cs
+++ b/src/PlatformBenchmarks/HttpApplication.cs
@@ -100,7 +100,7 @@ namespace PlatformBenchmarks
 
                 if (!buffer.IsEmpty)
                 {
-                    ParseHttpRequest(in buffer, out consumed, out examined);
+                    ParseHttpRequest(buffer, out consumed, out examined);
 
                     if (_state != State.Body && result.IsCompleted)
                     {
@@ -128,8 +128,8 @@ namespace PlatformBenchmarks
             consumed = buffer.Start;
             examined = buffer.End;
 
-            var parsedStartLine = _state == State.StartLine;
-            if (parsedStartLine)
+            var parsingStartLine = _state == State.StartLine;
+            if (parsingStartLine)
             {
                 if (Parser.ParseRequestLine(this, buffer, out consumed, out examined))
                 {
@@ -139,7 +139,7 @@ namespace PlatformBenchmarks
 
             if (_state == State.Headers)
             {
-                if (Parser.ParseHeaders(this, parsedStartLine ? buffer.Slice(consumed) : buffer, out consumed, out examined, out int consumedBytes))
+                if (Parser.ParseHeaders(this, parsingStartLine ? buffer.Slice(consumed) : buffer, out consumed, out examined, out int consumedBytes))
                 {
                     _state = State.Body;
                 }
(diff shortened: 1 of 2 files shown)


sebastienros:
6,460,818 RPS

benaadams:
:)

benaadams:
Added PR to change parsing to pass ReadOnlySequence via `in` https://github.com/aspnet/KestrelHttpServer/pull/2389

src/PlatformBenchmarks/HttpApplication.cs
```diff
@@ -64,85 +67,90 @@ public virtual ValueTask OnReadCompletedAsync()
...
+                var consumed = buffer.Start;
+                var examined = buffer.End;
 
-                    var result = await task;
-                    var buffer = result.Buffer;
-                    var consumed = buffer.Start;
-                    var examined = buffer.End;
+                if (!buffer.IsEmpty)
+                {
+                    ParseHttpRequest(in buffer, out consumed, out examined);
```
davidfowl: Why are you passing in here explicitly?

src/PlatformBenchmarks/HttpApplication.cs
```diff
@@ -64,85 +67,90 @@ public virtual ValueTask OnReadCompletedAsync()
...
         }
 
-        private void ParseHttpRequest(ReadOnlySequence<byte> buffer, out SequencePosition consumed, out SequencePosition examined)
+        private void ParseHttpRequest(in ReadOnlySequence<byte> buffer, out SequencePosition consumed, out SequencePosition examined)
         {
             consumed = buffer.Start;
             examined = buffer.End;
 
-            if (_state == State.StartLine)
+            var parsedStartLine = _state == State.StartLine;
```
davidfowl: parsingStartLine?

src/PlatformBenchmarks/HttpApplication.cs
davidfowl: Why are you specifying in.

benaadams: No reason; will change :)

src/PlatformBenchmarks/HttpApplication.cs
benaadams: Changed

src/PlatformBenchmarks/HttpApplication.cs
```diff
@@ -64,85 +67,90 @@ public virtual ValueTask OnReadCompletedAsync()
...
+        private void ParseHttpRequest(in ReadOnlySequence<byte> buffer, out SequencePosition consumed, out SequencePosition examined)
         {
             consumed = buffer.Start;
             examined = buffer.End;
 
-            if (_state == State.StartLine)
+            var parsingStartLine = _state == State.StartLine;
+            if (parsingStartLine)
             {
                 if (Parser.ParseRequestLine(this, buffer, out consumed, out examined))
```
KrzysztofCwalina: Would it make sense to allow passing buffer by-in?

davidfowl: Yes, we just changed this in Kestrel.

ErikEJ:
Now go fix SqlClient 😁

sebastienros:
@ErikEJ yeah, and we are actively working on making it faster too, but FYI SqlClient is already faster than the implementations for NodeJs and Java that we tested, measured on Windows (SNI) and Linux (managed). The SqlClient team has already made great improvements that will be visible with the next release (faster on Linux, faster azure encrypted connections, reliability, ...), and it will also benefit 2.0 apps.

PR #414, opened 2018-03-14
Add an explicit reference to Kestrel.Libuv

davidfowl:


halter73:
It looks like that's everything other than the docs.

https://github.com/search?p=1&q=org%3Aaspnet+UseLibuv&type=Code

davidfowl:
Nice!

PR #416, opened 2018-03-14
Eliminated some calls to Write in PlatformBenchmark

KrzysztofCwalina:
DateHeader value now includes both leading and trailing CRLF
Also added constant for EOH (end of headers)

@davidfowl, is it possible to get numbers from this branch before we merge?

Also, I assume the heartbeat interval is large enough such that the lock does not matter.

### DIFF
--- a/src/PlatformBenchmarks/DateHeader.cs
+++ b/src/PlatformBenchmarks/DateHeader.cs
@@ -5,6 +5,7 @@ using System;
 using System.Buffers.Text;
 using System.Diagnostics;
 using System.Text;
+using System.Threading;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
@@ -12,20 +13,21 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
     /// <summary>
     /// Manages the generation of the date header value.
     /// </summary>
-    internal class DateHeader : IHeartbeatHandler
+    internal static class DateHeader 
     {
         const int prefixLength = 8; // "\r\nDate: ".Length
         const int dateTimeRLength = 29; // Wed, 14 Mar 2018 14:20:00 GMT
         const int suffixLength = 2; // crlf
         const int suffixIndex = dateTimeRLength + prefixLength;
 
-        private byte[] s_headerBytesMaster = new byte[prefixLength + dateTimeRLength + suffixLength];
-        private byte[] s_headerBytesScratch = new byte[prefixLength + dateTimeRLength + suffixLength];
+        private static readonly Timer s_timer = new Timer((s) => {
+            SetDateValues(DateTimeOffset.UtcNow);
+        }, null, 1000, 1000);
 
-        /// <summary>
-        /// Initializes a new instance of the <see cref="DateHeaderValueManager"/> class.
-        /// </summary>
-        public DateHeader()
+        private static byte[] s_headerBytesMaster = new byte[prefixLength + dateTimeRLength + suffixLength];
+        private static byte[] s_headerBytesScratch = new byte[prefixLength + dateTimeRLength + suffixLength];
+
+        static DateHeader()
         {
             var utf8 = Encoding.ASCII.GetBytes("\r\nDate: ").AsSpan();
             utf8.CopyTo(s_headerBytesMaster);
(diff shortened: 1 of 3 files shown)


src/PlatformBenchmarks/DateHeader.cs
```diff
@@ -0,0 +1,60 @@
...
+            utf8.CopyTo(s_headerBytesScratch);
+            s_headerBytesMaster[suffixIndex] = (byte)'\r';
+            s_headerBytesMaster[suffixIndex + 1] = (byte)'\n';
+            s_headerBytesScratch[suffixIndex] = (byte)'\r';
+            s_headerBytesScratch[suffixIndex + 1] = (byte)'\n';
+            SetDateValues(DateTimeOffset.UtcNow);
+        }
+
+        // Called by the Timer (background) thread
+        public void OnHeartbeat(DateTimeOffset now) => SetDateValues(now);
```
davidfowl: This needs to be wired up.

KrzysztofCwalina: how do I do that?

benaadams: Without getting too fancy, Timer?
```csharp
new Timer((s) => ((DateHeader)s).OnHeartbeat(DateTimeOffset.UtcNow), this, 1000, 1000);
```

KrzysztofCwalina: Yeah, but I was hoping to be able to hookup to Kestrel's heartbeat times.

sebastienros:
@KrzysztofCwalina 

| Before | After |
| -- | -- |
| 5,935,146 | 6,063,694 |

If you wonder why the base is lower than yesterday, you are not alone, looking into it...

src/PlatformBenchmarks/DateHeader.cs
benaadams: /cc @pakrym @halter73

benaadams: > Yeah, but I was hoping to be able to hookup to Kestrel's heartbeat times.

/cc @pakrym @halter73

halter73: https://github.com/aspnet/KestrelHttpServer/blob/9901f0f3e4f9bf7053fa7f4b072cb72d67bedb43/src/Kestrel.Core/KestrelServer.cs#L50

halter73: I realized that probably won't help you much in the Benchmarks repo.

There are a few options. The easiest is probably to add a constructor to KestrelServer that takes in `IEnumerable<IHeartbeatHandler>` and constructs the Heartbeat object with any additional IHeartbeatHandlers registered as services.

KrzysztofCwalina: I assumed that I would be able to just add my implementation to your DI system. Why not?

davidfowl: @halter73 we should change the heartbeat to allow adding things after the fact.

davidfowl: @KrzysztofCwalina you can't today, let me look into this.

halter73: It's a super trivial change to pull IHeartbeatHandlers from DI and that's exactly what I suggested would be the easiest to make work.

Just because it's easy doesn't mean we necessarily should do this though. I expect our API check will complain about taking a type from an pubternal namespace in a public ctor.

Not only are there the docs and other stuff involved in adding public API surface to worry about, we *really* don't want to risk badly-written heartbeat handlers causing heartbeats to be skipped and preventing timeouts from firing. We log when a heartbeat gets skipped, but the cause of a missed beat isn't always obvious and happens sometimes today without any extra handlers (or attached debuggers for that matter).

When the alternative is just using a Timer yourself like Heartbeat itself does, the performance of a second Timer would have to be pretty bad to make me think it's a good idea to support user defined heartbeat handlers in our product.

davidfowl:
@sebastienros can you measure again?

sebastienros:
| Before | After |
| -- | -- |
| 5,988,537 | 6,079,617 |

+1.5%

KrzysztofCwalina:
Should I merge then?

davidfowl:
@KrzysztofCwalina Yep

src/PlatformBenchmarks/DateHeader.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Text;
+using System.Threading;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
+{
+    /// <summary>
+    /// Manages the generation of the date header value.
+    /// </summary>
+    internal static class DateHeader
```
davidfowl: @halter73 Maybe we should change the code in kestrel to do this and measure?

PR #419, opened 2018-03-19
Add some connection string checks for Npgsql

roji:
To make sure we're in high-perf mode.

/cc @sebastienros 

Is it possible to also add something similar to https://github.com/aspnet/FrameworkBenchmarks to prevent accidental runs with non-optimized connection strings?

### DIFF
--- a/src/Benchmarks/Startup.cs
+++ b/src/Benchmarks/Startup.cs
@@ -59,6 +59,12 @@ namespace Benchmarks
             switch (appSettings.Database)
             {
                 case DatabaseServer.PostgreSql:
+                     var settings = new NpgsqlConnectionStringBuilder(appSettings.ConnectionString);
+                     if (!settings.NoResetOnClose)
+                         throw new ArgumentException("No Reset On Close=true must be specified for Npgsql");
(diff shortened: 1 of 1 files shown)


roji:
Ping, in the default mode Npgsql will attempt to enlist to an ambient TransactionScope, which will reduce Open() perf, so it's important to make sure this doesn't happen in the benchmarks...

sebastienros:
I already started to apply the connection string changes first, otherwise it would blow up existing scenarios.
I will soon be able to merge it (as I restart the servers I apply it).

However I thought that the value was false by default, I looked at the documentation.

roji:
It's true that it was off by default, but I've changed this for 4.0.0 - see the discussion in https://github.com/npgsql/npgsql/issues/1483. I've updated the performance page in the docs to make sure people know to disable enlist in "high perf mode". However the docs on npgsql.org are still for 3.2, this is why you're still seeing that it's false by default.

That's why I'm submitting this PR :)

sebastienros:
Correct, I can merge this PR. Just waiting to restart the long running jobs. Should be done by tomorrow morning.

Issue #381, opened 2018-02-17
Use optimized content type for MVC actions


sebastienros:
TE warning in the build log

```
     Content encoding found in "application/json; charset=utf-8" where "application/json" is acceptable.
     Additional response bytes may negatively affect benchmark performance.
     See http://frameworkbenchmarks.readthedocs.org/en/latest/Project-Information/Framework-Tests/#specific-test-requirements
```

All these MVC actions use `[Produces("application/json")]`. Maybe we can use a better way to set the content type to improve it.

https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/342443158

sebastienros:
@davidfowl any idea? Otherwise I will investigate

benaadams:
`System.Net.Http.StringContent` always setting encoding whether you want it or not?
 https://github.com/dotnet/corefx/issues/25290

[System/Net/Http/StringContent.cs#L26-L34](https://github.com/dotnet/corefx/blob/cac1f05d705072c4bb20045f84ef3370d87a976a/src/System.Net.Http/src/System/Net/Http/StringContent.cs#L26-L34)
```csharp
headerValue.CharSet = (encoding == null) ? HttpContent.DefaultStringEncoding.WebName : encoding.WebName;

Headers.ContentType = headerValue;
```

DamianEdwards:
Any idea if this is still something we should look into?

sebastienros:
I don't think it matters, it's for MVC which is lower in the results, and this would only impact networking which wouldn't improve the results.

PR #420, opened 2018-03-19
Add IIS Support to benchmark runner

pakrym:


src/BenchmarksDriver/Program.cs
```diff
@@ -1188,7 +1184,8 @@ public static int Main(string[] args)
         private static async Task<ClientJob> RunClientJob(string scenarioName, Uri clientUri, Uri serverJobUri, string serverBenchmarkUri)
         {
             var clientJob = new ClientJob(_clientJob) { ServerBenchmarkUri = serverBenchmarkUri };
-
+            var benchmarkUri = new Uri(serverBenchmarkUri);
+            clientJob.Headers["Host"] = benchmarkUri.Host + ":" + benchmarkUri.Port;
```
pakrym: IIS validates host header

src/BenchmarksServer/Startup.cs
```diff
@@ -1336,21 +1342,39 @@ private static Process StartProcess(string hostname, string benchmarksRepo, Serv
...
             }
 
-            Log.WriteLine($"Starting process '{dotnetFilename} {arguments}'");
+            if (iis)
+            {
+                Log.WriteLine($"Generating application host config for '{executable} {arguments}'");
+
+                var apphost = GenerateApplicationHostConfig(job, benchmarksBin, executable, arguments, hostname);
+                arguments = $"-h \"{apphost}\"";
+                executable = @"C:\Windows\System32\inetsrv\w3wp.exe";
```
davidfowl: Can you use `Environment.GetFolderPath(SpecialFolder.Windows)`

src/BenchmarksServer/Startup.cs
```diff
@@ -1391,54 +1415,107 @@ private static Process StartProcess(string hostname, string benchmarksRepo, Serv
...
+                    throw new InvalidOperationException("Element not found");
+                }
+
+                element.SetAttributeValue(name, value);
+            }
+
+            using (var resourceStream = Assembly.GetCallingAssembly().GetManifestResourceStream("BenchmarksServer.applicationHost.config"))
+            {
+                var applicationHostConfig = XDocument.Load(resourceStream);
+                SetAttribute(applicationHostConfig, "/configuration/system.webServer/aspNetCore", "processPath", executable);
```
davidfowl: Wow XPath, what year is it.

src/BenchmarksServer/Startup.cs
```diff
@@ -1309,22 +1310,27 @@ private static string GetDotNetExecutable(string dotnetHome)
...
+            var benchmarksBin = job.UseRuntimeStore ? $"bin/Release/netcoreapp2.0" : $"published";
+            var benchmarksDll = Path.Combine(benchmarksBin, $"{projectFilename}.dll");
+            var iis = job.WebHost == WebHost.IISInProcess || job.WebHost == WebHost.IISOutOfProcess;
 
             var arguments = $"{benchmarksDll}" +
                     $" {job.Arguments} " +
                     $" --nonInteractive true" +
-                    $" --scenarios {job.Scenario}" +
-                    $" --server.urls {serverUrl}";
+
```
davidfowl: nit: remove whitespace?

src/BenchmarksServer/BenchmarksServer.csproj
```diff
@@ -6,6 +6,14 @@
     <OutputType>Exe</OutputType>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Content Remove="applicationHost.config" />
```
davidfowl: Do we need this?

src/BenchmarksServer/Startup.cs
```diff
@@ -1309,22 +1310,27 @@ private static string GetDotNetExecutable(string dotnetHome)
...
         private static Process StartProcess(string hostname, string benchmarksRepo, ServerJob job, string dotnetHome, bool perfview, StringBuilder standardOutput)
         {
+            job.BasePath = Path.Combine(benchmarksRepo, Path.GetDirectoryName(job.Source.Project));
+
             var serverUrl = $"{job.Scheme.ToString().ToLowerInvariant()}://{hostname}:{job.Port}";
-            var dotnetFilename = GetDotNetExecutable(dotnetHome);
+            var executable = GetDotNetExecutable(dotnetHome);
             var projectFilename = Path.GetFileNameWithoutExtension(job.Source.Project);
-            var benchmarksDll = job.UseRuntimeStore ? $"bin/Release/netcoreapp2.0/{projectFilename}.dll" : $"published/{projectFilename}.dll";
+            var benchmarksBin = job.UseRuntimeStore ? $"bin/Release/netcoreapp2.0" : $"published";
```
davidfowl: Why is this netcoreapp2.0?

src/BenchmarksServer/Startup.cs
```diff
@@ -1309,22 +1310,27 @@ private static string GetDotNetExecutable(string dotnetHome)
...
-                    $" --server.urls {serverUrl}";
+
+                    $" --scenarios {job.Scenario}";
 
             if (!string.IsNullOrEmpty(job.ConnectionFilter))
             {
                 arguments += $" --connectionFilter {job.ConnectionFilter}";
             }
 
+
```
davidfowl: nit: remove extra whitespace

src/BenchmarksServer/applicationHost.config
```diff
@@ -0,0 +1,733 @@
...
+
+    <fastCgi />
+
+    <!--
+
+      The <globalModules> section defines all native-code modules.
+      To enable a module, specify it in the <modules> section.
+
+    -->
+    <globalModules>
```
shirhatti: It's worth considering whether we should remove all modules besides ANCM (and I guess Anonymous auth which is required)?

src/BenchmarksServer/Startup.cs
```diff
@@ -1309,22 +1310,27 @@ private static string GetDotNetExecutable(string dotnetHome)
 
         private static Process StartProcess(string hostname, string benchmarksRepo, ServerJob job, string dotnetHome, bool perfview, StringBuilder standardOutput)
         {
+            job.BasePath = Path.Combine(benchmarksRepo, Path.GetDirectoryName(job.Source.Project));
+
             var serverUrl = $"{job.Scheme.ToString().ToLowerInvariant()}://{hostname}:{job.Port}";
-            var dotnetFilename = GetDotNetExecutable(dotnetHome);
+            var executable = GetDotNetExecutable(dotnetHome);
```
jkotalik: GetDotNetExecutable() => GetExecutable() and add logic to see if we are iis in that method.

src/BenchmarksServer/Startup.cs
```diff
@@ -1391,54 +1415,107 @@ private static Process StartProcess(string hostname, string benchmarksRepo, Serv
...
-                        {
-                            job.PerfViewTraceFile = Path.Combine(job.BasePath, "benchmarks.etl");
-                            var perfViewArguments = new Dictionary<string, string>();
-                            perfViewArguments["AcceptEula"] = "";
-                            perfViewArguments["NoGui"] = "";
-                            perfViewArguments["Process"] = process.Id.ToString();
+            if (iis)
+            {
+                // IIS doesn't print anything
+                Thread.Sleep(1000);
```
jkotalik: I would like to figure out a better way to detect if IIS is up or not. This is before first request, correct? Were you running into issues sending a request at this point or were there issues with perfView?

src/BenchmarksServer/Startup.cs
```diff
@@ -1391,54 +1415,107 @@ private static Process StartProcess(string hostname, string benchmarksRepo, Serv
...
+                element.SetAttributeValue(name, value);
+            }
+
+            using (var resourceStream = Assembly.GetCallingAssembly().GetManifestResourceStream("BenchmarksServer.applicationHost.config"))
+            {
+                var applicationHostConfig = XDocument.Load(resourceStream);
+                SetAttribute(applicationHostConfig, "/configuration/system.webServer/aspNetCore", "processPath", executable);
+                SetAttribute(applicationHostConfig, "/configuration/system.webServer/aspNetCore", "arguments", arguments);
+
+                var ancmPath = Path.Combine(job.BasePath, benchmarksBin, "x64\\aspnetcore.dll");
```
shirhatti: This path isn't correct. The shim won't be in the bin folder

src/BenchmarksServer/Startup.cs
davidfowl: Yea, we should send a request. This is similar to what docker does. See https://github.com/aspnet/benchmarks/blob/557be1ebf14297ff0910f905ddd38bb30c99b514/src/BenchmarksServer/Startup.cs#L684-L700

shirhatti:
```
D:\benchmarks\src\BenchmarksDriver [pakrym/IIS ≡ +0 ~2 -0 !]> dotnet run -c release -- --server "http://13.93.212.70:5001" --client "http://104.42.10.121:5002" -j ..\Benchmarks\benchmarks.plaintext.json -n PlaintextNonPipelined --webHost IISInProcess
[11:58:53.073] Using worker Wrk
[11:58:53.161] Running session '3aca0069ee6e42ecb5dc347c87b7bef5' with description ''
[11:58:53.162] Starting scenario PlaintextNonPipelined on benchmark server...
[11:59:07.615] Warmup
[11:59:07.622] Starting scenario PlaintextNonPipelined on benchmark client...
[11:59:23.521] Scenario PlaintextNonPipelined completed on benchmark client
[11:59:23.523] Stopping scenario PlaintextNonPipelined on benchmark client...
[11:59:23.545] Starting scenario PlaintextNonPipelined on benchmark client...
[11:59:39.443] Scenario PlaintextNonPipelined completed on benchmark client
[11:59:39.444] Stopping scenario PlaintextNonPipelined on benchmark client...
[11:59:39.502] RequestsPerSecond:           36758
[11:59:39.502] Latency on load (ms):        7.7
[11:59:39.506] Max CPU (%):                 9940
[11:59:39.506] WorkingSet (MB):             685
[11:59:39.507] Startup Main (ms):           1044
[11:59:39.508] First Request (ms):          5.8
[11:59:39.508] Latency (ms):                4.4
[11:59:39.509] Socket Errors:               0
[11:59:39.510] Bad Responses:               555084
[11:59:39.510] SDK:                         2.2.0-preview1-007522
[11:59:39.511] Runtime:                     2.1.0-preview2-26314-02
[11:59:39.512] ASP.NET Core:                2.1.0-preview3-30383
[11:59:39.512] Stopping scenario PlaintextNonPipelined on benchmark server...
[11:59:41.577] Deleting scenario PlaintextNonPipelined on benchmark server...
```

Any ideas on why I'm seeing Bad Responses?

pakrym:
@shirhatti are you using benchmarking app from this branch? `-o Benchmarks@pakrym/IIS`

src/BenchmarksServer/Startup.cs
pakrym: We shouldn't send requests or we would skew first request time statistics. Socket being able to connect is enough.

src/BenchmarksServer/Startup.cs
pakrym: No idea.
/cc @sebastienros .

src/BenchmarksServer/Startup.cs
pakrym: The year where it still works :)

src/BenchmarksServer/Startup.cs
pakrym: Sorry, I don't follow.

pakrym:
> We need to explode if the machine you're running on isn't a windows machine when you attempt to use the IIS* WebHost

It won't find w3wp and explode.

src/BenchmarksServer/applicationHost.config
DamianEdwards: I think that's reasonable for benchmarks, as I know many folks do that in their own apps today. For the stress (long running) scenarios, we could add them back, or have two versions to help get better coverage of different environments.

src/BenchmarksServer/Startup.cs
sebastienros: The idea is that if we don't do a `dotnet publish` we'll use the runtime store. This was done initially to measure the impact with or without it, but in practice we only measure a published application. We can ignore for now and I'll remove this code path once we decide to get rid of it.

src/BenchmarksDriver/Program.cs
mikeharder: I believe TechEmpower uses `Host: TFB-server`:

https://github.com/TechEmpower/FrameworkBenchmarks/search?utf8=%E2%9C%93&q=server_host&type=

Is there any reason to think adding `:port` would make the server slower?  If so, we should measure the real-world impact of changing this header before merging.

src/BenchmarksServer/Startup.cs
mikeharder: @sebastienros: I think @davidfowl is asking why it's `netcoreapp2.1` rather than `netcoreapp2.0`.

sebastienros: Now that you say it it's obvious ;) That's an issue then, filing a bug.

src/BenchmarksDriver/Program.cs
sebastienros: FYI, related to https://github.com/aspnet/benchmarks/issues/381

src/BenchmarksServer/Startup.cs
mikeharder: It would be good to compare the perf of 3 configs:

1. netcoreapp2.0 with 2.0 packages
2. netcoreapp2.0 with 2.1 packages
3. netcoreapp2.1 with 2.1 packages

I'm not sure if there will be any difference between 2 and 3, but 1 and 3 are the recommended customer configurations.

src/BenchmarksServer/Startup.cs
jkotalik: Move this: https://github.com/aspnet/benchmarks/pull/420/files/557be1ebf14297ff0910f905ddd38bb30c99b514#diff-d3807a7e35147b0c86da6287c24065b7R1364 to https://github.com/aspnet/benchmarks/blob/557be1ebf14297ff0910f905ddd38bb30c99b514/src/BenchmarksServer/Startup.cs#L1304

src/BenchmarksDriver/Program.cs
pakrym: I asked @halter73 about this, he said that is shouldn't affect results.

src/BenchmarksServer/Startup.cs
pakrym: But we still need to get dotnet path even for IIS case

src/BenchmarksServer/applicationHost.config
pakrym: @shirhatti do you have the minimal list of required modules? I tried removing all but `AspNetCore` and got an empty page. 

I think I'll merge fist iteration as is file an issue to reduce the module count.

src/BenchmarksDriver/Program.cs
halter73: After @Tratcher changes, Kestrel will validate this a little differently than it did for just "localhost" (e.g. Kestrel will call HttpUtilities.IsNumeric(portChar)), but I can't imagine that having a measurable impact on these benchmarks.

src/BenchmarksServer/Startup.cs
halter73: Will changing this change the default runtime to 2.1 when I use the driver? If so, 🎉

src/BenchmarksServer/Startup.cs
```diff
@@ -1391,54 +1424,106 @@ private static Process StartProcess(string hostname, string benchmarksRepo, Serv
...
-                        if (perfview)
-                        {
-                            job.PerfViewTraceFile = Path.Combine(job.BasePath, "benchmarks.etl");
-                            var perfViewArguments = new Dictionary<string, string>();
-                            perfViewArguments["AcceptEula"] = "";
-                            perfViewArguments["NoGui"] = "";
-                            perfViewArguments["Process"] = process.Id.ToString();
+            if (iis)
+            {
+                await WaitToListen(job, hostname);
```
halter73: This seems even better than waiting for stdout from Kestrel. Maybe we should use this logic for all servers.

PR #421, opened 2018-03-20
Fix bad PerfView collection parameter

vancem:
We use the /NoRundown switch on stop.
This is not meant for general use, and causes the modules of methods to not
be resolved as well as methods (which means that source code lookup also
fails.

I also limited the collection to a circular buffer of 1 GB by default.
This avoids creating very large files which are generally not useful.
ALso increased the bufferSize to 1GB to avoid lost events (this scenario
CAN produce alot of events).

@davidfowl  @steveharter 

David, Steve:  If we check this in, does it immediately become live on say asp-perf-win?  
If not what is needed to make them take the update?

### DIFF
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -29,7 +29,7 @@ namespace BenchmarksDriver
         private static string _tableName = "AspNetBenchmarks";
 
         // Default to arguments which should be sufficient for collecting trace of default Plaintext run
-        private const string _defaultTraceArguments = "";
+        private const string _defaultTraceArguments = "BufferSizeMB=1024;CircularMB=1024";
 
         public static int Main(string[] args)
         {
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -29,7 +29,7 @@ namespace BenchmarksDriver
         private static string _tableName = "AspNetBenchmarks";
 
         // Default to arguments which should be sufficient for collecting trace of default Plaintext run
-        private const string _defaultTraceArguments = "BufferSize=1024";
+        private const string _defaultTraceArguments = "";
 
         public static int Main(string[] args)
         {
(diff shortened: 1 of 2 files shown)


vancem:
By the way the CPU counters we have available (I am pretty sure) are these 
```
TotalIssues
BranchInstructions
CacheMisses
BranchMispredictions
TotalCycles
UnhaltedCoreCycles
InstructionRetired
UnhaltedReferenceCycles
LLCReference
LLCMisses
BranchInstructionRetired	
BranchMispredictsRetired	
```
LLC probably stands for Lowest Level Cache.     CacheMisses is probably a synonym for that, (I am still validating).   

You can turn up to 4 of these on at once (along with normal CPU sampling) using the perfVIew option

 *  /CpuCounters=InstructionRetired:1000000,CacheMisses:100000

The number represents the interval between taking stack samples.    I will have more information soon.   I will probably make a markdown file on instructions for doing investigations like this.

sebastienros:
This will require me to do a deployment once ready, but will be super quick. Will comment here when available,

I have questions though,

We already set `/BufferSize` from the BenchmarksClient application directly as we can pass more arguments from it. We also use `1024` by default, however you seem to use a different one, named `BufferSizeMB`. I found mention of both in the documentation, which one is the right one?

Apparently we tried `CircularMB` by the past, and it outputed a message saying it wasn't supported in command line. Did it change recently (we are using perfview 2.0.2, which I can bump if necessary)

sebastienros:
I will deploy the `NoRundown` right now, and once I get more info about the default values we can merge it which doesn't require any deployment.

vancem:
I have updated the PR slightly, I did not see that there was a default-arguments concept already.  I agree it should go there.   

To answer your questions

1. BufferSize, and BufferSizeMB are synonyms (however the second is preferred, because it is clear what the units are (the first one is kept only for compatibility).   
2. In general anything you can do from the GUI you can do from the command line.  In particular /CircularMB has always works (I just tried it).  
3. ETW is inherently machine wide.   The /Process qualifier is only present for other scenarios like taking a heap snapshot, or triggering a stop on a particular ETW event.     You should remove it as it does not server a useful purpose.  

All of this is just convenience (because users can fix it by passing parameters) EXCEPT for the /NoRundown (which is on stop so it can't be overridden).

mikeharder:
The error we hit earlier was trying to use `InMemoryCircularBuffer` with separate Start and Stop processes, which is blocked:

https://github.com/Microsoft/perfview/blob/master/src/PerfView/App.cs#L192

But `InMemoryCircularBuffer` and `CircularBufferMB` are different options.

sebastienros:
It's now available on asp-perf-win

PR #423, opened 2018-03-21
Add automatic runs for IIS

pakrym:


sebastienros:
I'll merge myself when the servers are updated

PR #427, opened 2018-03-23
Continuous jobs

sebastienros:
Removing dependency on KoreBuild
Moving msbuild file based scenarios to .sh script


sebastienros:
I will address the build/cli issues. Though I think only `Platforms` requires 2.1 to build.

As for the jobs reports:
- We need to know when something goes wrong automatically (email)
- Be able to analyze the runs logs without logging

Proposed actions:
a- As this PR is already generating a log file with all the driver outputs like the CI does, we can send an email with this log file if there is an exit code > 0. 
To be able to look at past jobs we can expose an html file with links to scenarios and results metadata. Each run will then append a line to this file.
b- Use a very simple CI solution that would run the jobs, alert, and serve reports.

sebastienros:
A CI might be a solution for more that just this, by handling the reliability runs too (4 different jobs right now, signalr to come). And it would also handle Citrine runs once we have a benchmarks relay.

I couldn't do it on TC because each long running job would take a full agent, same thing for scenarios (we have 4 sets now), but having a dedicated CI machine for it is fine. And any solution would be ok as the need is just about running a script.

If you agree with that, suggest products (jenkins, buildbot, ...)
It needs to run scheduled jobs (not just source trigger), handle long running jobs, be able to run multiple jobs concurrently, provide http reports, and alerting. I assume there are very simple requirements that any CI would handle OOB.

Issue #429, opened 2018-03-25
Update Results spreadsheet 


grahamehorner:
Please update the results spreadsheet to show current performance and and IIS inproc/outproc metrics

Eilon:
I think we have a public Power BI page that has live results: https://aka.ms/aspnet/benchmarks (I *think* it's public!)

davidfowl:
We should delete the spread sheet.

Eilon:
Update the spreadsheet to have one cell that is a link to that? 😄

PR #430, opened 2018-03-26
Removed very old results spread sheet

davidfowl:
Fixes https://github.com/aspnet/benchmarks/issues/429


davidfowl:
> Approved, but my idea was better.

I tried to downvote this but it seems you can't do it on PR approvals 😄

PR #433, opened 2018-03-27
Add new SignalR scenarios

BrennanConroy:
@sebastienros Should I close #431 ?


sebastienros:
I think you should keep it, and merge it. And in this PR just update the scenarios. I'll merge them both.
I will merge continuous after that.

BrennanConroy:
Ok, fixed up

PR #434, opened 2018-03-28
Harness improvements

vancem:
1. Start PerfView before the benchmark starts.  This allows perfVIew to see the startup which is valuable.
2. Stamp the files by default with the date and RPS value.  Also allow the user
to specify the file prefix (which can be a directory).   This avoids having to rename (or worse loose track of which data is which).

Note that the diff in startup is a bit confusing.  It looks like I removed stuff, but all I did was move the PerfView logic from one function to another (to before benchmark start).  

@sebastienros

### DIFF
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -118,7 +118,7 @@ namespace BenchmarksDriver
                 $"Arguments used when collecting a PerfView trace.  Defaults to \"{_defaultTraceArguments}\".",
                 CommandOptionType.SingleValue);
             var traceOutputOption = app.Option("--trace-output",
-                "An optional location to download the trace file to, e.g., --trace-output c:\traces", CommandOptionType.SingleValue);
+                "Can be a file prefix (app will add *.DATE.etl.zip) , or a specific name (end in *.etl.zip) and no DATE will be added e.g. --trace-output c:\traces\trace", CommandOptionType.SingleValue);
             var disableR2ROption = app.Option("--no-crossgen",
                 "Disables Ready To Run.", CommandOptionType.NoValue);
             var collectR2RLogOption = app.Option("--collect-crossgen",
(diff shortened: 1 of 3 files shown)


sebastienros:
Code-wise I am ok with the changes. I will assume everyone involved with perfview traces should be able to identify where the actual load is starting, and not get confused by the fact the numbers include startup traces too.

If nobody emits concerns I will merge it.

vancem:
As it is right now, the PerfView trace shows both warmup and the test, so you have to do that kind of analysis already.    I have put two, 200msec sleeps, one after startup, and one after warmup, that lets you find the regions with reasonable ease (so you can pick the region you care about). 

The value of starting the trace early is that there is valuable information at startup (e.g. whether you are JITTing and what exactly, The version numbers of all the DLLs, as well as frankly the performance of startup (which we are likely to care about at some point).   

Finally one very nice thing we can now do is to instrument the benchmark with EventSource/DiagnosticSource/ILogger  events that tell us the configuration values of the benchmark.  Typically you read this early, and the won't show up unless PerfView is started early.

vancem:
Thanks @sebastienros.   I am running into cases where it would be useful to this update working in the server for  asp-perf-win at least.   When is that likely to happen?

Thanks

Vance

sebastienros:
I think I did it right after I merged. Or did you try and it didn't work?
________________________________

vancem:
I did try it and it does not seem to work (it has the old behavior).  In particular I would have expected to see the 'Process-Start' of the benchmark program (since it is now AFTER PerfView start), but I don't see that.

sebastienros:
Ok I will fix it. Should be done in 30 minutes.

vancem:
Thanks!

sebastienros:
I fixed an issue (ProcessId) and tested it successfully.

vancem:
Works for me as well.  Thanks!

PR #441, opened 2018-04-07
Increase StringBuilder initial capacity for batch update query string generation

jo-ninja:
Query count (1, 5, 10, 15...) is being used as StringBuilder initial capacity:
https://github.com/aspnet/benchmarks/blob/e135b1e439c5174cc46b2cf7d87f3bf04d461442/src/Benchmarks/Data/DapperDb.cs#L64-L68

https://github.com/aspnet/benchmarks/blob/281d5dd82054819f02d8111253c36ff5f9e61f66/src/Benchmarks/Data/RawDb.cs#L90-L94

In most of the cases, it is lower than the default capacity of 16. 
Wonder if increasing this value should contribute to a better result, I tried to simulate it here:
https://gist.github.com/jo-ninja/33d05778f4c283f1c27f63d8beb38743

Result:
image

It seems that with the initial capacity of 64x, overall it's about 1.5x faster and a slightly lower memory usage.
/cc @sebastienros @benaadams @davidfowl 
But it is a nanoseconds world, so I don't think it will really help... 😶

### DIFF
--- a/src/Benchmarks/Data/DapperDb.cs
+++ b/src/Benchmarks/Data/DapperDb.cs
@@ -65,7 +65,7 @@ namespace Benchmarks.Data
         {
             var results = new World[count];
             IDictionary<string, object> parameters = new ExpandoObject();
-            var updateCommand = new StringBuilder(count);
+            var updateCommand = new StringBuilder(count * 64);
 
             using (var db = _dbProviderFactory.CreateConnection())
             {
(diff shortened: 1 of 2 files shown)


Tornhoof:
Not directly related, but in your first code snippet is an ``ExpandoObject``. There is a specific Parameter type for dapper  (``DynamicParameter`` I think). Might save another few nano seconds.

jo-ninja:
Yes, I knew it from the Community Standup video last month 😄 What an amazing work!
I am closing this as we can't spot the throughput improvements here.

PR #442, opened 2018-04-07
Use a newer version of PerfView when collecting traces.

vancem:
There are some fixes in a later version of PerfView that would be nice to have.   Upgrade to latest official version.


vancem:
@sebastienros

sebastienros:
Will be available when I deploy the servers. This WE probably.

vancem:
Thanks

PR #446, opened 2018-04-14
Update SignalRWorker

BrennanConroy:
@anurse @sebastienros

### DIFF
--- a/src/BenchmarksClient/Workers/SignalRWorker.cs
+++ b/src/BenchmarksClient/Workers/SignalRWorker.cs
@@ -174,7 +174,7 @@ namespace BenchmarksClient.Workers
                 var tasks = new List<Task>(_connections.Count);
                 foreach (var connection in _connections)
                 {
-                    tasks.Add(connection.StopAsync());
+                    tasks.Add(connection.DisposeAsync());
                 }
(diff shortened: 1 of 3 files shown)


BrennanConroy:
🆙 📅 
Just take a quick glance to make sure I didn't do anything weird

src/BenchmarksClient/Workers/SignalRWorker.cs
```diff
@@ -159,23 +159,22 @@ public SignalRWorker(ClientJob job)
...
-
-                _stopped = true;
-
                 // stop connections
                 Log("Stopping connections");
                 var tasks = new List<Task>(_connections.Count);
                 foreach (var connection in _connections)
                 {
-                    tasks.Add(connection.StopAsync());
+                    tasks.Add(connection.DisposeAsync());
```
davidfowl: Because we never restart?

BrennanConroy: Yeah, each run is completely separate

davidfowl: Good this will free up http clients 😄

PR #448, opened 2018-04-20
Making Workers handle multiple jobs

mikaelm12:
This is still very WIP. Just want some initial feedback.

The initial idea is to have SpanIds associated with jobs that we want to be run by the same worker.This means I had to change the worker internals to to not be tied to a single client job instance but instead make it such that a single worker instance can be started multiple times with different jobs.

This is work to allow us to do more stress testing : https://github.com/aspnet/SignalR/issues/1700

Roast Me.

### DIFF
--- a/src/Benchmarks.ClientJob/ClientJob.cs
+++ b/src/Benchmarks.ClientJob/ClientJob.cs
@@ -19,6 +19,8 @@ namespace Benchmarks.ClientJob
         {
             Id = clientJob.Id;
             Threads = clientJob.Threads;
+            SpanId = clientJob.SpanId;
+            SendDelay = clientJob.SendDelay;
             Connections = clientJob.Connections;
             Duration = clientJob.Duration;
             ClientProperties = new Dictionary<string, string>(clientJob.ClientProperties);
@@ -38,9 +40,14 @@ namespace Benchmarks.ClientJob
 
         public int Id { get; set; }
 
+        public string SpanId { get; set; }
+
         public Worker Client { get; set; } = Worker.Wrk;
 
         public int Threads { get; set; } = 32;
+
+        // Send delay in minutes for long running connection tests.
+        public TimeSpan SendDelay { get; set; } = TimeSpan.FromMinutes(10);
 
         public int Connections { get; set; } = 256;
 
@@ -57,7 +64,7 @@ namespace Benchmarks.ClientJob
         public string Query { get; set; }
 
         [JsonConverter(typeof(StringEnumConverter))]
-        public ClientState State { get; set; }
+        public ClientJobState State { get; set; }
 
         public double RequestsPerSecond { get; set; }
         public int Requests { get; set; }
--- a/src/Benchmarks.ClientJob/ClientJob.cs
+++ b/src/Benchmarks.ClientJob/ClientJob.cs
@@ -20,7 +20,6 @@ namespace Benchmarks.ClientJob
             Id = clientJob.Id;
             Threads = clientJob.Threads;
             SpanId = clientJob.SpanId;
-            SendDelay = clientJob.SendDelay;
             Connections = clientJob.Connections;
             Duration = clientJob.Duration;
             ClientProperties = new Dictionary<string, string>(clientJob.ClientProperties);
@@ -45,9 +44,6 @@ namespace Benchmarks.ClientJob
         public Worker Client { get; set; } = Worker.Wrk;
 
         public int Threads { get; set; } = 32;
-
-        // Send delay in minutes for long running connection tests.
-        public TimeSpan SendDelay { get; set; } = TimeSpan.FromMinutes(10);
 
         public int Connections { get; set; } = 256;
 
--- a/src/BenchmarksClient/Startup.cs
+++ b/src/BenchmarksClient/Startup.cs
@@ -111,6 +111,7 @@ namespace BenchmarkClient
                     if (job.State == ClientJobState.Waiting)
                     {
                         Log($"Starting '{job.Client}' worker");
+                        Log($"Current Job SpanId '{job.SpanId}'");
                         job.State = ClientJobState.Starting;
 
                         try
--- a/src/BenchmarksClient/Startup.cs
+++ b/src/BenchmarksClient/Startup.cs
@@ -169,20 +169,26 @@ namespace BenchmarkClient
                 await Task.Delay(100);
 
                 allJobs = _jobs.GetAll();
-                job = allJobs.FirstOrDefault(clientJob =>
+                if (job != null)
                 {
-                    return string.Equals(clientJob.SpanId, job.SpanId);
-                });
-                // If no more jobs with the span id exist then we can clear
-                // out the worker to signal to the worker factory to create
-                // a new worker.
-                if (job == null && worker != null)
+                    job = allJobs.FirstOrDefault(clientJob =>
+                    {
+                        return string.Equals(clientJob.SpanId, job.SpanId);
+                    });
+                }
+                else 
                 {
-                    await worker.DisposeAsync();
-                    worker = null;
-
                     // Get another job for the new worker we are going to create
                     job = allJobs.FirstOrDefault();
+
+                    // No more jobs with the same span id exist so we can clear
+                    // out the worker to signal to the worker factory to create
+                    // a new one.
+                    if (worker != null)
+                    {
+                        await worker.DisposeAsync();
+                        worker = null;
+                    }
                 }
             }
         }
--- a/src/BenchmarksClient/Startup.cs
+++ b/src/BenchmarksClient/Startup.cs
@@ -176,7 +176,8 @@ namespace BenchmarkClient
                         return string.Equals(clientJob.SpanId, job.SpanId);
                     });
                 }
-                else 
+                // job will be null if there aren't any more jobs with the same spanId.
+                if (job == null)
                 {
                     // Get another job for the new worker we are going to create
                     job = allJobs.FirstOrDefault();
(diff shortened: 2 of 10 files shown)


src/BenchmarksClient/Startup.cs
```diff
@@ -100,72 +100,89 @@ public static int Main(string[] args)
...
 
                         try
                         {
-                            worker = WorkerFactory.CreateWorker(job);
+                            if (worker == null)
+                            {
+                                worker = WorkerFactory.CreateWorker(job);
+                            }
 
                             if (worker == null)
```
sebastienros: There should be an `else` here and check the if `SpanId` matches the current `worker`. If not, `continue` skip the job.

src/BenchmarksClient/Workers/SignalRWorker.cs
```diff
@@ -147,10 +159,10 @@ public SignalRWorker(ClientJob job)
             }
 
             cts.Token.WaitHandle.WaitOne();
-            await StopAsync();
+            await StopJobAsync();
```
sebastienros: `Stop` is called from both the Client service and the worker?

src/BenchmarksClient/Workers/WorkerFactory.cs
```diff
@@ -15,10 +15,10 @@ static public class WorkerFactory
         static WorkerFactory()
         {
             // Wrk
-            Workers[Worker.Wrk] = clientJob => new WrkWorker(clientJob);
+            Workers[Worker.Wrk] = clientJob => new WrkWorker();
```
sebastienros: I assume we can just remove the `clientJob` argument and pass it in `Start` and `Stop`

src/BenchmarksClient/Startup.cs
mikaelm12: https://github.com/aspnet/benchmarks/pull/448/files#diff-d375e6377c60b0a87d2709204a22dd73R174
This should handle that logic. The first job we select is just the first one queued but after that we only select jobs that have the same SpanId

src/BenchmarksClient/Workers/SignalRWorker.cs
mikaelm12: Hm, this was the previous implementation.

mikaelm12: Actually this is handled by a check in Stopped. There is a stopped flag that is checked to make sure that we don't call stop more than once but your comment also made me realize that we need to flip this flag back every time start is called.

src/BenchmarksClient/Workers/WorkerFactory.cs
mikaelm12: Yeah, we don't need to pass the clientJob into the worker constructors anymore

src/BenchmarksClient/Startup.cs
```diff
@@ -100,72 +100,88 @@ public static int Main(string[] args)
...
                             _jobs.Remove(job.Id);
                         }
                     }
                 }
                 await Task.Delay(100);
+
+                allJobs = _jobs.GetAll();
+                job = allJobs.FirstOrDefault(clientJob =>
+                {
+                    return string.Equals(clientJob.SpanId, job.SpanId);
```
BrennanConroy: I don't think you will run anymore jobs after the first one finishes.

sebastienros: Right, you need to loop for the first available one that matches the current span id, and timeout the current worker if no more job is coming.

src/BenchmarksClient/Workers/SignalRWorker.cs
```diff
@@ -135,6 +139,16 @@ public SignalRWorker(ClientJob job)
                             }
                         }
                         break;
+                    case "echoLongRunning":
+                        while (!cts.IsCancellationRequested)
+                        {
+                            await Task.Delay(TimeSpan.FromMinutes(30));
```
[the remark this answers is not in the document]
mikaelm12: Changing this to be passed similar to the other client job options

mikaelm12:
Currently rebasing. Have to update the new worker to agree with the the new IWorker interface

mikaelm12:
Updated

src/Benchmarks.ClientJob/ClientJob.cs
```diff
@@ -38,10 +40,15 @@ public ClientJob(ClientJob clientJob)
...
         public int Id { get; set; }
 
+        public string SpanId { get; set; }
+
         public Worker Client { get; set; } = Worker.Wrk;
 
         public int Threads { get; set; } = 32;
 
+        // Send delay in minutes for long running connection tests.
+        public TimeSpan SendDelay { get; set; } = TimeSpan.FromMinutes(10);
```
BrennanConroy: This belongs on `ClientProperties` not a separate field

src/BenchmarksClient/Startup.cs
```diff
@@ -100,72 +100,91 @@ public static int Main(string[] args)
...
+                    else if (job.State == ClientJobState.Deleting)
                     {
                         Log($"Deleting job {worker?.JobLogText ?? "no worker found"}");
 
                         try
                         {
                             if (worker != null)
                             {
-                                await worker.StopAsync();
+                                // We stop the worker but
```
BrennanConroy: but what?

src/BenchmarksClient/Startup.cs
```diff
@@ -100,72 +100,91 @@ public static int Main(string[] args)
...
+                // If no more jobs with the span id exist then we can clear
+                // out the worker to signal to the worker factory to create
+                // a new worker.
+                if (job == null && worker != null)
+                {
+                    await worker.DisposeAsync();
+                    worker = null;
+
+                    // Get another job for the new worker we are going to create
+                    job = allJobs.FirstOrDefault();
```
BrennanConroy: If there are no other jobs in the queue yet you'll never get a new job it looks like

src/BenchmarksClient/Workers/SignalRWorker.cs
```diff
@@ -135,6 +139,16 @@ public SignalRWorker(ClientJob job)
                             }
                         }
                         break;
+                    case "echoLongRunning":
+                        while (!cts.IsCancellationRequested)
+                        {
+                            await Task.Delay(_job.SendDelay);
+                            for (var i = 0; i < _connections.Count; i++)
+                            {
+                                _ = _connections[i].SendAsync("EchoAll", DateTime.UtcNow);
```
BrennanConroy: you can await this

src/BenchmarksClient/Workers/SignalRWorker.cs
```diff
@@ -162,25 +176,8 @@ public SignalRWorker(ClientJob job)
...
-                {
-                    tasks.Add(connection.DisposeAsync());
-                }
-
                 CalculateStatistics();
 
-                await Task.WhenAll(tasks);
-
                 // TODO: Remove when clients no longer take a long time to "cool down"
                 await Task.Delay(5000);
```
BrennanConroy: move this to the dispose function

src/BenchmarksClient/Workers/SignalRWorker.cs
```diff
@@ -189,10 +186,33 @@ public SignalRWorker(ClientJob job)
...
+            // stop connections
+            Log("Stopping connections");
+            var tasks = new List<Task>(_connections.Count);
+            foreach (var connection in _connections)
+            {
+                tasks.Add(connection.DisposeAsync());
+            }
+
+            await Task.WhenAll(tasks);
+            _httpClientHandler.Dispose();
```
BrennanConroy: log that connections have been disposed

src/BenchmarksDriver/Program.cs
```diff
@@ -170,6 +170,8 @@ public static int Main(string[] args)
                 CommandOptionType.SingleValue);
             var jobsOptions = app.Option("-j|--jobs",
                 "The path or url to the jobs definition.", CommandOptionType.SingleValue);
+            var sendDelayOption = app.Option("-- sendDelay",
```
BrennanConroy: remove this option

src/BenchmarksClient/Startup.cs
```diff
@@ -100,72 +100,96 @@ public static int Main(string[] args)
...
+
+                allJobs = _jobs.GetAll();
+                if (job != null)
+                {
+                    job = allJobs.FirstOrDefault(clientJob =>
+                    {
+                        return string.Equals(clientJob.SpanId, job.SpanId);
+                    });
+                }
+                else
```
BrennanConroy: make this `if (job == null)` instead of an else
It'll save 100ms per job 😀

src/BenchmarksClient/Workers/SignalRWorker.cs
```diff
@@ -84,14 +85,23 @@ public SignalRWorker(ClientJob job)
                 throw new Exception("Scenario wasn't specified");
             }
 
+            if(_job.ClientProperties.TryGetValue("SendDelay", out var sendDelay))
```
BrennanConroy: nit: space

src/BenchmarksClient/Workers/SignalRWorker.cs
```diff
@@ -135,6 +145,16 @@ public SignalRWorker(ClientJob job)
                             }
                         }
                         break;
+                    case "echoLongRunning":
+                        while (!cts.IsCancellationRequested)
+                        {
+                            await Task.Delay(_sendDelay);
```
BrennanConroy: Maybe we should do this differently. If the send delay isn't an even division of the job duration then 2 things can happen:
1. The driver sees a "hanging" job and kills the job
2. We never send the messages

mikaelm12:
Closing this for a little bit since I'll be pushing a lot of changes while debugging

PR #455, opened 2018-04-30
Allow equals = in arguments passed to --trace-arguments parameter.

vancem:
Some PerfVIew arguments may need to pass strings with = characters in them.
Fix it so that this is allowed (add a count to the split that is used, so only the first = is treated specially).

@sebastienros


sebastienros:
It's deployed now

vancem:
Thanks for the rapid turn around.   It really does make a difference for me.

PR #460, opened 2018-05-09
More realistic echo scenario

BrennanConroy:
@davidfowl @anurse


src/BenchmarksClient/Workers/SignalRWorker.cs
```diff
@@ -304,6 +292,30 @@ private void CreateConnections(HttpTransportType transportType = HttpTransportTy
...
         }
 
+        private void ReceivedDateTime(DateTime dateTime, int connectionId)
+        {
+            if (_stopped)
+            {
+                return;
+            }
+
+            // TODO: Collect all the things
```
davidfowl: ..

Issue #462, opened 2018-05-11
Disable IIS benchmark logs


pakrym:
It takes ridiculous amount of space on disk and may affect performance.

sebastienros:
ooohhhhhh!

PR #463, opened 2018-05-11
dontLog=true

pakrym:
https://github.com/aspnet/benchmarks/issues/462


pakrym:
This seems to gain ~10K RPS.

PR #464, opened 2018-05-13
Use NTP type protocol to get more accurate latency

BrennanConroy:
@sebastienros @davidfowl @anurse

### DIFF
--- a/src/BenchmarksClient/Workers/SignalRWorker.cs
+++ b/src/BenchmarksClient/Workers/SignalRWorker.cs
@@ -115,7 +115,7 @@ namespace BenchmarksClient.Workers
                 switch (_scenario)
                 {
                     case "broadcast":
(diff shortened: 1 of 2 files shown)


src/BenchmarksClient/Workers/SignalRWorker.cs
```diff
@@ -404,6 +407,30 @@ private double GetPercentile(int percent, List<double> sortedData)
             return (1.0 - fractionPart) * sortedData[(int)Math.Truncate(i)] + fractionPart * sortedData[(int)Math.Ceiling(i)];
         }
 
+        private async Task DoNTP()
```
anurse: Are you literally using the NTP protocol? If not, I suggest we just call this "Time Synchronization" or something. Don't want to confuse people.

PR #471, opened 2018-05-18
Add ClientProperties to DB

BrennanConroy:
Already updated the DBs

### DIFF
--- a/src/BenchmarksDriver/Serializers/SignalRSerializer.cs
+++ b/src/BenchmarksDriver/Serializers/SignalRSerializer.cs
@@ -39,7 +39,7 @@ namespace BenchmarksDriver.Serializers
                         [WebHost] [nvarchar](50) NOT NULL,
                         [Transport] [nvarchar](50) NOT NULL,
                         [HubProtocol] [nvarchar](50) NOT NULL,
-                        [ClientProperties] [nvarchar](200) NOT NULL,
+                        [ClientProperties] [nvarchar](max) NOT NULL,
                         [Connections] [int] NOT NULL,
                         [Duration] [int] NOT NULL,
                         [Path] [nvarchar](200) NULL,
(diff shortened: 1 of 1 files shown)


src/BenchmarksDriver/Serializers/SignalRSerializer.cs
```diff
@@ -195,6 +198,7 @@ class SignalRSerializer : IResultsSerializer
                     p.AddWithValue("@WebHost", serverJob.WebHost.ToString());
                     p.AddWithValue("@Transport", clientJob.ClientProperties["TransportType"]);
                     p.AddWithValue("@HubProtocol", clientJob.ClientProperties["HubProtocol"]);
+                    p.AddWithValue("@ClientProperties", string.Join(',', clientJob.ClientProperties.ToArray()));
```
sebastienros: That, or a json serialized array in case you need to process it programatically ?

sebastienros: Or a json document, with property names.

src/BenchmarksDriver/Serializers/SignalRSerializer.cs
```diff
@@ -39,6 +39,7 @@ class SignalRSerializer : IResultsSerializer
                         [WebHost] [nvarchar](50) NOT NULL,
                         [Transport] [nvarchar](50) NOT NULL,
                         [HubProtocol] [nvarchar](50) NOT NULL,
+                        [ClientProperties] [nvarchar](200) NOT NULL,
```
sebastienros: Not big enough, you'll get properties that are bigger than this. Try 2K.

src/BenchmarksDriver/Serializers/SignalRSerializer.cs
BrennanConroy: `string.Split` not good enough?

sebastienros: There is no escaping, so no guaranty you can deserialize it with a split.

BrennanConroy:
🆙 📅

PR #472, opened 2018-05-21
Run SQL off of the main loop

BrennanConroy:
Since this was in the main driver loop, if the SQL query was too slow or timed out it could cause the server to close because the driver hasn't communicated in 30 seconds

### DIFF
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -1080,7 +1080,6 @@ namespace BenchmarksDriver
                                     // Headers
                                     foreach (var field in fields)
                                     {
-                                        
                                         var size = Math.Max(field.Key.Length, field.Value.Length);
                                         header.Append("| ").Append(field.Key.PadLeft(size)).Append(" ");
                                         separator.Append("| ").Append(new String('-', size)).Append(" ");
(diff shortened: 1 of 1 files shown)


sebastienros:
Good catch, I wondered why the retry logic didn't work and the jobs were still failing.

src/BenchmarksDriver/Program.cs
```diff
@@ -1139,6 +1142,8 @@ public static int Main(string[] args)
                         spanLoop = spanLoop + 1;
                     } while (DateTime.UtcNow - startTime < span);
 
+                    await sqlTask;
```
sebastienros: Check if not yet completed, and in this case write a message?

src/BenchmarksDriver/Program.cs
```diff
@@ -1142,7 +1142,18 @@ public static int Main(string[] args)
...
 
-                    await sqlTask;
+                    if (!sqlTask.IsCompleted)
+                    {
+                        Log("Job finished, waiting for SQL to complete.");
+                        try
+                        {
+                            await sqlTask;
+                        }
+                        catch (Exception ex)
```
sebastienros: Looks too easy to just catch the exception and display it ... but I don't know better.

BrennanConroy: Oh wait, I already catch in the `ContinueWith` I'll remove this

PR #473, opened 2018-05-23
Use SQL transactions to avoid accidental duplicate inserts

BrennanConroy:
@sebastienros This should avoid the dupe issue. I can add it to the `WrkSerializer` if you're ok with the change


sebastienros:
ok do it

sebastienros:
beat you

PR #475, opened 2018-05-23
Fix bug in percentile calculation

BrennanConroy:
@sebastienros 

Off by 1 error, could cause index out of bounds exceptions. Also, added support for 100th Percentile just cause.


sebastienros:
What about HyperLogLog ;)

PR #482, opened 2018-05-25
Get max latency value

BrennanConroy:
Didn't write the regex for wrk

### DIFF
--- a/src/BenchmarksClient/Workers/WrkWorker.cs
+++ b/src/BenchmarksClient/Workers/WrkWorker.cs
@@ -233,6 +233,9 @@ namespace BenchmarksClient.Workers
                 var p99Match = Regex.Match(job.Output, String.Format(LatencyPattern, "99%"));
                 job.Latency.Within99thPercentile = ReadLatency(p99Match);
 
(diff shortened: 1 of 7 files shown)


sebastienros:
All good, unless you also prefer to use `MaxLatency`.

Why didn't you write the regex?

BrennanConroy:
I would prefer MaxLatency actually.

Because I don't like regex

BrennanConroy:
Ok, tried to Regex

BrennanConroy:
Ping @sebastienros

src/BenchmarksClient/Workers/WrkWorker.cs
```diff
@@ -233,6 +233,9 @@ private static Process StartProcess(ClientJob job)
                 var p99Match = Regex.Match(job.Output, String.Format(LatencyPattern, "99%"));
                 job.Latency.Within99thPercentile = ReadLatency(p99Match);
 
+                var p100Match = Regex.Match(job.Output, @"\s+Latency\s+[\d\.]+\w+\s+[\d\.]+\w+\s+([\d\.]+)(\w+)");
```
sebastienros: Did you try it?

BrennanConroy: On an online regex visualizer with an example of wrk's output

sebastienros: I also tested it, it works

PR #483, opened 2018-05-25
SqlServer - Use MEMORY_OPTIMIZED tables

benaadams:


sebastienros:
Intent?

benaadams:
Doesn't everyone use memory optimized tables for small tables these days?

[Demonstration: Performance Improvement of In-Memory OLTP](https://docs.microsoft.com/en-us/sql/relational-databases/in-memory-oltp/demonstration-performance-improvement-of-in-memory-oltp?view=sql-server-2017)

```
10453 ms , A: Disk-based table and interpreted Transact-SQL.  
5626 ms , B: memory-optimized table with hash index and interpreted Transact-SQL.  
3937 ms , C: memory-optimized table with hash index and native SP.  
```

benaadams:
[Overview and Usage Scenarios](https://docs.microsoft.com/en-us/sql/relational-databases/in-memory-oltp/overview-and-usage-scenarios?view=sql-server-2017)

> In-Memory OLTP can provide great performance gains, for the right workloads. One customer, bwin, managed to achieve 1.2 Million requests per second with a single machine running SQL Server 2016 (13.x), leveraging In-Memory OLTP. Another customer, Quorum, managed to double their workload while reducing their resource utilization by 70%, by leveraging In-Memory OLTP in SQL Database. While customers have seen up to 30X performance gain in some cases, how much gain you see depends on the workload.

[Performance and Scalability](https://docs.microsoft.com/en-us/sql/relational-databases/in-memory-oltp/introduction-to-memory-optimized-tables?view=sql-server-2017#performance-and-scalability)

> Concurrency: Applications whose performance is affected by engine-level concurrency, such as latch contention or blocking, improves significantly when the application moves to In-Memory OLTP.

sebastienros:
Ok, it might make the queries faster. But should we make the db as fast as we can?
For instance we don't apply optimizations on postgres or mysql to follow what TE is doing, and have some stable engine so that our graphs have only one variable, the runtime and packages.

Today TE is not testing SQL Server, but I can imagine they will at some point. I'd rather wait for their configuration and copy what they do.

So yeah, we could take this change because we are alone on it right now, but I don't really see any reason to when I see our goals of improving the dotnet performance. This is artificial gain. Do we have any proof also it will make a difference on our kind of load (dataset and queries)?

benaadams:
>  This is artificial gain. Do we have any proof also it will make a difference on our kind of load (dataset and queries)?

It pushes more load into the driver and less on the database; so its clear what in the driver needs optimizing.

> Today TE is not testing SQL Server, but I can imagine they will at some point. I'd rather wait for their configuration and copy what they do.

I'd like to suggest a configuration to TE based on this and also do SqlServer tests

benaadams:
Which is why I also went for `SCHEMA_AND_DATA` as it applies to all tables; whereas `SCHEMA_ONLY` only applies to temp tables

sebastienros:
> It pushes more load into the driver and less on the database; so its clear what in the driver needs optimizing.

I agree on that point

> I'd like to suggest a configuration to TE based on this and also do SqlServer tests

If they decide to go with SQL Server, then we will align with anything they do. I'd like to hear from their intention then. And then do the same.

benaadams:
Ok https://github.com/TechEmpower/FrameworkBenchmarks/issues/3786

benaadams:
https://github.com/TechEmpower/FrameworkBenchmarks/pull/3807

benaadams:
PR for comment: https://github.com/TechEmpower/FrameworkBenchmarks/pull/3807

sebastienros:
We don't need it right now.

PR #495, opened 2018-06-24
Add scenarios for MVC benchmark apps

dougbu:
- part of aspnet/Mvc#7805


build/scenarios.sh
```diff
@@ -148,6 +150,18 @@ jobs=(
...
+  "--scenario BasicApi.GetUsingRouteValueWithoutToken $trend $basicApiJobs"
+  "--scenario BasicApi.GetUsingRouteValueWithoutAuthorization $trend $basicApiJobs"
+  "--scenario BasicApi.GetUsingRouteValue $trend $basicApiJobs"
+  "--scenario BasicApi.Post $trend $basicApiJobs"
+
+  # BasicViews
+  "--scenario BasicViews.GetTagHelpers $trend $basicViewsJobs"
+  "--scenario BasicViews.PostWithoutToken $trend $basicViewsJobs"
+  "--scenario BasicViews.PostIgnoringToken $trend $basicViewsJobs"
+  "--scenario BasicViews.Post $trend $basicViewsJobs"
```
rynowak: Do we really need to track all of these things? I imaged that a single Get/Post for each app would be enough. I think it's great that we can **measure** all of the various increments, but I'm no sure how much value we get out of tracking them all.

dougbu: Ah, I misunderstood your comments in aspnet/Mvc@7944 and thought you wanted to track the Get and Post variants over time as well. If we focus on the "full boat" scenarios, the list would be:
- BasicApi.GetUsingRouteValue
- BasicApi.Post
- BasicViews.GetTagHelpers
- BasicViews.Post

Thougths?

rynowak: Sounds great!

dougbu:
86dc85338c

PR #503, opened 2018-08-01
Use OSS package versions consistent with aspnet/Mvc and Microsoft.AspNetcore.All/App 2.1.2

dougbu:
- bump our 2.1.x versions (particularly important in `current` runs) to match OSS dependencies
- get slightly newer SDK to enable builds using our 2.1.2 packages

nits:
- add MySql and PostgreSql-specific services
- fix BenchmarksDriver help typos
- clean up IIS-related dependencies (new in 2.2)
- remove trailing whitespace
- remove an unused `using`

### DIFF
--- a/src/Benchmarks/Startup.cs
+++ b/src/Benchmarks/Startup.cs
@@ -51,15 +51,15 @@ namespace Benchmarks
 
             // Common DB services
             services.AddSingleton<IRandom, DefaultRandom>();
-            services.AddEntityFrameworkSqlServer();
 
             var appSettings = Configuration.Get<AppSettings>();
 
             Console.WriteLine($"Database: {appSettings.Database}");
-            
+
             switch (appSettings.Database)
             {
                 case DatabaseServer.PostgreSql:
+                    services.AddEntityFrameworkNpgsql();
                      var settings = new NpgsqlConnectionStringBuilder(appSettings.ConnectionString);
                      if (!settings.NoResetOnClose)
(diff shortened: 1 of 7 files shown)


src/Benchmarks/Startup.cs
```diff
@@ -84,22 +85,25 @@ public void ConfigureServices(IServiceCollection services)
...
                     services.AddDbContextPool<ApplicationDbContext>(options => options.UseMySql(appSettings.ConnectionString));
 
                     if (Scenarios.Any("Raw") || Scenarios.Any("Dapper"))
                     {
                         services.AddSingleton<DbProviderFactory>(MySql.Data.MySqlClient.MySqlClientFactory.Instance);
                     }
                     break;
+
                 case DatabaseServer.MongoDb:
+                    // Reviewers: Is this actually needed for Mongo scenarios?
```
dougbu: @ajcvickers / @sebastienros ??

src/Benchmarks/Startup.cs
```diff
@@ -84,22 +85,25 @@ public void ConfigureServices(IServiceCollection services)
...
+                    services.AddEntityFrameworkSqlServer();
```
sebastienros: I don't think so. I'd suggest to remove it. Worst case we'll see the mongodb benchmark fail and we can add it back.

divega: Definitively shouldn't be necessary for anything but the SQL Server case.

dougbu:
586f97cb9a

Issue #361, opened 2017-12-21
Add Plaintext Benchmark w Application Insights


benaadams:
Sample rate 0 (so no events are sent) - to record monitor impact of having just having it attached

sebastienros:
The AI team is measuring it internally

PR #538, opened 2018-10-19
[DO NOT MERGE] Add Ai Auto-Collection to Benchmarks

Dmitry-Matveev:
PR to add AI ASP.NET auto-collection to Benchmarks app.
This PR can be used to compare benchmarks with/without AI (based on `EnableAi` setting) from the platform perspective.

The Instrumentation Key provided (in `AiInstrumentationKey` setting) is a temporary AI resource provisioned specifically for this purpose. We can think of changing this IKey / removing this resource after the test or switch to reading it from the environment variable to avoid exposing the IKey.

I do not expect that this PR is to be merged unless @sebastienros would think otherwise.

P.S. I validated that AI collects telemetry with this app if I run Benchmarks.dll locally in Kestrel/Sockets.

### DIFF
--- a/src/Benchmarks/Configuration/AppSettings.cs
+++ b/src/Benchmarks/Configuration/AppSettings.cs
@@ -5,6 +5,10 @@ namespace Benchmarks.Configuration
 {
     public class AppSettings
     {
+        public string AiInstrumenationKey { get; set; }
+
+        public bool EnableAi { get; set; }
+
         public string ConnectionString { get; set; }
 
         public DatabaseServer Database { get; set; } = DatabaseServer.None;
--- a/src/Benchmarks/Configuration/AppSettings.cs
+++ b/src/Benchmarks/Configuration/AppSettings.cs
@@ -5,7 +5,7 @@ namespace Benchmarks.Configuration
 {
     public class AppSettings
     {
-        public string AiInstrumenationKey { get; set; }
+        public string AiInstrumentationKey { get; set; }
 
         public bool EnableAi { get; set; }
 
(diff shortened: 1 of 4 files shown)


davidfowl:
What this for?

Eilon:
I think the idea is to see what impact there is for apps when AppInsights is enabled.

src/Benchmarks/appsettings.json
```diff
@@ -1,3 +1,5 @@
 ﻿{
-  "ConnectionString": "Server=(localdb)\\mssqllocaldb;Database=aspnet5-Benchmarks;Trusted_Connection=True;MultipleActiveResultSets=true"
+  "ConnectionString": "Server=(localdb)\\mssqllocaldb;Database=aspnet5-Benchmarks;Trusted_Connection=True;MultipleActiveResultSets=true",
+  "EnableAI": true,
+  "AiInstrumenationKey": "6a70388f-9193-463b-9117-4a88ba0e04c6"
```
sebastienros: You can remove these, we can pass them by command line args, as it's a source for the app settings. This way we can keep it secret and change it at will.

src/Benchmarks/Configuration/AppSettings.cs
```diff
@@ -5,6 +5,10 @@ namespace Benchmarks.Configuration
 {
     public class AppSettings
     {
+        public string AiInstrumenationKey { get; set; }
```
sebastienros: Small typo (won't matter for the tests)

Dmitry-Matveev:
@sebastienros , addressed PR comments, thanks!

@davidfowl , this PR is to temporary add AI with sending ~5 telemetry items per second (unlike #361 where sample rate is 0) to measure AI's impact if it's considered "part of the platform". I would expect some heavy overhead on Benchmarks because currently App is doing mostly nothing and AI code will essentially become app's logic code. I would not think we need to merge this PR if we can measure perf off this PR commit, but you guys may think otherwise. Let me know if you need some more tuning in PR.

Dmitry-Matveev:
@sebastienros , please let me know if something else is needed in this PR to allow a perf run off this PR with AI enabled with some actual IKey. I haven't signed CLA yet to make sure it's not accidentally merged unless ASP.NET Core team would think it's the right thing to do.

PR #598, opened 2019-01-10
Fix breaking change introduced to MVC apps for 3.0

davidfowl:
- Added the new Newtonsoft JSON formatter package as a dependency in 3.0
- Matched the version of the Libuv package
- Fixed the code to call the new API in the netcoreapp3.0 case

cc @pranavkm

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -41,7 +41,7 @@
     <!-- The Npqsql driver for EF Core 3.0 doesn't exist yet -->
     <MicrosoftEntityFrameworkCoreSqlServerVersion30>2.2.0</MicrosoftEntityFrameworkCoreSqlServerVersion30>
     <MicrosoftAspNetCoreLibuvPackageVersion30>3.0.0-preview-19059-0308</MicrosoftAspNetCoreLibuvPackageVersion30>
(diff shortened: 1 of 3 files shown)


build/dependencies.props
```diff
@@ -40,7 +40,8 @@
 
     <!-- The Npqsql driver for EF Core 3.0 doesn't exist yet -->
     <MicrosoftEntityFrameworkCoreSqlServerVersion30>2.2.0</MicrosoftEntityFrameworkCoreSqlServerVersion30>
-    <MicrosoftAspNetCoreLibuvPackageVersion30>3.0.0-preview-18577-0035</MicrosoftAspNetCoreLibuvPackageVersion30>
+    <MicrosoftAspNetCoreLibuvPackageVersion30>3.0.0-preview-19059-0308</MicrosoftAspNetCoreLibuvPackageVersion30>
+    <MicrosoftAspNetCoreMvcNewtonsoftJson30> 3.0.0-preview-19059-0308</MicrosoftAspNetCoreMvcNewtonsoftJson30>
```
pranavkm: Extra space

sebastienros: We could use `$BenchmarksAspNetCoreVersion` as packages are released together. If fallback to a static version if it's not defined. I can do the change on the PR if you are ok with it.

Same for Libuv.

davidfowl:
LGTM

PR #614, opened 2019-02-04
Use Pipelines on HttpContext

benaadams:
Guessing it would be a regression until https://github.com/aspnet/AspNetCore/pull/7110 is merged?

/cc @davidfowl

### DIFF
--- a/src/Benchmarks/Middleware/PlaintextMiddleware.cs
+++ b/src/Benchmarks/Middleware/PlaintextMiddleware.cs
@@ -32,6 +32,19 @@ namespace Benchmarks.Middleware
             return _next(httpContext);
         }
 
+#if NETCOREAPP3_0
+        public async static Task WriteResponse(HttpResponse response)
+        {
+            var payloadLength = _helloWorldPayload.Length;
+            response.StatusCode = 200;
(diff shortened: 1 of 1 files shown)


benaadams:
Guessing it would be a regression until https://github.com/aspnet/AspNetCore/pull/7110 is merged?

src/Benchmarks/Middleware/PlaintextMiddleware.cs
```diff
@@ -32,6 +32,19 @@ public Task Invoke(HttpContext httpContext)
...
 
+#if NETCOREAPP3_0
+        public async static Task WriteResponse(HttpResponse response)
+        {
+            var payloadLength = _helloWorldPayload.Length;
+            response.StatusCode = 200;
+            response.ContentType = "text/plain";
+            response.ContentLength = payloadLength;
+
+            var pipe = response.BodyPipe;
```
davidfowl: Why not call WriteAsync on PipeWriter?

benaadams: Good idea :)

benaadams:
Not needed

PR #655, opened 2019-02-16
Use async IO (async ish)

davidfowl:
- The extra state machine might regress performance slightly but we recently removed the ability to do sync IO by default on request/response body streams.
- This PR does a proper async flush on the StreamWriter since the write is tiny should always complete synchronously.


davidfowl:
@sebastienros I'm going to merge this to unblock the perf CI, if it regresses performance too much then we can revisit the extra state machine.

PR #660, opened 2019-02-17
Made some fixes to the benchmarks UI

davidfowl:
- Caught any errors that can be thrown in the async void code path
- Added more scenarios to the JobDefinition
- Remove WeatherForecastService and pages
- Fix ETW trace collection


davidfowl:
I'm going to work on updating the UI to razor components

PR #772, opened 2019-03-20
Fix exception when running with a baseline without a description

roji:
The first run was executed with `--save baseline` but without `--description`, so the resulting baseline.bench.json has an empty Description node.

The second run was executed with `-diff baseline` and with a `--description` (required when using `-diff`, but not when using `--save`). The run failed with:

```
[09:02:45.507] System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at BenchmarksDriver.Program.Run(Uri serverUri, Uri clientUri, String sqlConnectionString, ServerJob serverJob, String session, String description, Int32 iterations, Int32 exclude, String shutdownEndpoint, TimeSpan span, List`1 downloadFiles, Boolean fetch, String fetchDestination, Boolean collectR2RLog, String traceDestination, CommandOption outputFileOption
, CommandOption sourceOption, CommandOption scriptFileOption, CommandOption markdownOption, CommandOption writeToFileOption, Nullable`1 requiredOperatingSystem, CommandOption saveOption, CommandOption diffOption) in /home/roji/projects/Benchmarks/src/BenchmarksDriver/Program.cs:line 1672                                                                          
```

This PR changes `BuildFields()` to always generate a description, even if it's null or empty. A better fix may be more appropriate but this does unblock me.

### DIFF
--- a/src/BenchmarksDriver/Program.cs
+++ b/src/BenchmarksDriver/Program.cs
@@ -1951,23 +1951,18 @@ namespace BenchmarksDriver
             }
         }
 
-        private static List<KeyValuePair<string, string>> BuildFields(Statistics average)
-        {
-            var fields = new List<KeyValuePair<string, string>>();
-            if (!String.IsNullOrEmpty(average.Description))
+        private static List<KeyValuePair<string, string>> BuildFields(Statistics average) 
+            => new List<KeyValuePair<string, string>>
             {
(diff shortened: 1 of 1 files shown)


sebastienros:
The intent was to use the save arg as the description if not defined. Will review when not of phone.

Issue #825, opened 2019-04-02
Add support for request body in drivers/clients


JamesNK:
I want to benchmark HTTP/2 when POSTing data.

**Test scenario 1 - JSON**
Version: HTTP/2
Method: POST
Headers: content-type: application/json
Body: UTF8 JSON data

**Test scenario 2 - gRPC**
Version: HTTP/2
Method: POST
Headers: content-type: application/grpc
Body: binary gRPC data

h2load is how the ASP.NET Core benchmarks are currently testing HTTP/2 traffic.

I have talked about this with @sebastienros and our setup doesn't support sending data in the body.

I'd like a new client property that allows a benchmark to set a `RequestBody`. This is base64 encoded content, so it can support binary messages that gRPC would send. The content in `RequestBody` would be sent with the request via h2load.

PR #826, opened 2019-04-02
Add request body support to h2load

JamesNK:
Fixes https://github.com/aspnet/Benchmarks/issues/825

### DIFF
--- a/src/BenchmarksClient/Workers/H2LoadWorker.cs
+++ b/src/BenchmarksClient/Workers/H2LoadWorker.cs
@@ -21,6 +21,8 @@ namespace BenchmarksClient.Workers
 
         private ClientJob _job;
         private Process _process;
+        private string _requestBodyTempFile;
+        private int? _requestBodyLength;
 
         public string JobLogText { get; set; }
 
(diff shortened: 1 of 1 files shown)


JamesNK:
Note: completely untested!

Will add some tests that use this to the gRPC repo.

sebastienros:
Note you can use @responsefile if the body is big enough, so you can externalize it from the command line

JamesNK:
Part of https://github.com/grpc/grpc-dotnet/issues/6

src/BenchmarksClient/Workers/H2LoadWorker.cs
```diff
@@ -51,14 +58,26 @@ private void InitializeJob()
         public async Task StartJobAsync(ClientJob job)
         {
             _job = job;
+
+            if (job.ClientProperties.TryGetValue("RequestBody", out var requestBodyText))
```
sebastienros: I would just put this block in `StartProcess`

You get bonus points if you handle it on wrk, wrk2 and bombardier ;)

PR #903, opened 2019-04-19
Proxy use HeaderNames rather than string const

benaadams:
It be faster after https://github.com/aspnet/AspNetCore/pull/9341

/cc @sebastienros


sebastienros:
Will that really be faster?

benaadams:
Will be a `ReferenceEquals` test rather than a `string.Equals(..., OrdinalIgnoreCase)`

benaadams:
Might not be huge, but every little helps :)

sebastienros:
Meh!

PR #913, opened 2019-04-19
Add a benchmark for endpoint plaintext

pranavkm:


pranavkm:
```
[04:05:42.329] Scenario EndpointPlaintext completed on benchmark client
RequestsPerSecond:           1,253,059

[04:08:18.888] Scenario Plaintext completed on benchmark client
RequestsPerSecond:           2,085,460
```

src/Benchmarks/Startup.cs
```diff
@@ -319,6 +325,18 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
                 app.UseMvc();
             }
 #elif NETCOREAPP3_0
+            if (Scenarios.Any("EndpointPlaintext"))
+            {
+                app.UseRouting();
+
+                app.UseEndpoints(endpoints =>
+                {
+                    endpoints.MapGet(
```
pranavkm: Does this look correct?

JamesNK: How close should this be to the other plain text benchmark?

The other benchmark writes text bytes:
https://github.com/aspnet/Benchmarks/blob/aa253347ad9a41365446d5706e81c7eb166e61d9/src/Benchmarks/Middleware/PlaintextMiddleware.cs#L37-L41

Routing wise, this is doing a MapGet which the other benchmark doesn't care about the HTTP verb. Also "/ep/plaintext" has slightly more routing overhead compared to "/ep-plaintext"

pranavkm: Updated to make it more like PlainTextMiddleware: 

| Description |       RPS | CPU (%) | Memory (MB) | Avg. Latency (ms) | Startup (ms) | First Request (ms) | Latency (ms) |
| ----------- | --------- | ------- | ----------- | ----------------- | ------------ | ------------------ | ------------ |
|             | 2,192,976 |      98 |          67 |              1.23 |          273 |              95.38 |         0.67 |

PR #986, opened 2019-05-16
Refactoring the Update statement

sebastienros:
Creating a PR for awareness as it's a significant change. Will create a PR to TE later.

It's reacting to this issue and how the requirements have been cleared up to allow such queries.
https://github.com/TechEmpower/FrameworkBenchmarks/issues/4773

### DIFF
--- a/src/Benchmarks/Data/BatchUpdateString.cs
+++ b/src/Benchmarks/Data/BatchUpdateString.cs
@@ -10,48 +10,33 @@ namespace Benchmarks.Data
     internal class BatchUpdateString
     {
         private const int MaxBatch = 500;
+
         private static string[] _queries = new string[MaxBatch];
 
-        public static IList<BatchUpdateString> Strings { get; } =
-            Enumerable.Range(0, MaxBatch)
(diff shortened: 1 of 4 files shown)


sebastienros:
/cc @benaadams @davidfowl @roji 

+20% at least

src/Benchmarks/Data/BatchUpdateString.cs
```diff
@@ -10,48 +10,33 @@ namespace Benchmarks.Data
...
         private static string[] _queries = new string[MaxBatch];
 
-        public static IList<BatchUpdateString> Strings { get; } =
-            Enumerable.Range(0, MaxBatch)
-                      .Select(i => new BatchUpdateString
-                      {
-                          Id = $"Id_{i}",
-                          Random = $"Random_{i}",
-                          BatchSize = i
-                      }).ToArray();
```
bgrainger: Is there still value in keeping this, to string-interpolate `$"Id_{i}"` just once for each _i_, instead of once per query?

PR #1054, opened 2019-06-01
React to breaking changes

davidfowl:


benaadams:
Related to this change?

image

davidfowl:
Dunno yet, turns out we've been measuring older bits since we changed versioning format...

davidfowl:
This was the change that broke it https://github.com/aspnet/AspNetCore/commit/08ed1afbbc1d1e11115eafcf4bd2c771c00ffd90. So since this commit, we've been testing against the older ASP.NET Core version, this regression is likely a runtime regression.

PR #1156, opened 2019-06-25
Fix BenchmarkDotNet results formatting on Linux

adamsitnik:
Anytime I was trying to run some BDN benchmarks on Linux machine I was getting a strangely formatted output:

obraz

So I have fixed it:

obraz

/cc @sebastienros @roji

### DIFF
--- a/src/BenchmarksDriver/BenchmarkDotNetUtils.cs
+++ b/src/BenchmarksDriver/BenchmarkDotNetUtils.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Linq;
+
+namespace BenchmarksDriver
+{
+    internal static class BenchmarkDotNetUtils
+    {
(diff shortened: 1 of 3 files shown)


roji:
Strange - I run BDN on Linux all the time and the output has always been fine - maybe it's related to something specific in your output... But anyway great to see Linux support improving :)

adamsitnik:
@roji which OS and cmd are you using? I am using powershell from Windows 10.

I most probably hit some edge case bug with weird whitespaces handling, hence I just added `.Trim()` to the code and did not investigate it too much.

roji:
Ah, you're on Powershell on Linux? It may be related to that...

I work on xubuntu 19.04 using zsh in the standard xfce4-terminal. In any case makes sense to have a trim there...

sebastienros:
Let's see how it goes with it

PR #1169, opened 2019-06-28
React to the JSON changes

davidfowl:
- Updated packages
- Use new methods (Read/Write/ToString -> Serialize/Deserialize)

cc @pranavkm @BrennanConroy


src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -42,21 +42,20 @@ public async Task Invoke(HttpContext httpContext)
                 httpContext.Response.ContentType = "application/json";
                 httpContext.Response.ContentLength = _bufferSize;
 
+#if !NETCOREAPP3_0
```
BrennanConroy: Why?

rynowak: Looks like the IO feature is only needed without S.T.J

PR #1216, opened 2019-07-23
Fix version prefix support since 5.0

sebastienros:


sebastienros:
I didn't mention the specific issue for BAR on purpose, at least it's fixing the baselines benchmarks

PR #1223, opened 2019-07-26
Use up-to-date Npgsql.EntityFrameworkCore.PostgreSQL nightly

roji:
We've been pinned to running TechEmpower on preview5 because of the Npgsql provider. With the recent port to bring it up to date we can now run on near-master.

Note that we're using a nightly and not preview7 because we need EF compiled query support, but that has only been very recently merged.

/cc @ajcvickers @divega

### DIFF
--- a/src/Benchmarks/Data/EfDb.cs
+++ b/src/Benchmarks/Data/EfDb.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Benchmarks.Configuration;
 using Microsoft.EntityFrameworkCore;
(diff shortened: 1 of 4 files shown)


roji:
Note: the Npgsql nightly version has been pinned on myget to prevent breakage, but the idea is to attempt a more regular sync schedule from now on.

src/Benchmarks/Data/EfDb.cs
```diff
@@ -70,9 +71,13 @@ public async Task<World[]> LoadMultipleUpdatesRows(int count)
             return results;
         }
 
+#if NETCOREAPP2_1 || NETCOREAPP2_2
         private static readonly Func<ApplicationDbContext, AsyncEnumerable<Fortune>> _fortunesQuery
             = EF.CompileAsyncQuery((ApplicationDbContext context) => context.Fortune);
-
+#else
+        private static readonly Func<ApplicationDbContext, IAsyncEnumerable<Fortune>> _fortunesQuery
```
sebastienros: Funny, I was working on the PR for TechEmpower to move to preview7, and fought some time with this change trying to figure out what to do, and did exactly the same change with the new extension method.

roji: Yeah, compiled query wouldn't have worked on preview7 :)

PR #1236, opened 2019-07-30
Move SignalR app to benchmarks

BrennanConroy:
Still need to test this


davidfowl:
> Still need to test this

Time for PR checks?

BrennanConroy:
Tested it, fixed path and now it works.

BrennanConroy:
Added to sln, made it run locally, and changed to using endpoint routing

PR #1284, opened 2019-08-08
add a sample template app that can be reused by others

adamsitnik:
@sebastienros


adamsitnik:
@sebastienros I've added the missing docs with sample investigation. The PR is ready for review.

@danmosemsft this should make it much easier for the CoreFX Team members to use ASP.NET Perf lab. PTAL at the doc.

src/BenchmarksApps/Samples/Template/Controllers/StringComparerController.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Template.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    public class StringComparerController : ControllerBase
```
sebastienros: I think we should use a piece of middleware instead, of endpoint routing, to remove the mvc stack that could add a lot of noise to the profile. Endpoint routing might be the easiest way, so the dev would just edit the Startup file to edit the code to run.

sebastienros: It would look like this:

```csharp
routeBuilder.Map("InvariantCultureIgnoreCase/{count}", context =>
{
    var count = context.GetRouteValue("count");
    return context.Response.WriteAsync($"Hi! {count}");
});
```

src/BenchmarksApps/Samples/Template/Startup.cs
```diff
@@ -0,0 +1,41 @@
...
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseRouting();
+
+            app.UseEndpoints(endpoints =>
+            {
+                endpoints.MapControllers();
```
sebastienros: This is where the endpoints would be defined.

src/BenchmarksApps/Samples/Template/Template.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <FrameworkReference Update="Microsoft.AspNetCore.App" RuntimeFrameworkVersion="$(MicrosoftAspNetCoreAppPackageVersion)" />
```
sebastienros: Should it be conditioned in order to work locally? Or are these values already defined by the sdk?

src/BenchmarksApps/Samples/Template/Startup.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            Configuration = configuration;
+        }
+
+        public IConfiguration Configuration { get; }
+
+        // This method gets called by the runtime. Use this method to add services to the container.
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddControllers()
```
sebastienros: This is unnecessary with endpoint routing

adamsitnik:
@sebastienros I've updated the PR,  PTAL. And you were right, middleware has much smaller overhead. 👍

src/BenchmarksApps/Samples/Template/Template.csproj
adamsitnik: I am not sure. From CoreFX team member perspective I don't want to run it locally, only using the ASP.NET Perf lab

sebastienros: Until someone wants to run it locally in order to debug what is not working on the server. Conditioning it is easy, I can commit the change to your PR if your are ok with that.

src/BenchmarksApps/Samples/Template/Program.cs
```diff
@@ -0,0 +1,17 @@
...
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            CreateWebHostBuilder(args).Build().Run();
+        }
+
+        public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
+            WebHost.CreateDefaultBuilder(args)
+                .UseStartup<Startup>();
```
sebastienros: Are you ok if I disable logging here also, such that it becomes opt-in (and documented) as logging adds some significant allocations (AsyncLocal to create scopes).

sebastienros: The idea is that there is the least noise to the actual benchmarks.

adamsitnik: another good idea, I've disabled the logging. The RPS went from 200k to 230k ;)

src/BenchmarksApps/Samples/Template/Template.csproj
adamsitnik: > I can commit the change to your PR if your are ok with that.

I think that it would be the best option since I simply don't know what exactly to change.

PR #1333, opened 2019-08-19
Use Npgsql EF Core 3.0.0-preview8

roji:
Also update EF microbenchmark SDK to preview8


roji:
Note that this depends on Npgsql 4.1.0-preview1, which probably still contains the perf regression @sebastienros identified in the nullability work. I hope to concentrate on that very soon.

/cc @ajcvickers

PR #1340, opened 2019-08-23
Fix Npgsql perf regression

roji:
By depending on 4.1.0-ci.2164


benaadams:
@roji Should TE be updated to pick this up? https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/CSharp/aspnetcore/PlatformBenchmarks/PlatformBenchmarks.csproj

PR #1374, opened 2019-09-10
Add small-single-exe scenarios

rynowak:


sebastienros:
Copy the `scenarios-frameworks.sh` file to `scenarios-singleexe.sh`, and replace the jobs with yours. Add the execute attribute in git.
Then in the command line at the bottom remove `--self-contained`.

rynowak:
updated

PR #1417, opened 2019-09-20
Add timeout to gRPC worker

JamesNK:


src/BenchmarksClient/Workers/GrpcWorker.cs
```diff
@@ -197,7 +199,13 @@ public async Task StartJobAsync(ClientJob job)
             cts.Token.WaitHandle.WaitOne();
 
             Log($"Waiting for call tasks to complete");
-            await Task.WhenAll(callTasks);
+
+            // Ensure calls never cause worker to hang with a timeout
+            var timeoutTask = Task.Delay(duration * 5);
+            if (timeoutTask == await Task.WhenAny(timeoutTask, Task.WhenAll(callTasks)))
+            {
+                _job.Error += Environment.NewLine + "Calls did not complete in a timely manner.";
```
JamesNK: If there is any error text will there be any notification?

sebastienros: Yes, but the job needs to end with the Completed state still.

PR #1424, opened 2019-10-02
Adding custom tags in bot

sebastienros:
/cc @halter73 @jkotalik @anurse @BrennanConroy


src/BenchmarksBot/Tags.cs
```diff
@@ -0,0 +1,99 @@
...
+            {"Plaintext-FastHttp", new Tags { Labels = new [] { "" } } },
+            {"Plaintext-NodeJs", new Tags { Labels = new [] { "" } } },
+            {"PlaintextNonPipelined", new Tags { Labels = new [] { "area-servers" } } },
+            {"PlaintextNonPipelinedLogging", new Tags { Labels = new [] { "" } } },
+            {"PlaintextNonPipelinedLoggingNoScopes", new Tags { Labels = new [] { "" } } },
+            {"PlaintextPlatform", new Tags { Labels = new [] { "" } } },
+            {"ResponseCachingPlaintextCached", new Tags { Labels = new [] { "area-middleware" } } },
+            {"ResponseCachingPlaintextResponseNoCache", new Tags { Labels = new [] { "area-middleware" } } },
+            {"ResponseCachingPlaintextRequestNoCache", new Tags { Labels = new [] { "area-middleware" } } },
+            {"ResponseCachingPlaintextVaryByCached", new Tags { Labels = new [] { "area-middleware" } } },
```
BrennanConroy:
```suggestion
            {"ResponseCachingPlaintextVaryByCached", new Tags { Labels = new [] { "area-middleware" } } },
            {"SignalRBroadcast", new Tags { Labels = new [] { "area-signalr" } } },
            {"SignalREcho", new Tags { Labels = new [] { "area-signalr" } } },
            {"SignalREchoAll", new Tags { Labels = new [] { "area-signalr" } } },
```

src/BenchmarksBot/Tags.cs
```diff
@@ -0,0 +1,99 @@
...
+            {"DbSingleQueryEf", new Tags { Labels = new [] { "" } } },
+            {"DbSingleQueryRaw", new Tags { Labels = new [] { "" } } },
+            {"EndpointPlaintext", new Tags { Labels = new [] { "area-mvc" } } },
+            {"FortunesPostgreSql-Actix", new Tags { Labels = new [] { "" } } },
+            {"FortunesPostgreSql-FastHttp", new Tags { Labels = new [] { "" } } },
+            {"FortunesPostgreSql-NodeJs", new Tags { Labels = new [] { "" } } },
+            {"HttpClient", new Tags { Labels = new [] { "area-httpclientfactory" } } },
+            {"HttpClientFactory", new Tags { Labels = new [] { "area-httpclientfactory" } } },
+            {"HttpClientParallel", new Tags { Labels = new [] { "area-httpclientfactory" } } },
+            {"Json", new Tags { Labels = new [] { "" } } },
```
anurse: We should probably assign *some* kind of label for all the issues, otherwise they'll end up in the routing queue and some human will have to route them anyway. Maybe we need an `area-perf` for other things and we cover that in the perf sync?

sebastienros: Ok with adding `area-perf` as a default, though it wouldn't add more value than the new default label we are also adding with this pr (`perf`, `perf-xyz-reason`).

I will start with what is there, and now it will be easy to just change what is created.

I am also thinking of adding more properties to this metadata class, like which script name is used in the jenkins so we can generate direct links, or anything else we would like to see in each issue related to a scenario.

anurse: The only value it would add is that the routing logic (MikLabel) literally checks for labels starting `area-`.

PR #1425, opened 2019-10-04
Add Crossgen2 scenario to benchmarks

nattress:
Add the aggressively linked, IBC-tuned, version bubbled crossgen2
compilation scenario to the benchmark lab runs.


nattress:
@rynowak PTAL

src/Benchmarks/link-a-thon.api-template.json
```diff
@@ -74,5 +74,8 @@
     "Source": {
       "Project": "src/GrpcService/GrpcService.csproj"
     }
+  },
+  "LinkAThonTrimmedAndR2RSingleFileWithTrimListCrossgen2": {
```
rynowak: I'll update the zip and try this out before merging.

sebastienros: Try to create names that are more concise. Like `LinkAThonTrimmedR2RSingleCrossgen2`

But technically it doesn't matter so it's just a question of preference.

sebastienros:
Should add the scenario in the script too

PR #1430, opened 2019-10-29
Add PR benchmarking infrastructure

halter73:
TODO:
- Use the Generic Host, appsettings.json, etc...
- Comment when starting benchmark run, and edit later.

Part of https://github.com/aspnet/AspNetCore/issues/14867


src/PRJobProducer/Program.cs
```diff
@@ -0,0 +1,359 @@
...
+                    }
+                }
+
+                return 0;
+            });
+
+            return app.Execute(args);
+        }
+
+        private static async IAsyncEnumerable<PullRequest> GetPRsToBenchmark(GitHubClient client, string botLoginName)
```
davidfowl: Fancy

src/JobConsumer/Program.cs
```diff
@@ -0,0 +1,316 @@
...
+                    BaselineStderr = errorBuilder.ToString(),
+                };
+            }
+
+            var baselineStdout = outputBuilder.ToString();
+            var baselinseStderr = errorBuilder.ToString();
+            outputBuilder.Clear();
+            errorBuilder.Clear();
+
+            Process.Start("git", $"checkout {buildRules.PullRequestSHA}").WaitForExit();
```
BrennanConroy: nit: GitHub merges the base branch and the PR branch when building, we might want to do something similar.

src/JobConsumer/Program.cs
```diff
@@ -0,0 +1,316 @@
...
+                        .OrderByDescending(f => f.LastWriteTime)
+                        .FirstOrDefault();
+
+
+                    // If no file was found, wait some time
+                    if (nextFile is null)
+                    {
+                        if (Console.KeyAvailable)
+                        {
+                            if (Console.ReadKey().Key == ConsoleKey.Enter)
```
davidfowl: Debugging?

src/JobConsumer/Program.cs
```diff
@@ -0,0 +1,316 @@
...
+                        }
+
+                        await Task.Delay(1000);
+                        continue;
+                    }
+
+                    Console.WriteLine($"Found '{nextFile.Name}'");
+
+                    // Attempting to move the file to the processing folder in order to lock it
+                    var session = nextFile.Name.Split('.', 2)[0];
```
davidfowl: This code needs to be rock solid, so it should be extremely defensive with lots of logs.

src/PRJobProducer/Program.cs
```diff
@@ -0,0 +1,359 @@
...
+                    var jsonDictionary = await JsonSerializer.DeserializeAsync<Dictionary<string, object>>(newJobStream);
+
+                    jsonDictionary["BuildInstructions"] = new BuildInstructions
+                    {
+                        BuildCommands = BuildCommands,
+                        BaselineSHA = pr.Base.Sha,
+                        PullRequestSHA = pr.Head.Sha,
+                    };
+
+                    newJobStream.SetLength(0);
```
davidfowl: Is this needed?

src/JobConsumer/Program.cs
halter73: I'm going to move both console apps to use Generic Host, but for now this was a quick and dirty alternative to a proper ctrl+c handler.

src/JobConsumer/Program.cs
halter73: The way I see it, it's better to let an exception bubble up and crash the process and let a watchdog observe and report the failure. If there's a bug in the file-moving logic, we cannot reasonably report it back to the PRJobProducer anyway since doing so would rely on the said file-moving logic, having the correct session id, etc...

Not to mention, long term I would love to replace all of this with something like SSH, so I don't want to spend a ton of time over-engineering this.

src/PRJobProducer/Program.cs
halter73: If the new contents of the file are shorter than the old contents, yes. Before adding whitespace to the serialized output, I would see the truncated end of the original file after the serialized output if I didn't call SetLength(0). Because I cannot guarantee that the base job file doesn't have a ton of extra whitespace or something that could make it larger than the new job file, I kept it.

src/JobConsumer/Program.cs
davidfowl: > so I don't want to spend a ton of time over-engineering this.

Like adding an if statement?

src/PRJobProducer/Program.cs
davidfowl: Add a comment? I wouldn't have guessed that.

PR #1431, opened 2019-10-29
Allow Blazor runs to be serialized correctly

pranavkm:


src/BenchmarksDriver/Program.cs
```diff
@@ -1192,8 +1192,13 @@ CommandOption diffOption
             ClientJob[] clientJobs = null;
 
             var serializer = WorkerFactory.CreateResultSerializer(_clientJob);
+            if (serializer is null)
+            {
+                Log($"No serializer found for {_clientJob.Client}. Using {nameof(WrkSerializer)} by default.");
+                serializer = new WrkSerializer();
```
sebastienros: ok

PR #1435, opened 2019-11-07
Support submitting PR benchmarks using Azure SDK

halter73:


sebastienros:
For your testing you might want to add `SdkVersion: "5.0.100-alpha1-015673"` in the server job until the sdk is fixed

src/PRJobProducer/IFileSystem.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.IO;
+using System.Threading.Tasks;
+
+namespace PRJobProducer
+{
+    public interface IFileSystem
+    {
+        Task CreateDirectoryIfNotExists(string destination);
+        Task<bool> FileExists(string location);
+        Task<Stream> ReadFile(string source);
```
davidfowl: nit: I’d have OpenRead and OpenWrite that return a stream you can read from and write to

src/PRJobProducer/Program.cs
```diff
@@ -29,94 +33,110 @@ public class Program
...
+
+                if (jobsPath.HasValue())
+                {
+                    JobFileSystem = new LocalFileSystem(jobsPath.Value());
+                }
+                else if (azureStorageConnectionString.HasValue())
+                {
+                    if (!azureStorageFileShareName.HasValue())
+                    {
+                        Console.WriteLine("--azure-storage-connection-string was provided with no --azure-storage-file-share.");
```
BrennanConroy: nit: this check could be more robust by checking the existence of both instead of one and then the other.

src/PRJobProducer/IFileSystem.cs
halter73: If I did that, I think I'd have to implement my own Stream type for AzureStorageFileSystem.OpenWrite(location). The Stream could be a pretty thin wrapper around MemoryStream that simply uploads the content on dispose, but I don't think it's worthwhile for an internal abstraction.

src/PRJobProducer/Program.cs
halter73: I added an else to cover this case.

PR #1439, opened 2019-12-04
Strong name Benchmarks.dll so it can use Kestrel's HttpParser temporarily

halter73:
See https://github.com/aspnet/AspNetCore/pull/17571 which will make Kestrel.Core's internals visible to the Benchmarks project, but for that to work, the Benchmarks assembly needs to be strong named as far as I'm aware.


src/Benchmarks/Benchmarks.csproj
```diff
@@ -19,6 +19,7 @@
     <!-- Prevent the SDK from validating the supported TFM. -->
     <NETCoreAppMaximumVersion>99.9</NETCoreAppMaximumVersion>
     <NoWarn>0618;NU1605</NoWarn>
+    <AssemblyOriginatorKeyFile>AspNetCore.snk</AssemblyOriginatorKeyFile>
```
halter73: REVIEW: This change doesn't seem to cause `dotnet build` to strong name the assembly, but I created the PR anyway to encourage someone to show me the correct way to do this without porting this project to arcade.

halter73:
Closing in favor of https://github.com/aspnet/AspNetCore/pull/17573

PR #1442, opened 2020-01-17
Add golang grpc benchmarks

JamesNK:


src/BenchmarksApps/Grpc/GoServer/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:alpine AS build-env
```
JamesNK: I added the go benchmark code here because it doesn't make sense to add them to our usual grpc-dotnet repo.

PR #1454, opened 2020-02-18
Add h2 benchmarks to golang gRPC

JamesNK:


JamesNK:
@sebastienros If benchmarking has exploded when you get back then this PR is probably the cause 😋

PR #1464, opened 2020-03-19
ProcessRequest is not async method

adamsitnik:
while working on https://github.com/dotnet/runtime/issues/33669 I've realized that `ProcessRequestAsync` is not really an async method, so we can remove this one extra `await`

It gives an extra +20-40k for Plaintext RPS

@sebastienros PTAL

### DIFF
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
@@ -60,7 +60,7 @@ namespace PlatformBenchmarks
 
                     if (_state == State.Body)
                     {
-                        await ProcessRequestAsync();
+                        ProcessRequest();
 
                         _state = State.StartLine;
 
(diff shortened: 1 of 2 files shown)


sebastienros:
We'll let it run and then submit the change to TE.

benaadams:
> We'll let it run and then submit the change to TE.

Trickier because TE does have async methods e.g DB stuff (benchmarks isn't in line with TE code) https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/CSharp/aspnetcore/PlatformBenchmarks/BenchmarkApplication.cs#L106-L140

sebastienros:
Should we refactor TE's code for this non-async path?
And probably sync the two apps so we are testing what's on TE.

benaadams:
> Should we refactor TE's code for this non-async path?

It returns a Task.Completed so could be changed to

```diff
  if (_state == State.Body)
  {
-     await ProcessRequestAsync();
+     var reqTask = ProcessRequestAsync();
+     if (!reqTask.IsCompletedSuccessfully)
+     {
+         await reqTask;
+     }

      _state = State.StartLine;
```

Or split the app into two with `#define`s?

> And probably sync the two apps so we are testing what's on TE.

Probably :)

sebastienros:
I can try this change here, while syncing the app. thanks.

benaadams:
Added sync https://github.com/TechEmpower/FrameworkBenchmarks/pull/5579

and sync here to include the db async stuff https://github.com/aspnet/Benchmarks/pull/1472

PR #1465, opened 2020-03-21
Sync EFCore.PG to 5.0.0-preview3-ci.20200321T191415

roji:
/cc @sebastienros @ajcvickers 

Test passed:

```
dotnet run -- --server http://asp-perf-lin:5001 --client http://asp-perf-load:5002 -j ../Benchmarks/benchmarks.html.json -n DbFortunesEf --database PostgresQL --duration 15 -i 1 --display-output --display-error -r http://github.com/roji/Benchmarks -b EFSync

RequestsPerSecond:           71,849
Max CPU (%):                 97
WorkingSet (MB):             397
Avg. Latency (ms):           3.66
Startup (ms):                372
First Request (ms):          883.06
Latency (ms):                1.39
Total Requests:              1,083,800
Duration: (ms)               15,080
Socket Errors:               0
Bad Responses:               0
Build Time (ms):             11,503
Published Size (KB):         26,126
SDK:                         5.0.100-preview.2.20120.3
Runtime:                     5.0.0-preview.3.20169.13
ASP.NET Core:                5.0.0-preview.3.20171.1
```

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -33,7 +33,7 @@
     <NpgsqlEntityFrameworkCorePostgreSQLVersion22>2.2.4</NpgsqlEntityFrameworkCorePostgreSQLVersion22>
     <NpgsqlEntityFrameworkCorePostgreSQLVersion30>3.0.0</NpgsqlEntityFrameworkCorePostgreSQLVersion30>
     <NpgsqlEntityFrameworkCorePostgreSQLVersion31>3.1.1</NpgsqlEntityFrameworkCorePostgreSQLVersion31>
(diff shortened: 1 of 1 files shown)


sebastienros:
Good initiative to merge without my approval since you have tested a benchmark targeting this branch. Maybe we could even imagine a bot could do this automatically of it could rub a benchmark. Or maybe I could add a CI job on PRs to run one. Then darc could handle that. Or an auto merge bot.

roji:
Yep, we could definitely do that. I don't really mind manually testing once a week like this, but this should also improve the general stability of the project because it would catch other kinds of bugs too. On the downside, if the benchmark servers aren't 100% reliable this would be flaky (I'm also thinking of cases where they have many queued jobs and the CI times out because of this).

PR #1466, opened 2020-03-24
Minor platform benchmarks improvements

adamsitnik:
It gives around +20k RPS for both JSON and PlainText

@sebastienros PTAL

### DIFF
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.cs
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.cs
@@ -102,7 +102,7 @@ namespace PlatformBenchmarks
 
         private static void Json(PipeWriter pipeWriter)
         {
-            var writer = GetWriter(pipeWriter);
+            var writer = GetWriter(pipeWriter, 160);
 
             // HTTP 1.1 OK
             writer.Write(_http11OK);
@@ -134,7 +134,7 @@ namespace PlatformBenchmarks
 
         private static void Default(PipeWriter pipeWriter)
         {
-            var writer = GetWriter(pipeWriter);
+            var writer = GetWriter(pipeWriter, 100);
 
             // HTTP 1.1 OK
             writer.Write(_http11OK);
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.cs
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.cs
@@ -74,7 +74,7 @@ namespace PlatformBenchmarks
 
         private static void PlainText(PipeWriter pipeWriter)
         {
-            var writer = GetWriter(pipeWriter);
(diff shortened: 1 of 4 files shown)


src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.cs
```diff
@@ -116,20 +118,23 @@ private static void Json(PipeWriter pipeWriter)
...
+            writer.WriteNumeric((uint)JsonPayload.Length);
 
             // End of headers
             writer.Write(_eoh);
+            writer.Commit();
 
             // Body
-            writer.Write(jsonPayload);
-            writer.Commit();
+            using (var utf8jsonWriter = new Utf8JsonWriter(writer.Output))
```
adamsitnik: the main difference for using `Utf8JsonWriter` is that we don't allocate an array of bytes

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.cs
```diff
@@ -27,6 +27,8 @@ public partial class BenchmarkApplication
         private readonly static AsciiString _plainTextBody = "Hello, World!";
 
         private static readonly JsonSerializerOptions SerializerOptions = new JsonSerializerOptions();
+        private static readonly JsonMessage _jsonMessage = new JsonMessage { message = "Hello, World!" };
```
benaadams: Doesn't meet the requirements https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#requirements
```
i. For each request, an object mapping the key message to Hello, World! must be instantiated.
```

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BufferExtensions.cs
```diff
@@ -26,34 +26,32 @@ internal static unsafe void WriteNumeric<T>(ref this BufferWriter<T> buffer, uin
             const byte AsciiDigitStart = (byte)'0';
 
             var span = buffer.Span;
-            var bytesLeftInBlock = span.Length;
 
             // Fast path, try copying to the available memory directly
             var advanceBy = 0;
-            fixed (byte* output = span)
+            if (span.Length >= 3)
```
adamsitnik: there is no need to pin this span, with size hint we always execute the "happy path" with size big enough and no boundaries check when indexing

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.cs
adamsitnik: thanks for the catch, I am going to fix it right away

benaadams:
Has changed a little, but I think this is now included in https://github.com/aspnet/Benchmarks/pull/1473 and https://github.com/TechEmpower/FrameworkBenchmarks/pull/5579; if you want to double check?

adamsitnik:
I am going to send a new PR

PR #1470, opened 2020-03-30
create writer once per request, reduce the number of calls to pipe.GetSpan and pipe.Advance

adamsitnik:
+200k RPS for Plaintext (from 8.2kk to 8.5kk) and no regression for JsonPlatform ;)

the gain is from:

* creating the writer once which calls `pipe.GetSpan` which might acquire a lock 
* delaying the commit which calls `pipe.Advance` which always takes a lock

@davidfowl

### DIFF
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
@@ -41,41 +41,50 @@ namespace PlatformBenchmarks
         {
             while (true)
             {
-                var task = Reader.ReadAsync();
-
-                if (!task.IsCompleted)
+                var readResult = await Reader.ReadAsync();
+                
+                if (!HandleRequest(readResult))
                 {
-                    // No more data in the input
-                    await OnReadCompletedAsync();
+                    return;
                 }
 
-                var result = await task;
-                var buffer = result.Buffer;
-                while (true)
-                {
-                    if (!ParseHttpRequest(ref buffer, result.IsCompleted, out var examined))
-                    {
-                        return;
-                    }
-
-                    if (_state == State.Body)
-                    {
-                        ProcessRequest();
-
-                        _state = State.StartLine;
-
-                        if (!buffer.IsEmpty)
-                        {
-                            // More input data to parse
-                            continue;
-                        }
-                    }
-
-                    // No more input or incomplete data, Advance the Reader
-                    Reader.AdvanceTo(buffer.Start, examined);
-                    break;
-                }
+                await Writer.FlushAsync();
             }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private bool HandleRequest(in ReadResult result)
+        {
+            var buffer = result.Buffer;
+            var writer = GetWriter(Writer);
+
+            while (true)
+            {
+                if (!ParseHttpRequest(ref buffer, result.IsCompleted, out var examined))
+                {
+                    return false;
+                }
+
+                if (_state == State.Body)
+                {
+                    ProcessRequest(ref writer);
+
+                    _state = State.StartLine;
+
+                    if (!buffer.IsEmpty)
+                    {
+                        // More input data to parse
+                        continue;
+                    }
+                }
+
+                // No more input or incomplete data, Advance the Reader
+                Reader.AdvanceTo(buffer.Start, examined);
+                break;
+            }
+
+            writer.Commit();
+            return true;
         }
 
         private bool ParseHttpRequest(ref ReadOnlySequence<byte> buffer, bool isCompleted, out SequencePosition examined)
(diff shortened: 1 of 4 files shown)


src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
```diff
@@ -41,41 +41,50 @@ private async Task ProcessRequestsAsync()
...
-                while (true)
+                await Writer.FlushAsync();
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private bool HandleRequest(in ReadResult result)
+        {
+            var buffer = result.Buffer;
+            var writer = GetWriter(Writer);
```
halter73: ParseHttpRequest can (at least theoretically) return true without switching `_state` to `State.Body`. If you move this inside `if (_state == State.Body)` it might save on some unneeded initialization.

Also, maybe I'm missing something, but wasn't the BufferWriter initialized once per request/response before this change?

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
```diff
@@ -41,41 +41,50 @@ private async Task ProcessRequestsAsync()
...
             {
-                var task = Reader.ReadAsync();
-
-                if (!task.IsCompleted)
+                var readResult = await Reader.ReadAsync();
+                
+                if (!HandleRequest(readResult))
                 {
-                    // No more data in the input
-                    await OnReadCompletedAsync();
```
halter73: Is there any concern that removing this logic this will hurt pipelined performance? Now there's at least once call to Writer.FlushAsync() per read.

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
adamsitnik: > wasn't the BufferWriter initialized once per request/response before this change

It was for a non-pipelined scenario (JSON). For pipelined scenario (Plaintext only) it was executed 16 times (at least to my understanding)

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
adamsitnik: > Is there any concern that removing this logic this will hurt pipelined performance?

I've run both JSON and Plaintext benchmarks and it won't.

> Now there's at least one call to Writer.FlushAsync() per read.

So far there was also one but in a less obvious way: when there was no input data ready, we were flushing the output. 

With this change we always do it in explicit way: read request, handle it and write the response.

benaadams: > So far there was also one but in a less obvious way: when there was no input data ready, we were flushing the output.

That was to batch the 16 requests in the smallest number of TCP packets? Flushing on each request would fragment the response across many packets?

adamsitnik: Please excuse me for the delayed answer, I had to switch to Linux and attach a debugger to check the exact request sizes.

The pipelining is transparent to the web server.  

When it's disabled, we receive 49 bytes from `Reader.ReadAsync`.
When it's enabled, we receive 49 * 16 bytes from `Reader.ReadAsync`

The point is that so far, we were calling `GetWriter` and `GetSpan` 16 times for Plaintext in the body loop, now we do it just once, hence the 200k RPS gain.

When it comes to flushing, the behaviour is the same, but the code is different. So far we were flushing when there was no input data available, which was when server has sent a request, but did not receive the response yet. With my change we do it in more explicit way. As I said, I've run both Plaintext and Json platform benchmarks and they did not regress.

benaadams: Ah because `HandleRequest` processes multiple?

adamsitnik: yes, exactly

adamsitnik:
@sebastienros could you please merge it?

adamsitnik:
@halter73 @benaadams thanks for the review!

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BufferWriter.cs
```diff
@@ -17,7 +17,7 @@ public BufferWriter(T output)
         {
             _buffered = 0;
             _output = output;
-            _span = output.GetSpan();
+            _span = output.GetSpan(sizeHint: 16 * 160);
```
davidfowl, line 20: clever

PR #1471, opened 2020-04-01
Port SignalR Client to work with new driver

BrennanConroy:


src/SignalRClient/BenchmarksEventSource.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System.Diagnostics.Tracing;
+
+namespace SignalRClient
+{
+    internal sealed class BenchmarksEventSource : EventSource
```
davidfowl, line 5: @sebastienros does everyone end up copying this around?

src/SignalRClient/Program.cs
```diff
@@ -0,0 +1,468 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Net.Http;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using McMaster.Extensions.CommandLineUtils;
```
davidfowl, line 11: We should migrate to System.CommandLine but it's not important in this PR

src/SignalRClient/BenchmarksEventSource.cs
sebastienros: Just Brennan, initiation ritual. Will put it in a class library at least.

davidfowl:
This looks really clean. The benchmarking infrastructure has come a long way. If only we had some global tools and some guides written up so others could easily consume it 😉

src/BenchmarksDriver2/benchmarks.signalr.yml
```diff
@@ -0,0 +1,46 @@
+﻿# Examples:
+# --config benchmarks.signalr.yml --scenario broadcast --profile aspnet-physical
```
sebastienros, line 2: wrong copy-paste, this file doesn't have 'profiles'

src/BenchmarksApps/SignalR/README.md
```diff
@@ -0,0 +1,17 @@
...
+The SignalRWorker that runs against this server is located at https://github.com/aspnet/benchmarks/blob/master/src/BenchmarksClient/Workers/SignalRWorker.cs.
+
+## Usage
+
+1. Push changes you would like to test to a branch on GitHub
+2. Clone aspnet/benchmarks repo to your machine or install the global BenchmarksDriver tool https://www.nuget.org/packages/BenchmarksDriver/
+3. If cloned go to the BenchmarksDriver project
+4. Use the following command as a guideline for running a test using your changes
+
+`benchmarks --server <server-endpoint> --client <client-endpoint> -p TransportType=WebSockets -p HubProtocol=messagepack -j https://raw.githubusercontent.com/aspnet/SignalR/dev/benchmarks/BenchmarkServer/signalr.json`
```
sebastienros, line 15: wrong

BrennanConroy: What is the command line?
I have `--config benchmarks.signalr.yml --scenario broadcast --variable serverUri=http://10.0.0.102` so far

PR #1473, opened 2020-04-03
Using smaller routes

sebastienros:
Plaintext
Before: 8,455,689 
After: 8,527,662 

3 runs

### DIFF
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/benchmarks.plaintext.json
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/benchmarks.plaintext.json
@@ -14,14 +14,14 @@
     "Port": 8080
   },
   "PlaintextPlatform": {
-    "Path": "/plaintext"
+    "Path": "/p"
   },
   "PlaintextPlatform-Mono": {
-    "Path": "/plaintext",
+    "Path": "/p",
     "UseMonoRuntime":  true
   },
   "PlaintextNonPipelinedPlatform": {
-    "Path": "/plaintext",
+    "Path": "/p",
     "ClientProperties": {
       "ScriptName": "",
       "PipelineDepth": 0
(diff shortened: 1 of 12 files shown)


src/BenchmarksApps/Kestrel/PlatformBenchmarks/benchmarks.plaintext.json
```diff
@@ -14,14 +14,14 @@
     "Port": 8080
   },
   "PlaintextPlatform": {
-    "Path": "/plaintext"
+    "Path": "/p"
```
rynowak, line 17: lol how is this not cheating?

sebastienros: Exactly. I filed an issue on TE to clarify the requirements.

sebastienros:
With preamble:
Json: 942,135
Plaintext: 8,758,381

benaadams:
This will now be a sync with TE https://github.com/TechEmpower/FrameworkBenchmarks/pull/5579 [merged]

src/BenchmarksApps/Kestrel/PlatformBenchmarks/benchmarks.plaintext.json
benaadams: Not invalid for current round, may be for next https://github.com/TechEmpower/FrameworkBenchmarks/issues/5582#issuecomment-609895369

davidfowl:
I don't like this change... Feels wrong.

sebastienros:
@davidfowl is it the new route or the preamble that you don't like ?

davidfowl:
😢

benaadams:
> 😢

Baselines `aspnet/Benchmarks` so any improvements to code quality or performance are measured from where we are rather than from a very diverged old fork.

davidfowl:
The /p and /j are BS improvements though. It doesn't feel right and it's something that can be easily rejected by TE if they decided to change that rule.

benaadams:
> The /p and /j are BS improvements though

Completely agreed; and @sebastienros has raised it https://github.com/TechEmpower/FrameworkBenchmarks/issues/5582; it is valid for this round however and pv, actrix etc are currently doing it, it is likely not be valid for next round:

> @sebastienros I did merge this. I'd like to see the impact as well. After discussing internally, we're going to make the routes required after this round. I'll make a change to the rules here shortly for round 20 and beyond.

However the current round is being measured soon so seems wrong to drop down the pack due to a technicality? 🤷‍♂️

PR #1480, opened 2020-04-09
Add new optional SocketPipe transport

adamsitnik:
This PR adds a simplified transport that is still using the abstraction of pipes, but it has it's own simplified logic that does not take any locks and does not use any extra tasks

For JSON Platform we jump from 920k to 1065k, in a separate PR I am going to send some improvements that take it to 1100k

@davidfowl could you please take a look?


src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/SocketPipeWriter.cs
```diff
@@ -0,0 +1,66 @@
...
+            _socket = socket;
+            _array = new byte[BufferSize];
+            _offset = 0;
+        }
+
+        public override void Advance(int bytes)
+        {
+            _offset += bytes;
+
+            if (_offset >= _array.Length)
```
davidfowl, line 30: How does this happen? This would be a bug right? You can't Advance more than we gave you in GetMemory

src/BenchmarksApps/Kestrel/PlatformBenchmarks/Program.cs
```diff
@@ -62,7 +62,7 @@ public static IWebHost BuildWebHost(string[] args)
 
             var host = new WebHostBuilder()
                 .UseBenchmarksConfiguration(config)
-                .UseKestrel((context, options) =>
+                .ConfigureKestrel((context, options) =>
```
davidfowl, line 65: Why this change?

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkConfigurationHelpers.cs
```diff
@@ -19,26 +25,42 @@ public static IWebHostBuilder UseBenchmarksConfiguration(this IWebHostBuilder bu
...
-                    options.ApplicationSchedulingMode = PipeScheduler.Inline;
-                });
+                    builder.UseSockets(options =>
+                    {
+                        if (int.TryParse(builder.GetSetting("threadCount"), out int threadCount))
+                        {
+                            options.IOQueueCount = threadCount;
+                        }
+#if NETCOREAPP5_0
+                    typeof(SocketTransportOptions).GetProperty("WaitForDataBeforeAllocatingBuffer")?.SetValue(options, false);
```
davidfowl, line 51: @sebastienros @halter73 What's the deal with this? Don't we have the bits yet?

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkConfigurationHelpers.cs
```diff
@@ -19,26 +25,42 @@ public static IWebHostBuilder UseBenchmarksConfiguration(this IWebHostBuilder bu
...
+            if (string.Equals(webHost, "SocketPipe", StringComparison.OrdinalIgnoreCase))
             {
-                builder.UseSockets(options =>
+                builder.ConfigureServices(services =>
                 {
-                    if (int.TryParse(builder.GetSetting("threadCount"), out int threadCount))
-                    {
-                       options.IOQueueCount = threadCount;
-                    }
+                    services.AddSingleton<IConnectionListenerFactory, SocketPipeTransportFactory>();
```
davidfowl, line 34: This should be the only thing you need to add, the other 2 types aren't required.

src/BenchmarksApps/Kestrel/PlatformBenchmarks/Program.cs
davidfowl: Revert this, I saw the other change.

src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/KestrelServerOptionsSetup.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.Extensions.Options;
+
+namespace PlatformBenchmarks
+{
+    // copy of https://github.com/dotnet/aspnetcore/blob/master/src/Servers/Kestrel/Core/src/Internal/KestrelServerOptionsSetup.cs
+    internal sealed class KestrelServerOptionsSetup : IConfigureOptions<KestrelServerOptions>
```
davidfowl, line 8: Remove this.

src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/SocketAwaitableEventArgs.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Net.Sockets;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace PlatformBenchmarks
+{
+    // more or less a copy of https://github.com/dotnet/aspnetcore/blob/master/src/Servers/Kestrel/Transport.Sockets/src/Internal/SocketAwaitableEventArgs.cs
+    // but without PipeScheduler (always inlining)
+    internal sealed class SocketAwaitableEventArgs : SocketAsyncEventArgs, ICriticalNotifyCompletion
```
davidfowl, line 11: Do you need this at all? You can use the SocketTask APIs that do this internally. Kestrel only does this because it's trying to schedule directly to the IOQueue from continuations.

src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/SocketPipeReader.cs
```diff
@@ -0,0 +1,75 @@
...
+            }
+            else
+            {
+                // in theory it's possible
+                Array.Resize(ref _array, _array.Length * 2);
+            }
+
+            var array = _array;
+            var args = _awaitableEventArgs;
+            args.SetBuffer(new Memory<byte>(array, _length, array.Length - _length));
```
davidfowl, line 61: You should be able to call _socket.ReadAsync here directly instead of using SocketAsyncEventArgs (see https://github.com/dotnet/runtime/blob/e22cf553e11d500c1523034f2c7ff745014e0629/src/libraries/System.Net.Sockets/src/System/Net/Sockets/Socket.Tasks.cs#L18).

src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/SocketPipeWriter.cs
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+
+        public override void CancelPendingFlush() { } // nop
+
+        public override void Complete(Exception exception = null) => _offset = 0;
+
+        public override ValueTask<FlushResult> FlushAsync(CancellationToken cancellationToken = default)
+        {
+            // we take advantage of the fact that all writes in TE are always small and non-blocking
```
davidfowl, line 42: Feels like cheating a little...

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkConfigurationHelpers.cs
sebastienros: The aspnet repos might not be up to date with the latest sdk. Alternatively we can force it in the driver command line.

halter73: This was merged a while ago. It should work with a recent enough `$(MicrosoftAspNetCoreAppPackageVersion)`. We probably need to look at updating the csproj or dependencies.props.

src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/SocketPipeReader.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            if (_socket.ReceiveAsync(args))
```
halter73, line 63: We shouldn't wait to receive more data if everything hasn't been examined yet. There may be no more data coming.

src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/SocketPipeReader.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                // ReceiveAsync "returns true if the I/O operation is pending"
+                // so we await only in that case (it's ugly but gives nice perf boost in JSON benchmark
+                await args;
+            }
+
+            _length += args.GetResult();
+
+            return new ReadResult(new System.Buffers.ReadOnlySequence<byte>(array, _offset, _length - _offset), isCanceled: false, isCompleted: true);
```
halter73, line 72: `isCompleted` should only be true if a FIN was received.

src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/SocketPipeWriter.cs
```diff
@@ -0,0 +1,66 @@
...
+                int bytesSent = _socket.Send(new ReadOnlySpan<byte>(_array, start, toSent), SocketFlags.None);
+
+                start += bytesSent;
+                toSent -= bytesSent;
+            }
+            while (toSent > 0);
+
+            _offset = 0;
+
+            return new ValueTask<FlushResult>(new FlushResult(isCanceled: false, isCompleted: true));
```
halter73, line 59: This `isCompleted` should also only be true if the socket is closed. In Kestrel, we expect PipeWriter.FlushAsync() to never throw. Our transports just log, trip a ConnectionClosed token and return completed FlushResults.

src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/SocketPipeWriter.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        public override Memory<byte> GetMemory(int sizeHint = 0) => new Memory<byte>(_array, _offset, _array.Length - _offset);
```
halter73, line 62: Here's where you'd want to resize the array based on the sizeHint.

halter73:
I know this isn't a "real" transport, so a lot of my comments are FYI. We can decide how important (or not) they are to address. If this is just to get a sort of upper bound on performance, this is probably fine once we get rid of the extraneous code.

src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/SocketPipeWriter.cs
adamsitnik: @davidfowl what if I make it conditional base on the size? Let's say everything below 1024 is sync, everything more is async? (maybe we could push this upstream to Kestrel as well)

src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/SocketPipeReader.cs
adamsitnik: I know, but it's slow. I am going to create an issue in runtime repo

src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/SocketAwaitableEventArgs.cs
adamsitnik: The cost of awaiting `socket.ReadAsync` is quite big, by using this custom class I was able  to get +15k RPS. I am going to create an issue in runtime repo

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkConfigurationHelpers.cs
sebastienros: I set `--sdk 5.0.100-preview.4.20202.8` in the baseline KPI runs.

Let me try to update the sdk in the aspnet repos, I did it last time.

src/BenchmarksApps/Kestrel/PlatformBenchmarks/SocketPipes/SocketAwaitableEventArgs.cs
davidfowl: cc @stephentoub

stephentoub: > The cost of awaiting socket.ReadAsync is quite big, by using this custom class I was able to get +15k RPS.

"quite big" is ~15K out of ~1065k in a benchmark that's basically seeing how fast it can call this method?  To me that's not worth all the extra code and complexity and diverging from how we want developers to use Socket.

> I am going to create an issue in runtime repo

Recommending what?  If you have concrete ideas for improving it, great, but it has to account for a bunch of things you're not here, like the possibility it might be used concurrently, being used via an interface via ValueTask, that there might be a custom SynchronizationContext/TaskScheduler to be used, etc.

davidfowl:
@adamsitnik time to close this?

PR #1481, opened 2020-04-09
More JSON improvements

adamsitnik:
this gives something around +15-20k RPS for JSON, with the main benefit being the usage of `Utf8JsonWriter`

@sebastienros PTAL


adamsitnik:
@davidfowl PTAL

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.Json.cs
```diff
@@ -7,6 +7,8 @@ namespace PlatformBenchmarks
 {
     public partial class BenchmarkApplication
     {
+        private readonly static uint _jsonPayloadSize = (uint)JsonSerializer.SerializeToUtf8Bytes(new JsonMessage { message = "Hello, World!" }, SerializerOptions).Length;
```
sebastienros, line 10: @benaadams do you know if that would be an issue? I don't think so.

benaadams: No, similar approach in https://github.com/aspnet/Benchmarks/pull/1473 copied from other frameworks; feels a bit of a cheat, but doesn't look to be ruled out

benaadams: More specifically actrix is doing this; but also combining all the headers into one static chunk (which  #1473 does)

benaadams: Note I copied some of the changes from @adamsitnik's previous PR https://github.com/aspnet/Benchmarks/pull/1466 for #1473

adamsitnik:
@sebastienros @davidfowl can this PR be merged?

adamsitnik:
@sebastienros thank you!

PR #1486, opened 2020-04-14
Add echo server with 3 basic implementations

davidfowl:
Getting something working so I can quickly test some ideas.

cc @halter73 @benaadams @tmds

### DIFF
--- a/src/BenchmarksApps/TcpEcho/Program.cs
+++ b/src/BenchmarksApps/TcpEcho/Program.cs
@@ -1,0 +1,82 @@
+﻿using System;
+using System.Net;
+using System.Net.Sockets;
+using System.Threading;
+using System.Threading.Tasks;
+using McMaster.Extensions.CommandLineUtils;
+
+namespace TcpEcho
+{
+    class Program
+    {
+        static async Task Main(string[] args)
+        {
+            var app = new CommandLineApplication();
+
+            app.HelpOption("-h|--help");
+            var optionIp = app.Option("-a|--address <IP>", "The server IP address", CommandOptionType.SingleValue);
+            var optionPort = app.Option("-p|--port <PORT>", "The server port. Default is 5201", CommandOptionType.SingleValue);
+            var optionBacklog = app.Option("-b|--backlog <BACKLOG>", "The TCP backlog. Default is 128.", CommandOptionType.SingleValue);
+            var optionType = app.Option("-t|--type <TYPE>", "The server implementation type. Value values are raw, socketpipe, socketduplexpipe.", CommandOptionType.SingleValue);
+
+            app.OnExecuteAsync(async cancellationToken =>
+            {
+                var ip = IPAddress.Parse(optionIp.Value() ?? "127.0.0.1");
+                var port = 8081;
+                var backlog = 128;
+
+                if (optionPort.HasValue())
+                {
+                    port = int.Parse(optionPort.Value());
+                }
+
+                if (optionBacklog.HasValue())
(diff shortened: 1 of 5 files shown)


src/BenchmarksApps/TcpEcho/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+                socket.Bind(endpoint);
+                socket.Listen(backlog);
+
+                using var reg = cancellationToken.Register(() => socket.Dispose());
+
+                try
+                {
+                    while (true)
+                    {
+                        var connection = await socket.AcceptAsync();
```
halter73, line 68: How are you driving load?

src/BenchmarksApps/TcpEcho/SocketConnection.cs
```diff
@@ -0,0 +1,41 @@
...
+            var buffer = new byte[4096];
+            while (true)
+            {
+                var read = await _socket.ReceiveAsync(buffer, SocketFlags.None);
+                if (read == 0)
+                {
+                    break;
+                }
+
+                await _socket.SendAsync(buffer, SocketFlags.None);
```
tmds, line 37: Slice buffer to `read`?

src/BenchmarksApps/TcpEcho/SocketDuplexPipeConnection.cs
```diff
@@ -0,0 +1,92 @@
...
+                {
+                    var buffer = input.Writer.GetMemory(2048);
+                    var read = await socket.ReceiveAsync(buffer, SocketFlags.None);
+
+                    if (read == 0)
+                    {
+                        break;
+                    }
+
+                    await input.Writer.FlushAsync();
```
tmds, line 41: same here, use `read`?

src/BenchmarksApps/TcpEcho/SocketPipeConnection.cs
```diff
@@ -0,0 +1,76 @@
...
+                {
+                    var buffer = pipe.Writer.GetMemory(2048);
+                    var read = await _socket.ReceiveAsync(buffer, SocketFlags.None);
+
+                    if (read == 0)
+                    {
+                        break;
+                    }
+
+                    await pipe.Writer.FlushAsync();
```
tmds, line 41: and here ...

src/BenchmarksApps/TcpEcho/SocketDuplexPipeConnection.cs
halter73: Good catch. Given how prescriptive the default Pipe implementation can be fore some things, it's surprising that it's OK with a missing call to PipeWriter.Advance(). I'm guessing nothing is actually being written to the socket.

src/BenchmarksApps/TcpEcho/Program.cs
davidfowl: The intent is to use this https://github.com/aspnet/Benchmarks/pull/1485.

src/BenchmarksApps/TcpEcho/SocketConnection.cs
davidfowl: oops 😄

src/BenchmarksApps/TcpEcho/SocketDuplexPipeConnection.cs
davidfowl: OOps

PR #1490, opened 2020-04-19
Plaintext, Json, eagerly create SequenceReader

benaadams:
Breaks 1.17M RPS on localhost for my 6 core machine running both client and server: 
wrk (WSL2) <-> Kestrel (Windows)
```
ben@Ben-Desktop:~/wrk$ wrk http://172.29.144.1/p -t 6 -c 256 -s ./scripts/pipeline.lua -- 16
Running 10s test @ http://172.29.144.1/p
  6 threads and 256 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     4.04ms    5.19ms 104.43ms   89.32%
    Req/Sec   197.43k    44.55k  576.00k    74.83%
  11838304 requests in 10.10s, 1.39GB read
Requests/sec: 1172409.14
Transfer/sec:    140.88MB
```

/cc @adamsitnik @sebastienros

### DIFF
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
@@ -52,32 +52,34 @@ namespace PlatformBenchmarks
 #if !DATABASE
         private async Task ProcessRequestsAsync()
         {
-            var reader = Reader;
-            var writer = Writer;
-            var token = CancellationToken.None;
-
             while (true)
             {
-                var readResult = await reader.ReadAsync(token);
+                var readResult = await Reader.ReadAsync(default);
+                var buffer = readResult.Buffer;
+                var isCompleted = readResult.IsCompleted;
 
-                if (!HandleRequest(readResult))
+                if (buffer.IsEmpty && isCompleted)
                 {
                     return;
                 }
 
-                await writer.FlushAsync(token);
+                if (!HandleRequests(buffer, isCompleted))
+                {
+                    return;
+                }
+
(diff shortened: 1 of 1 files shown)


src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
```diff
@@ -140,51 +146,35 @@ private async Task ProcessRequestsAsync()
...
                 {
-                    if (Parser.ParseRequestLine(new ParsingAdapter(this), buffer, out consumed, out examined))
-                    {
-                        state = State.Headers;
-                    }
+                    state = State.Headers;
 
-                    buffer = buffer.Slice(consumed);
+                    var parsedLength = unconsumedSequence.Slice(reader.Position, consumed).Length;
+                    reader.Advance(parsedLength);
```
benaadams, line 161: This is a bit horrible; but can't advance/reposition a `SequenceReader` with a `SequencePosition`.

However want to change `Parser.ParseRequestLine` to take a `SequenceReader` so it's only short term

benaadams:
Plaintext to 8,902,903

Matching new api pattern for https://github.com/dotnet/aspnetcore/pull/20885 though will require some more changes when that comes through (as it skips the calculation of `.Length` and additional `.Advance`)

src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
benaadams: Looks like there is an api proposal to fix this? https://github.com/dotnet/runtime/issues/30770

/cc @davidfowl

PR #1494, opened 2020-04-21
AzDo consumer

sebastienros:


src/AzDoConsumer/JobPayload.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public TimeSpan Timeout { get; set; } = DefaultJobTimeout;
+
+        public string[] Args { get; set; }
+
+        public string RawPayload { get; set; }
+
+        public static JobPayload Deserialize(byte[] data)
+        {
+            var str = Encoding.UTF8.GetString(data);
```
davidfowl, line 21: What else is in the payload?

sebastienros: DataContractSerializer custom namespace that breaks the json document

PR #1496, opened 2020-04-22
Use 5.0 for SignalR benchmarks

BrennanConroy:


src/SignalRClient/SignalRClient.csproj
```diff
@@ -2,13 +2,13 @@
...
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>netcoreapp5.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.SignalR.Protocols.MessagePack" Version="3.1.0" />
-    <PackageReference Include="Microsoft.AspNetCore.SignalR.Client" Version="3.1.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="3.1.0" />
+    <PackageReference Include="Microsoft.AspNetCore.SignalR.Protocols.MessagePack" Version="5.0.0-*" />
```
davidfowl, line 9: Do we do this in other benchmarks?

BrennanConroy: I think the majority of the other benchmarks have references to things in the shared framework.

davidfowl: EF?

sebastienros: The EF team updates the version manually once a week ... not sure if it's better than this. Could we report this version in the benchmarks results? We can't even assume it will match the aspnet runtime version as it could be set by the server while this one will look at the feeds.

src/SignalRClient/signalrclient.yml
```diff
@@ -5,9 +5,8 @@
       branchOrCommit: master
       project: src/SignalRClient/SignalRClient.csproj
     readyStateText: SignalR Client
-    sdkVersion: 3.1.201
-    aspNetCoreVersion: '3.1'
-    runtimeVersion: '3.1'
+    sdkVersion: 5.0.100-preview.2.20120.3
```
sebastienros, line 8: If it works with the latest sdk, and I don't see any reason why not, then you can omit this argument as by default it will use the latest 5.0.*

src/SignalRClient/SignalRClient.csproj
BrennanConroy: Should I be using the dependencies.props file?

davidfowl: Uhhhh, I think this is fine since this is the behavior I want.

BrennanConroy: > Could we report this version in the benchmarks results?

Any suggestions on how to do this?

sebastienros: https://github.com/aspnet/Benchmarks/blob/master/src/Benchmarks/Program.cs#L53-L66

src/SignalRClient/Program.cs
```diff
@@ -63,6 +64,20 @@ static async Task Main(string[] args)
             {
                 Console.WriteLine("SignalR Client");
 
+                Console.WriteLine("#StartJobStatistics");
```
sebastienros, line 67: Funny ;) There are two way to expose a metric, either using the event source, or using the console. I know I pointed you to a console example ... I was not clear. But you are already using the event source in the rest of the code, so i think you should do it there too, nicer.

```
BenchmarksEventSource.Log.Metadata("signalr/raw-errors", "all", "all", "Raw errors", "Raw errors", "object");
```

PR #1497, opened 2020-04-22
Update for safe startline parsing PR

benaadams:
Change for PR https://github.com/dotnet/aspnetcore/pull/20885 (unmerged)

/cc @halter73

### DIFF
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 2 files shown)


src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
```diff
@@ -310,15 +323,16 @@ public void OnHeader(ReadOnlySpan<byte> name, ReadOnlySpan<byte> value)
 
             public void OnHeadersComplete(bool endStream)
                 => RequestHandler.OnHeadersComplete(endStream);
+            public void OnStartLine(HttpVersionAndMethod versionAndMethod, TargetOffsetPathLength targetPath, Span<byte> startLine)
```
halter73, line 326: nit:

```suggestion

            public void OnStartLine(HttpVersionAndMethod versionAndMethod, TargetOffsetPathLength targetPath, Span<byte> startLine)
```

halter73:
@sebastienros The startline PR has been merged to master. Should we merge this now?

sebastienros:
I think it's ok as the change was in aspnet. But let me test it tomorrow first.

benaadams:
Currently Platform tests will break when aspnet is updated? (The regular ones should be fine as is internal)

PR #1507, opened 2020-05-08
Add YARP proxy benchmark

halter73:


src/BenchmarksDriver2/benchmarks.proxy.yml
```diff
@@ -12,6 +12,16 @@ variables:
     path: /
 
 jobs:
+  yarp:
+    source:
+      repository: https://github.com/microsoft/reverse-proxy.git
+      branchOrCommit: halter73/benchmark
```
halter73, line 18: This should be `master` once https://github.com/microsoft/reverse-proxy/pull/148 is merged.

sebastienros: we can change it once it's merged, and point to this branch for now.

sebastienros: Why not have a .yml file describing the app alongside the benchmarks app, and `import` it?

src/BenchmarksDriver2/benchmarks.proxy.yml
```diff
@@ -139,6 +163,7 @@ profiles:
       downstreamUri: http://10.0.0.103:5000
       downstreamAddress: 10.0.0.103
       downstreamPort: 5000
+      proxyUri: http://asp-perf-lin:5001
```
sebastienros, line 166: Can't you use `serverUri`? And also have the BenchmarksApp use the `serverPort` variable used as an argument?

src/BenchmarksDriver2/benchmarks.proxy.yml
```diff
@@ -161,6 +186,7 @@ profiles:
       downstreamUri: http://10.0.0.103:5000
       downstreamAddress: 10.0.0.103
       downstreamPort: 5000
+      proxyUri: http://asp-citrine-lin:5001
```
sebastienros, line 189: Same as before. Also use the 10.0.0.102 address as this is the ip on the private vlan for these machines, otherwise it it goes through NIC (1GB)

anurse:
Could the benchmark yaml live in Microsoft/reverse-proxy? I’d prefer it be all together (other than the AzDo goop to trigger it) so that people can see it and easily run benchmarks themselves.

sebastienros:
@anurse this file contains the scenarios we need for our CI. But i suggested to have the part that describes yarp in the same repos as the app. But there could be another yaml file in the repos with a simple scenario (or more scenarios).

anurse:
> @anurse this file contains the scenarios we need for our CI. But i suggested to have the part that describes yarp in the same repos as the app. But there could be another yaml file in the repos with a simple scenario (or more scenarios

I suppose that's reasonable. I think I'd still prefer that the proxy repo has all the proxy benchmarks defined within it, and the CI builds just reference it and fill in variables (if that's possible), but if that's not feasible (or if I'm the only one who wants that ;)) that's fine.

sebastienros:
Ah, didn't understand you'd want all the proxy scenarios defined in this repos. I don't mind.
Note that the last time we decide to put benchark apps somewhere, they were moved from /aspnetcore to /benchmarks, e.g. Platform, SignalR.

anurse:
My vote matters fairly little now, tbh, so I'll leave it up to @halter73 :). My suggestion was that having all the scenarios defined in the YARP repo would make it easier for YARP folks to see and use them locally (perhaps using the tool @davidfowl is cooking up) and to understand how the PowerBI charts and such are generated.

build/scenarios-proxy.sh
```diff
@@ -7,6 +7,7 @@ SESSION=`date '+%Y%m%d%H%M%S'`
 proxyJobs="--config https://raw.githubusercontent.com/aspnet/Benchmarks/master/src/BenchmarksDriver2/benchmarks.proxy.yml "
 
 jobs=(
+  "--scenario proxy-yarp        $proxyJobs --property proxy=yarp"
```
sebastienros, line 10: We are not using these files anymore, it's proxy-scenarios.yml

halter73: Should we delete these?

src/BenchmarksDriver2/Program.cs
```diff
@@ -126,7 +126,7 @@ public static int Main(string[] args)
             _profileOption = app.Option("--profile", "Profile name", CommandOptionType.MultipleValue);
             _outputOption = app.Option("-o|--output", "Output filename", CommandOptionType.SingleValue);
             _compareOption = app.Option("--compare", "An optional filename to compare the results to. Can be used multiple times.", CommandOptionType.MultipleValue);
-            _variableOption = app.Option("-v|--variable", "Variable", CommandOptionType.MultipleValue);
+            _variableOption = app.Option("--variable", "Variable", CommandOptionType.MultipleValue);
```
halter73, line 129: This was conflicting with "-v|--verbose".

PR #1509, opened 2020-05-09
Add scenarios for 2MB MVC + JSON data

pranavkm:
Precursor for https://github.com/dotnet/aspnetcore/issues/21245


build/scenarios.sh
```diff
@@ -61,6 +61,10 @@ jobs=(
   "-n MvcJsonNet2k --webHost KestrelSockets $trend $jsonJobs"
   "-n MvcJsonInput2k --webHost KestrelSockets $trend $jsonJobs"
   "-n MvcJsonNetInput2k --webHost KestrelSockets $trend $jsonJobs"
+  "-n MvcJsonInput2M --webHost KestrelSockets $trend $jsonJobs"
```
sebastienros, line 64: We are not using these files anymore, but trend-scenarios.yml

pranavkm:
🆙 📅

davidfowl:
Anyway to see the results before merging it? @sebastienros ?

sebastienros:
This is a change in the client service too, so it has to be deployed first, which is automatic at midnight. Which is also why it should be merged after 2pm such that the next automated run doesn't fail as the client doesn't have the scripts yet.

Will not be an issue when the new driver is used as the scenario will be able to upload its own wrk scripts.

PR #1517, opened 2020-05-16
Run YARP benchmarks with the latest runtime

halter73:


sebastienros:
Do you think you could also export the version metadata in your benchmark?

https://github.com/aspnet/Benchmarks/blob/master/src/BenchmarksApps/HttpClient/Proxy/Program.cs#L103-L122

PR #1519, opened 2020-05-19
try Suggestion from Stephen and use Write methods directly

adamsitnik:
@stephentoub +20k in JSON!

Before:

```log
RequestsPerSecond:           1,149,856
Max CPU (%):                 99
WorkingSet (MB):             407
Avg. Latency (ms):           1
Startup (ms):                196
First Request (ms):          31.96
Latency (ms):                0.12
Total Requests:              17,362,051
Duration: (ms)               15,100
Socket Errors:               0
Bad Responses:               0
Build Time (ms):             4,001
Published Size (KB):         102,238
SDK:                         5.0.100-preview.5.20264.2
Runtime:                     5.0.0-preview.6.20262.14
ASP.NET Core:                5.0.0-preview.5.20255.6
```

After:

```log
RequestsPerSecond:           1,171,304
Max CPU (%):                 100
WorkingSet (MB):             410
Avg. Latency (ms):           0.86
Startup (ms):                202
First Request (ms):          31.5
Latency (ms):                0.1
Total Requests:              17,686,603
Duration: (ms)               15,100
Socket Errors:               0
Bad Responses:               0
Build Time (ms):             4,001
Published Size (KB):         102,238
SDK:                         5.0.100-preview.5.20264.2
Runtime:                     5.0.0-preview.6.20262.14
```

With https://github.com/dotnet/runtime/pull/36371/

```log
RequestsPerSecond:           1,203,964
Max CPU (%):                 100
WorkingSet (MB):             408
Avg. Latency (ms):           0.77
Startup (ms):                198
First Request (ms):          36.87
Latency (ms):                0.11
Total Requests:              18,179,315
Duration: (ms)               15,100
Socket Errors:               0
Bad Responses:               0
Build Time (ms):             4,001
Published Size (KB):         102,238
SDK:                         5.0.100-preview.5.20264.2
Runtime:                     5.0.0-preview.6.20262.14
ASP.NET Core:                5.0.0-preview.5.20255.6
```

@karelz

### DIFF
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.Json.cs
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.Json.cs
@@ -27,7 +27,10 @@ namespace PlatformBenchmarks
             // Body
             using (Utf8JsonWriter utf8JsonWriter = new Utf8JsonWriter(writer.Output))
             {
-                JsonSerializer.Serialize<JsonMessage>(utf8JsonWriter, new JsonMessage { message = "Hello, World!" }, SerializerOptions);
+                var message = new JsonMessage { message = "Hello, World!" };
+                utf8JsonWriter.WriteStartObject();
+                utf8JsonWriter.WriteString("message", message.message);
+                utf8JsonWriter.WriteEndObject();
             }
         }
     }


stephentoub:
Does this violate the TE spec?
"A JSON serializer must be used to convert the object to JSON."

adamsitnik:
> "A JSON serializer must be used to convert the object to JSON."

adamsitnik:
> "A JSON serializer must be used to convert the object to JSON."

It does. 1200k looks tempting, but I am afraid I should close this issue.

benaadams:
OTOH if it could be reframed using C# source generators that would probably be ok

Haven't looked to deeply at source generators, but an example of a deserializater using generators and System.Text.Json https://github.com/TomaszRewak/C-sharp-stack-only-json-parser

benaadams:
Related issue from @jkotas https://github.com/dotnet/runtime/issues/1568; not sure if using source generators for System.Json is in the 5.0 timeframe?

jkotas:
Source generators for json serialization are not in .NET 5, unfortunately. I agree that source generators are the right way to get json serializers with best performance.

Note that the source generators should be able to do even better than the handwritten json writter code. They can pre-encode constant names to save even more cycles. The pre-encoding is not convenient to do by hand, but it is very simple to do in source generator. There are other tricks that we can consider for source generators, like having special methods on json writer that skip some of the validation that is important for hand-written code, but it is unnecessary for auto-generated code. 

cc @steveharter

PR #1523, opened 2020-05-24
Skip JSON validation in BenchmarkApplication.Json

stephentoub:
cc: @adamsitnik


benaadams:
Noticed we are boxing `writer.Output` https://github.com/aspnet/Benchmarks/pull/1524

PR #1524, opened 2020-05-24
Don't box struct writer in Json

benaadams:
`writer.Output` is a struct and boxes when given to `IBufferWriter<byte>`; so use the underlying writer instead.

### DIFF
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.Json.cs
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.Json.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Buffers;
 using System.Text.Json;
 
 namespace PlatformBenchmarks
@@ -19,7 +20,7 @@ namespace PlatformBenchmarks
         [ThreadStatic]
         private static Utf8JsonWriter t_writer;
 
-        private static void Json(ref BufferWriter<WriterAdapter> writer)
+        private static void Json(ref BufferWriter<WriterAdapter> writer, IBufferWriter<byte> bodyWriter)
         {
             writer.Write(_jsonPreamble);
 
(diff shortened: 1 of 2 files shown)


stephentoub:
Why does this WriterAdapter thing exist at all?  Doesn't PipeWriter implement IBufferWriter?

benaadams:
To specialize the generic `BufferWriter<T>` for direct calls rather than shared via interface calls

stephentoub:
Ok, why does `BufferWriter<T>` exist here? :)

benaadams:
Mainly for pipelined; otherwise the change would be something like
```diff
- private static void Json(ref BufferWriter<WriterAdapter> writer)
- {
-     writer.Write(_jsonPreamble);
-     writer.Write(DateHeader.HeaderBytes);
-     writer.Commit();
+ private static void Json(PipeWriter writer)
+ {
+     Span<byte> span = writer.GetSpan(_jsonHeadersLength);
+     _jsonPreamble.AsSpan().CopyTo(span);
+     span = span.Slice(_jsonPreamble.Length);
+     DateHeader.HeaderBytes.CopyTo(span);
+     writer.Advance(_jsonHeadersLength);
```

benaadams:
Mainly for pipelined; otherwise the change would be something like
```diff
- private static void Json(ref BufferWriter<WriterAdapter> writer)
- {
-     writer.Write(_jsonPreamble);
-     writer.Write(DateHeader.HeaderBytes);
-     writer.Commit();
+ private static void Json(PipeWriter writer)
+ {
+     Span<byte> span = writer.GetSpan(_jsonHeadersLength);
+     _jsonPreamble.AsSpan().CopyTo(span);
+     span = span.Slice(_jsonPreamble.Length);
+     DateHeader.HeaderBytes.CopyTo(span);
+     writer.Advance(_jsonHeadersLength);
```
Still has lots of length checks?

benaadams:
Would also need to fork Plaintext and Json with `#define`s as for pipelined it creates the `BufferWriter<T>` once for a batch, and we wouldn't want to pay that cost if we didn't use it? 🤷‍♂️

benaadams:
`BufferWriter<T>` also for the DB variants as we don't know lengths and need to ask it how much we wrote

benaadams:
@sebastienros can I tempt you for a merge? Issue @stephentoub picks up on would be a bigger refactor which should be a seperate PR

benaadams:
Aside the mixed mode `BufferWriter<T>` + `PipeWriter` in Json is also why it fails if you run it pipelined which had me confused for a while (as `BufferWriter<T>`'s cursor is bypassed)

PR #1533, opened 2020-06-01
Add UseMonoRuntime_LlvmJit option

fanyang-mono:
Add an option to support netcore Mono with LLVM JIT

### DIFF
--- a/src/Benchmarks.ServerJob/ServerJob.cs
+++ b/src/Benchmarks.ServerJob/ServerJob.cs
@@ -60,6 +60,7 @@ namespace Benchmarks.ServerJob
         public string RuntimeVersion { get; set; } = "";
         public string SdkVersion { get; set; } = "";
         public bool UseMonoRuntime { get; set; } = false;
+        public bool UseMonoRuntime_LlvmJit = false;
         public bool NoGlobalJson { get; set; }
         public Database Database { get; set; } = Database.None;
 
--- a/src/Benchmarks.ServerJob/ServerJob.cs
+++ b/src/Benchmarks.ServerJob/ServerJob.cs
@@ -59,8 +59,7 @@ namespace Benchmarks.ServerJob
         public string AspNetCoreVersion { get; set; } = "";
         public string RuntimeVersion { get; set; } = "";
         public string SdkVersion { get; set; } = "";
-        public bool UseMonoRuntime { get; set; } = false;
-        public bool UseMonoRuntime_LlvmJit = false;
+        public string UseMonoRuntime { get; set; } = "";
         public bool NoGlobalJson { get; set; }
         public Database Database { get; set; } = Database.None;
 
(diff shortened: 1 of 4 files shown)


src/Benchmarks.ServerJob/ServerJob.cs
```diff
@@ -60,6 +60,7 @@ public class ServerJob : IIdentifiable
         public string RuntimeVersion { get; set; } = "";
         public string SdkVersion { get; set; } = "";
         public bool UseMonoRuntime { get; set; } = false;
+        public bool UseMonoRuntime_LlvmJit = false;
```
sebastienros, line 63: Maybe merge these two fields in one `string` containing the type of jit to use. I assume there might be a third value at some point. The option can still have no value to be compatible with the existing command line argument.

src/BenchmarksDriver/Program.cs
```diff
@@ -747,6 +749,14 @@ public static int Main(string[] args)
                         Log("WARNING: '--self-contained' has been set implicitly as the mono runtime is used.");
                         Console.ResetColor();
                     }
+                    if (monoOption_LlvmJit.HasValue())
```
sebastienros, line 752: Just add an `||` with the previous statement. And if you merge both arguments then that works too.

src/Benchmarks.ServerJob/ServerJob.cs
fanyang-mono: Yes, there will be a third flavor, which is AOT LLVM. What do you mean by merging those two fields but the option could still have no value? If we merge them, then it has to have a value to differentiate which flavor the use wants to have, right?

fanyang-mono: I got it now...

src/BenchmarksServer/Startup.cs
```diff
@@ -3792,14 +3791,19 @@ private static void StartDotNetTrace(int processId, ServerJob job)
...
             {
-
-                var packageName = "Microsoft.NETCore.App.Runtime.Mono.linux-x64".ToLowerInvariant();
+                var packageName = "";
+                switch (mode) {
+                    case "jit":
+                        packageName = "Microsoft.NETCore.App.Runtime.Mono.linux-x64".ToLowerInvariant();
+                        break;
+                    case "llvm-jit":
+                        packageName = "Microsoft.NETCore.App.Runtime.Mono.LLVM.AOT.linux-x64".ToLowerInvariant();
```
SamMonoRT, line 3804: Is this right or should we use the LLVM.JIT package ? Seems confusing

fanyang-mono: We don't have the LLVM.JIT package. For some reason, they never got generated. The literal only difference between the AOT LLVM and JIT LLVM packages are the AOT packages also contain llc/opt. So it is totally fine to use the AOT LLVM package for JIT LLVM.

sebastienros:
LGTM. Have you tested the changes for the server?

fanyang-mono:
> LGTM. Have you tested the changes for the server?

Yes, I've tested this change on Mono lab machines.

PR #1547, opened 2020-06-09
DB Platform benchmarks microoptimizations

adamsitnik:
obraz

@roji PTAL

### DIFF
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/Data/RawDb.cs
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/Data/RawDb.cs
@@ -30,7 +30,8 @@ namespace PlatformBenchmarks
                     await db.OpenAsync();
                 }
 
-                using (var cmd = CreateReadCommand(db))
+                var (cmd, _) = CreateReadCommand(db);
+                using (cmd)
                 {
                     return await ReadSingleRow(cmd);
                 }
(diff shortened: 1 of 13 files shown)


src/BenchmarksApps/Kestrel/PlatformBenchmarks/Data/RawDb.cs
```diff
@@ -1,152 +1,134 @@
...
+            finally
             {
-                await rdr.ReadAsync();
-
-                return new World
-                {
-                    Id = rdr.GetInt32(0),
-                    RandomNumber = rdr.GetInt32(1)
-                };
+                db.Close();
```
roji, line 41: Did you see an actual perf difference between Dispose and Close on this?

src/BenchmarksApps/Kestrel/PlatformBenchmarks/Data/RawDb.cs
```diff
@@ -158,11 +140,39 @@ public async Task<List<Fortune>> LoadFortunesRows()
...
+        {
+            var cmd = new NpgsqlCommand("SELECT id, randomnumber FROM world WHERE id = @Id", connection);
+            var parameter = new NpgsqlParameter<int>(parameterName: "@Id", value: _random.Next(1, 10001));
+
+            cmd.Parameters.Add(parameter);
+
+            return (cmd, parameter);
+        }
+
+        private async Task<World> ReadSingleRow(NpgsqlCommand cmd)
```
roji, line 164: Consider inlining this to remove another async method

roji:
Interesting that multiplexing is slower in Updates and Multiple Queries, but I've already seen cases like this and have a pretty good idea on why - this may be a case of tweaking multiplexing parameters in the connection string. Maybe we can leave those specific scenarios without multiplexing for now, and I'll investigate this later? We'd need the ability to have different connection strings for different scenarios.

adamsitnik:
@roji I've re-run all the benchmarks because I realized that I was running the `updates` and `multiple queries` in the wrong way. Now everything is correct (512 connections, count parameter == 20) according to TechEmpower requirements

Machine | Benchmark | before | after | ratio
-- | -- | -- | -- | --
Citrine 28 cores | Fortunes | 330,403 | 341,120 | 3.24%
  | Fortunes Multiplexing | 406,303 | 420,724 | 3.55%
  | Updates | 17,223 | 17,271 | 0.28%
  | Updates Multiplexing | 16,457 | 17,597 | 6.93%
  | Single Query | 366,911 | 382,301 | 4.19%
  | Single Query Multiplexing | 411,250 | 434,686 | 5.70%
  | Multiple Queries | 37,857 | 39,878 | 5.34%
  | Multiple Queries Multiplexing | 23,192 | 24,551 | 5.86%

PR #1548, opened 2020-06-12
Add support for Mono LLVM AOT

fanyang-mono:


src/BenchmarksServer/Startup.cs
```diff
@@ -4082,6 +4087,107 @@ private static async Task UseMonoRuntimeAsync(string runtimeVersion, string outp
...
+
+            }
+            catch (Exception e)
+            {
+                Log.WriteLine("ERROR: Failed to AOT for mono. " + e.ToString());
+                throw;
+            }
+        }
+
+	private static string runBashCmd(string cmd, string wd)
```
sebastienros, line 4172: There is already a `ProcessUtil.RunAsync` that is used across this file to run commands. Search for `cgroup`, or `GetSwapBytes`

src/BenchmarksServer/Startup.cs
```diff
@@ -2566,6 +2566,11 @@ private static async Task<string> CloneRestoreAndBuild(string path, ServerJob jo
                 }
 		
                 await UseMonoRuntimeAsync(runtimeVersion, outputFolder, job.UseMonoRuntime);
+
+		if (job.UseMonoRuntime.Equals("llvm-aot"))
```
sebastienros, line 2570: You can call this method from `UseMonoRuntimeAsync` directly

src/BenchmarksServer/Startup.cs
```diff
@@ -4024,6 +4029,9 @@ private static async Task UseMonoRuntimeAsync(string runtimeVersion, string outp
                     case "llvm-jit":
                         packageName = "Microsoft.NETCore.App.Runtime.Mono.LLVM.AOT.linux-x64".ToLowerInvariant();
                         break;
+                    case "llvm-aot":
```
sebastienros, line 4032: The two cases can be merged

src/BenchmarksServer/Startup.cs
```diff
@@ -4082,6 +4087,107 @@ private static async Task UseMonoRuntimeAsync(string runtimeVersion, string outp
...
+                if (Directory.Exists(Path.GetDirectoryName(dotnetMonoPath)))
+                {
+                    Directory.Delete(Path.GetDirectoryName(dotnetMonoPath));
+                }
+                Directory.CreateDirectory(Path.GetDirectoryName(dotnetMonoPath));
+
+                Log.WriteLine($"Downloading dotnet skd package for mono AOT");
+
+                var found = false;
+                var url = $"https://dotnetcli.azureedge.net/dotnet/Sdk/{dotnetSdkVersion}/dotnet-sdk-{dotnetSdkVersion}-linux-x64.tar.gz";
```
sebastienros, line 4105: Isn't the sdk already downloaded and installed by a previous "standard" step?

src/BenchmarksServer/Startup.cs
```diff
@@ -4082,6 +4087,107 @@ private static async Task UseMonoRuntimeAsync(string runtimeVersion, string outp
...
+                }
+
+                Log.WriteLine($"Patching local dotnet with mono runtime and extracting llvm");
+
+                var packageName = "Microsoft.NETCore.App.Runtime.Mono.LLVM.AOT.linux-x64".ToLowerInvariant();
+                var runtimePath = Path.Combine(_rootTempDir, "RuntimePackages", $"{packageName}.{runtimeVersion}.nupkg");
+
+                var strCmdGetVer = "./dotnet --list-runtimes | grep -i \"Microsoft.NETCore.App\"";
+                var resultGetVer = runBashCmd(strCmdGetVer, Path.GetDirectoryName(dotnetMonoPath));
+                var MicrosoftNETCoreAppPackageVersion = resultGetVer.Split(' ')[1];
```
sebastienros, line 4129: You should respect the version that was requested by the driver (`runtimeVersion`)

src/BenchmarksServer/Startup.cs
fanyang-mono: I think this it makes more sense to call it here, because AOT4Mono is a post-process for the app publish folder.

src/BenchmarksServer/Startup.cs
fanyang-mono: Ideally, they are supposed to use different packages. But I guess that they could be merged, since this one package fit both scenarios.

src/BenchmarksServer/Startup.cs
sebastienros: So is the rest of `UseMonoRuntimeAsync` which already does things specific to `"llvm-aot"`. It's on the line right above, it's still post publication.

src/BenchmarksServer/Startup.cs
fanyang-mono: This is a local dotnet that I switch the runtime to mono, so that I could use it for AOT. I don't want to do that to the installed dotnet which is shared by CoreCLR and Mono.

src/BenchmarksServer/Startup.cs
fanyang-mono: This is for AOT only, so it doesn't matter much that this version is not the same as `runtimeVersion`. I get it because the folder named after this version.

src/BenchmarksServer/Startup.cs
fanyang-mono: Yes, I understand that it is post publication. UseMonoRuntimeAsync is a function to switch the runtime to using runtime in the publish folder. AOT4Mono is a function to pre-compile all assemblies in the publish folder. I don't see the first one could contain the second one. As they are parallel to each other. If you rename `UseMonoRuntimeAsync` to `PostProcess4Mono`. It could make sense to put both `UseMonoRuntimeAsync` and `AOT4Mono` inside. In that case, I will probably still create a separate function for `UseMonoRuntimeAsync` to make things more clear, according to single-responsibility principle.

PR #1553, opened 2020-06-17
Update platform benchmarks to Npgsql 5.0.0-alpha1

roji:
And start using multiplexing


roji:
@sebastienros merged this, hope to see some nice changes in the graph soon :)

PS I see that benchmarks.fortunes.yml here still uses Windows as the client with bombardier - for me this produces worse results than when I use load with wrk, should we switch?

PR #1554, opened 2020-06-17
More db platform benchmarks

adamsitnik:
Changes:

* added remaining platform db benchmarks and connection string per benchmark (multiplexing enabled only for fortunes and single query where it helps)
* switch from bombardier to wrk for better RPS and doing what TE is doing
* display app output to see the connection string that is used

@roji PTAL

### DIFF
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/benchmarks.fortunes.yml
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/benchmarks.fortunes.yml
@@ -26,20 +26,62 @@ scenarios:
   fortunesplatform:
     db:
       job: postgresql
-
     application:
       job: platformbenchmarks
       buildArguments: 
         - "/p:IsDatabase=true"
-
+      environmentVariables:
+        connectionstring: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;Enlist=false;Max Auto Prepare=4;Multiplexing=true;Write Coalescing Delay Us=500;Write Coalescing Buffer Threshold Bytes=1000
     load:
       job: bombardier
       variables:
         presetHeaders: html
         path: /fortunes
-        warmup: 5 
-        duration: 10
-        serverPort: 5000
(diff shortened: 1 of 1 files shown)


src/BenchmarksApps/Kestrel/PlatformBenchmarks/benchmarks.fortunes.yml
```diff
@@ -23,23 +23,65 @@ jobs:
     readyStateText: PostgreSQL init process complete
 
 scenarios:
-  fortunesplatform:
+  fortunes:
```
benaadams, line 26: Will changing the name cause issue for the results graphs? (i.e. history)

adamsitnik: @benaadams  that's a very good question

@sebastienros can I change this name? I just want to shorten it to make it easier to type the command

sebastienros: there is no risk to break anything because we are not recording the results of this file, but the old json ones. However it will once we start using the yml files. And we don't want to include a custom metadata like "platform=true", that wouldn't make sense, I'd rather see it's a platform benchmark in its name.

adamsitnik: @sebastienros but we know that it's a platform benchmark because the `.yml` file lives in the `PlatformBenchmarks` folder. Imho having "platform" would make more sense if we had one big `.yml` file with all the benchmarks. The same goes for "aspnet" in the profile name. I have access only to aspnet machines, so to me it's redundant

```cmd
--config ..\BenchmarksApps\Kestrel\PlatformBenchmarks\benchmarks.fortunes.yml --profile aspnet-citrine --scenario fortunesplatform
```

```cmd
--config ..\BenchmarksApps\Kestrel\PlatformBenchmarks\benchmarks.fortunes.yml --profile citrine --scenario fortunes
```

It's not super important to me, just a way to make the commands shorter and easier to memorize.

PR #1555, opened 2020-06-19
Checking NET5_0

sebastienros:
PR for visibility, since sdk preview-7 it seem the NET COREAPP5_0 directive is not used anymore.
Our benchmarks were using it. Checking both since we still use older sdks in some circumstances. 

@benaadams @adamsitnik @halter73


adamsitnik:
> since sdk preview-7 it seem the NET COREAPP5_0 directive is not used anymore

PR #1558, opened 2020-06-29
Get mono package based on computer architecture

fanyang-mono:


SamMonoRT:
https://github.com/dotnet/runtime/pull/38562 is needed for the arm64 aot-llvm nuget packages

PR #1562, opened 2020-07-01
Create yml file for mono scenarios

fanyang-mono:


scenarios/mono.benchmarks.yml
```diff
@@ -0,0 +1,234 @@
...
+      useMonoRuntime: jit
+    load:
+      job: wrk
+      variables:
+        presetHeaders: json
+        path: /j
+        serverPort: 5000
+        connections: 512
+
+  fortunes-mono-jit:
```
SamMonoRT, line 75: From the presentation in the .NET Town Hall, it is a goal to improve on Json, Fortunes and other benchmarks for CoreCLR. I believe Mono should do the same, and we should add new scenarios and reporting for all the one CoreCLR focusses heavily on.

scenarios/mono.benchmarks.yml
```diff
@@ -0,0 +1,299 @@
...
+      cores: 56
+    jobs: 
+      db:
+        endpoints: 
+          - http://asp-mono-load:5001
+      application:
+        endpoints: 
+          - http://asp-mono-lin:5001
+        variables:
+          databaseServer: 192.168.1.1
```
sebastienros, line 296: This is the ip of the db server from the perspective of the mono-lin one. Is it really 192.168.1.1 ?

fanyang-mono: 192.168.1.1 is the private connection between mono-lin and mono-load. Is that the one should be used for talk to the database as well, since for this configuration the mono-load is also the database server.

PR #1563, opened 2020-07-03
use Sockets as default transport for PlatformBenchmarks 

adamsitnik:
I was working on the last experiment for Sockets for 5.0 and I realized that Driver**2** does not pass sockets as a transport option by default (like the old driver does).

This explains why we were getting 1150k for JSON with the new driver and 1200k with the old one.

@sebastienros I think that we should just use Sockets by default. What do you think?

@roji this could be the source of some driver differences for other benchmarks as well


benaadams:
Corresponding TE https://github.com/TechEmpower/FrameworkBenchmarks/pull/5830

PR #1565, opened 2020-07-08
fix and tune the request parsing

adamsitnik:
@sebastienros  ;p


benaadams:
https://github.com/aspnet/Benchmarks/pull/1566 also has a change to return 404 on default route, also is less adventurous, maybe merge that an rebase this on it (if it improves)?

adamsitnik:
I've updated my branch

adamsitnik:
@sebastienros I think that it's still worth to merge this PR as it's unifying the 3.0 and 5.0 code (this time 5.0 had bug, 3.0 did not) and also improving the perf by using this dummy check of what is the second char in the buffer (every path has unique char[1])

PR #1571, opened 2020-07-16
Fix Mono LLVM AOT support for custom output files

fanyang-mono:
The first version of Mono LLVM AOT support didn't take custom output files into consideration. I made some adjustment in this PR to support that use case.

/cc @kouvel


kouvel:
LGTM, thanks!

PR #1579, opened 2020-08-05
Fixing wrong endpoint urls

Zoxive:
perf-win and citrine-amd had load and db endpoints backwards


sebastienros:
Thanks, but they don't. The intent is to be able to run some jobs in parallel, knowing that db and load machines are physically identical.

PR #1586, opened 2020-09-21
Use UnsafePreferInlineScheduling on Linux for Json Platform

BrennanConroy:
From 1.219m to 1.241m RPS

### DIFF
--- a/scenarios/platform.benchmarks.yml
+++ b/scenarios/platform.benchmarks.yml
@@ -39,6 +39,10 @@ scenarios:
   json:
     application:
       job: platformbenchmarks
+      buildArguments:
+        - "/p:IsJson=true"
+      environmentVariables:
+        DOTNET_SYSTEM_NET_SOCKETS_INLINE_COMPLETIONS: "{% if isLinux == true %}1{% else %}0{% endif %}"
     load:
       job: wrk
       variables:
@@ -134,6 +138,7 @@ profiles:
       serverPort: 5000
       serverAddress: 10.0.0.102
       cores: 28
+      isLinux: true
     jobs: 
       db:
         endpoints: 
@@ -206,6 +211,7 @@ profiles:
       serverPort: 5000
       serverAddress: 10.0.0.102
       cores: 12
+      isLinux: true
     jobs: 
       db:
         endpoints: 
--- a/scenarios/platform.benchmarks.yml
+++ b/scenarios/platform.benchmarks.yml
@@ -39,8 +39,6 @@ scenarios:
   json:
     application:
       job: platformbenchmarks
-      buildArguments:
-        - "/p:IsJson=true"
       environmentVariables:
         DOTNET_SYSTEM_NET_SOCKETS_INLINE_COMPLETIONS: "{% if isLinux == true %}1{% else %}0{% endif %}"
     load:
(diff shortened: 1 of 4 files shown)


src/BenchmarksApps/Kestrel/PlatformBenchmarks/Program.cs
```diff
@@ -79,8 +80,19 @@ public static IWebHost BuildWebHost(string[] args)
...
                     });
                 })
-                .UseStartup<Startup>()
-                .Build();
+                .UseStartup<Startup>();
+
+#if NET5_0
+            hostBuilder.UseSockets(options =>
+            {
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
benaadams, line 88: Aside, shouldn't this also have?
```csharp
options.WaitForDataBeforeAllocatingBuffer = false;
```

src/BenchmarksApps/Kestrel/PlatformBenchmarks/Program.cs
```diff
@@ -79,8 +80,20 @@ public static IWebHost BuildWebHost(string[] args)
...
+                {
+                    options.UnsafePreferInlineScheduling = Environment.GetEnvironmentVariable("DOTNET_SYSTEM_NET_SOCKETS_INLINE_COMPLETIONS") == "1";
+                    options.WaitForDataBeforeAllocatingBuffer = false;
```
benaadams, line 91: Doesn't need to be Linux for this one; does need to be 5.0 though

scenarios/platform.benchmarks.yml
```diff
@@ -39,6 +39,8 @@ scenarios:
   json:
     application:
       job: platformbenchmarks
+      environmentVariables:
+        DOTNET_SYSTEM_NET_SOCKETS_INLINE_COMPLETIONS: "{% if isLinux == true %}1{% else %}0{% endif %}"
```
halter73, line 43: Nit: Given we're doing the `RuntimeInformation.IsOSPlatform(OSPlatform.Linux)` before we consider setting `UnsafePreferInlineScheduling` anyway, this `isLinux` variable seems redundant.

```suggestion
        DOTNET_SYSTEM_NET_SOCKETS_INLINE_COMPLETIONS: 1
```

BrennanConroy: We don't want to set the environment variable on non-linux machines

halter73: Why not?

halter73 (approved):

scenarios/platform.benchmarks.yml
BrennanConroy: Because it makes perf worse. Remember the table we had before:
Sockets Default + Kestrel Default : 1,187,598 RPS
Sockets Inline + Kestrel Inline: 1,242,924 RPS
Sockets Inline + Kestrel Default: 1,072,542 RPS
Sockets Default + Kestrel Inline: 1,153,500 RPS

halter73: Right, but that was all on linux. Setting this environment variable wouldn't make a difference on Windows.

BrennanConroy: D'oh

sebastienros (approved):
What are the numbers with the new option?

BrennanConroy:
I've seen as high as 1.257m but it does seem a bit noisier with that option

PR #1587, opened 2020-09-22
Run some HTTPS proxy scenarios and enable HTTP/2

halter73:


build/proxies-scenarios.yml
```diff
@@ -50,20 +49,23 @@ parameters:
...
-        displayName: ${{ s.displayName }} ${{ payload.displayName }} ${{ protocol.displayName }}
-        inputs:
-          connectedServiceName: ${{ parameters.connection }}
-          waitForCompletion: true
-          messageBody: |
-            {
-              "name": "crank",
-              "args": [ "--table ProxyBenchmarks --sql SQL_CONNECTION_STRING --chart --session $(session) --profile ${{ parameters.profile }} --no-metadata --no-measurements --load.variables.warmup ${{ parameters.warmup }} --load.variables.duration ${{ parameters.duration }} ${{ s.arguments }} ${{ payload.arguments }} ${{ protocol.arguments }}" ]
-            }
+      - ${{ if or(eq(protocol.displayName, 'http'), or(eq(s.displayName, 'YARP'), eq(s.displayName, 'HttpClient'))) }}:
```
sebastienros, line 60: I would add a property on each scenario called `supportsServerHttps` and check for it here. Will be easier to add more scenarios without having to enumerate them here.

build/proxies-scenarios.yml
```diff
@@ -50,20 +49,23 @@ parameters:
   type: object
   default:
     - displayName: http
-      arguments: --variable protocol=http --property protocol=http
+      arguments: --variable serverScheme=http --variable downstreamScheme=http --property protocol=http
+    - displayName: https
+      arguments: --variable serverScheme=https --variable downstreamScheme=https --property protocol=https
```
sebastienros, line 54: Should we add a new property for `downstreamProtocol` as I assume we'll also do some benchmarks with http here

src/BenchmarksDriver2/benchmarks.proxy.yml
```diff
@@ -5,11 +5,12 @@
 # --config benchmarks.proxy.yml --scenario proxy-envoy --profile aspnet-physical
 
 imports:
-  - https://raw.githubusercontent.com/aspnet/Benchmarks/master/src/Wrk2Client/wrk2.yml
   - https://raw.githubusercontent.com/aspnet/Benchmarks/master/src/BombardierClient/bombardier.yml
 
 variables:
     path: /
+    serverScheme: http
```
sebastienros, line 12: There should be a file in /scenarios that is used now by the CI and documented. This file is obsolete. We can add a comment at the top to mention that.

build/proxies-scenarios.yml
halter73: I was considering that, I didn't want to create too many variations though. This would quadruple the variations instead of double them. Are we okay with that?

sebastienros: We don't have to run it now, and we'll maybe never do it, but saving the info is free. At least you supported it with the variables. We know that people will ask how much it cost to switch from http to https in the downstream.

sebastienros: And also this way I can generate a legend that makes it obvious that it's https all the way. Instead of hard coding in the charts.

build/proxies-scenarios.yml
```diff
@@ -46,24 +49,55 @@ parameters:
...
+- name: downstreamSchemes
   type: object
   default:
     - displayName: http
-      arguments: --variable protocol=http --property protocol=http
+      arguments: --variable downstreamScheme=http --property downstreamScheme=http
+    - displayName: https
+      arguments: --variable downstreamScheme=https --property downstreamScheme=https
+
+- name: serverHttpProtocols
```
sebastienros, line 68: We don't need two "sets" here, for server and downstream, but just one called protocols, that will define both server and downstream protocols, and just the combinations we want. Then add `requireServerHttp1` ... that we can check in the condition more easily with `supportServerHttp1`.

I can change it if you want

sebastienros: This way we just list and enumerate the combinations we want:
http - http
https - https
h2 - h2
https - http
h2 - http

halter73: That's how I had it when I first submitted the PR, but you wanted me to split protocols into upstream and downstream protocols. https://github.com/aspnet/Benchmarks/pull/1587#discussion_r492933409. I warned you this would lead to a combinatorial explosion. Did seeing how complicated this made the conditions make you change your mind?

sebastienros: No, I just wanted to save the two values in the properties. But here the loop can define the two values at the same time.

halter73: Also, h2 doesn't necessarily imply https. To fully define the scenario it would have to be something like `HTTP/1.1 with no TLS - HTTP/1.1 with no TLS`, `HTTP/2 with TLS - HTTP/2 with TLS`, etc...

But I think prior-knowledge HTTP/2 is rare enough that we can skip non-TLS HTTP/2 scenarios.

sebastienros: I realized there was a 3rd loop, right. So we can put everything in the same loop, and yes we don't want h2 without tls (h2c). I think "protocol" is sufficient, for both tiers, no need to mention it's with tls, and let's assume h2 is with it. worst case we can also use h2c to note it's without it (which will never happen). 
So http, https, h2, and define the combinations we measure with these.

PR #1588, opened 2020-09-24
Inline platform plaintext

BrennanConroy:
10.7m -> 11.7m


sebastienros:
What are the numbers for the other platform benchmarks?

BrennanConroy:
What other ones?

sebastienros:
How many scenarios does this file contain? 😁

BrennanConroy:
The db ones probably will not benefit. I tried Fortunes and it went down a couple dozen thousand RPS

PR #1589, opened 2020-09-27
Better batch update sql

benaadams:
CASE is faster than join to inline temp  table

### DIFF
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/Data/BatchUpdateString.cs
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/Data/BatchUpdateString.cs
@@ -11,8 +11,8 @@ namespace PlatformBenchmarks
 
         public static DatabaseServer DatabaseServer;
 
-        internal static readonly string[] Ids = Enumerable.Range(0, MaxBatch).Select(i => $"@Id_{i}").ToArray();
-        internal static readonly string[] Randoms = Enumerable.Range(0, MaxBatch).Select(i => $"@Random_{i}").ToArray();
+        internal static readonly string[] Ids = Enumerable.Range(0, MaxBatch).Select(i => $"@I{i}").ToArray();
+        internal static readonly string[] Randoms = Enumerable.Range(0, MaxBatch).Select(i => $"@R{i}").ToArray();
 
         private static string[] _queries = new string[MaxBatch + 1];
 
@@ -23,20 +23,22 @@ namespace PlatformBenchmarks
                 return _queries[batchSize];
             }
 
+            return CreateBatch(batchSize);
+        }
+
+        private static string CreateBatch(int batchSize)
+        {
             var lastIndex = batchSize - 1;
 
             var sb = StringBuilderCache.Acquire();
 
-            if (DatabaseServer == DatabaseServer.PostgreSql)
-            {
-                sb.Append("UPDATE world SET randomNumber = temp.randomNumber FROM (VALUES ");
-                Enumerable.Range(0, lastIndex).ToList().ForEach(i => sb.Append($"(@Id_{i}, @Random_{i}), "));
-                sb.Append($"(@Id_{lastIndex}, @Random_{lastIndex}) ORDER BY 1) AS temp(id, randomNumber) WHERE temp.id = world.id");
-            }
-            else
-            {
-                Enumerable.Range(0, batchSize).ToList().ForEach(i => sb.Append($"UPDATE world SET randomnumber = @Random_{i} WHERE id = @Id_{i};"));
-            }
+            sb.AppendLine("UPDATE world SET randomNumber = CASE id");
+            Enumerable.Range(0, batchSize).ToList().ForEach(i => sb.AppendLine($"when @I{i} then @R{i}"));
+            sb.AppendLine("else randomnumber");
+            sb.AppendLine("end");
+            sb.Append("where id in (");
+            Enumerable.Range(0, batchSize).ToList().ForEach(i => sb.AppendLine($"@I{i}{(lastIndex == i ? "" : ",")} "));
+            sb.Append(")");
 
             return _queries[batchSize] = StringBuilderCache.GetStringAndRelease(sb);
         }


sebastienros:
Would have been interesting to see the impact of each change. It's not just about changing the sql.

benaadams:
Spilt

sebastienros:
This branch seems broken. 

did I make a mistake?

```
dotnet run --config https://raw.githubusercontent.com/aspnet/Benchmarks/master/scenarios/platform.benchmarks.yml --scenario updates --profile aspnet-citrine-lin --application.source.repository https://github.com/benaadams/benchmarks --application.source.branchOrCommit batch-sql --output batch-case.json
```

| load                   |          |
| ---------------------- | -------- |
| CPU Usage (%)          |       -0 |
| Raw CPU Usage (%)      |     1.00 |
| Working Set (MB)       |        7 |
| Build Time (ms)        |    3,270 |
| Start Time (ms)        |        0 |
| Published Size (KB)    |   76,389 |
| First Request (ms)     |       80 |
| Requests/sec           |      122 |
| Requests               |    1,838 |
| Mean latency (ms)      |   142.49 |
| Max latency (ms)       | 2,000.00 |
| Bad responses          |        0 |
| Socket errors          |      290 |
| Read throughput (MB/s) |     0.09 |
| Latency 50th (ms)      |    17.87 |
| Latency 75th (ms)      |    52.31 |
| Latency 90th (ms)      |   992.61 |
| Latency 99th (ms)      | 1,160.00 |

n-stefan:
I tried CASE recently but ended up not opening a PR at TechEmpower because locally it didn't seem faster.
Annoying if it is after all :)

benaadams:
Case is the SQL used by the just js framework for updates

image

n-stefan:
Indeed, perhaps drogon too.

sebastienros:
I checked drogon, it's using "case" but it looks like it's not doing following the specs. 
I can only suspect I am wrong, but here is what I understand:

Instead of using parameters, it create a raw sql string that contains all ids and randoms values in it.

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C%2B%2B/drogon/drogon_benchmark/controllers/UpdatesCtrlRaw.cc#L20-L22

Then this sql query is cached, and on the next call it's reused. So I assume the same sql is reused, and the next request doesn't generate new random values.

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C%2B%2B/drogon/drogon_benchmark/controllers/UpdatesCtrlRaw.cc#L9-L13

This is how the sql cache is defined, not sure if it means it's "static" 
https://github.com/TechEmpower/FrameworkBenchmarks/blob/34f9e25c0c9dad559dc86748e5215e3968559b74/frameworks/C%2B%2B/drogon/drogon_benchmark/controllers/UpdatesCtrlRaw.h#L24

Do you agree with how I read that?

benaadams:
It creates a sql string using n `$` params which are postgres' parameter delimiter; then it uses that sql for everything just changing the parameter values

sebastienros:
Correct, I missed the `$` in front of the %d

benaadams:
I thought the same when I looked at the just sql :)

sebastienros:
Just has a nice way to compute the http fragments:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/JavaScript/just/techempower.js#L438-L443

You made something similar, but their code is nicer.

sebastienros:
I think Just is not following the spec (change my mind) as it creates a `const message` for the /json endpoint. Let me know how you read it.

sebastienros:
But yeah it looks like they all use CASE, so maybe just need to find why your PR is not working. Will try to debug it.

benaadams:
> a const message for the /json endpoint.

That just the headers; which refreshes every second (to update date)? We write the date after rest of the headers rather than merging them all into one block every second

sebastienros:
Was talking about the json object that is returned, and should be instantiated on each request.

benaadams:
Ah; I was looking at the header bit; yeah looks like the `const sJSON = sjs({ message: attr('string') })` is only created once rather than per request

sebastienros:
No this line is fine, it's the serializer. 

About your PR, it works, but looks like it doesn't handle load very well. Maybe the multiplexing is not liking it.

benaadams:
Isn't using multiplexing for updates?
```
connectionString: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
```

sebastienros:
If I enable it it's much faster, no more issues. I'll try to compare with and without CASE when multiplexing is enabled.

sebastienros:
| db                | nomultiplexing | multiplexing |         |
| ----------------- | -------------- | ------------ | ------- |
| CPU Usage (%)     |             96 |           55 | -42.56% |
| Raw CPU Usage (%) |       2,691.19 |     1,546.33 | -42.54% |
| Working Set (MB)  |            531 |          146 | -72.50% |
| Build Time (ms)   |          2,181 |        2,176 |  -0.23% |
| Start Time (ms)   |            327 |          319 |  -2.45% |


| application         | nomultiplexing | multiplexing |          |
| ------------------- | -------------- | ------------ | -------- |
| CPU Usage (%)       |             70 |           70 |   -0.48% |
| Raw CPU Usage (%)   |       1,963.38 |     1,944.69 |   -0.95% |
| Working Set (MB)    |            483 |          458 |   -5.31% |
| Build Time (ms)     |          5,215 |        4,722 |   -9.44% |
| Start Time (ms)     |          1,554 |        8,774 | +464.46% |
| Published Size (KB) |         98,006 |       98,006 |    0.00% |


| load                   | nomultiplexing | multiplexing |         |
| ---------------------- | -------------- | ------------ | ------- |
| CPU Usage (%)          |              2 |            2 | -16.67% |
| Raw CPU Usage (%)      |          44.35 |        42.03 |  -5.23% |
| Working Set (MB)       |              7 |            7 |  +5.00% |
| Build Time (ms)        |          3,695 |        3,681 |  -0.40% |
| Start Time (ms)        |              0 |            0 |         |
| Published Size (KB)    |         76,400 |       76,400 |   0.00% |
| First Request (ms)     |             79 |           78 |  -2.10% |
| Requests/sec           |         12,696 |       18,804 | +48.12% |
| Requests               |        191,581 |      283,743 | +48.11% |
| Mean latency (ms)      |          23.12 |        14.64 | -36.68% |
| Max latency (ms)       |         305.44 |       122.88 | -59.77% |
| Bad responses          |              0 |            0 |         |
| Socket errors          |              0 |            0 |         |
| Read throughput (MB/s) |           9.17 |        13.59 | +48.16% |
| Latency 50th (ms)      |          17.39 |        13.04 | -25.02% |
| Latency 75th (ms)      |          28.70 |        13.88 | -51.63% |
| Latency 90th (ms)      |          45.42 |        15.51 | -65.84% |
| Latency 99th (ms)      |         101.89 |        49.54 | -51.38% |

sebastienros:
Now comparing your branches with multiplexing enabled:

| db                | ben-master | ben-batch-sql |        |
| ----------------- | ---------- | ------------- | ------ |
| CPU Usage (%)     |         56 |            57 | +0.59% |
| Raw CPU Usage (%) |   1,578.76 |      1,584.68 | +0.37% |
| Working Set (MB)  |        139 |           142 | +2.15% |
| Build Time (ms)   |      2,082 |         2,163 | +3.89% |
| Start Time (ms)   |        332 |           310 | -6.63% |


| application         | ben-master | ben-batch-sql |        |
| ------------------- | ---------- | ------------- | ------ |
| CPU Usage (%)       |         71 |            69 | -1.89% |
| Raw CPU Usage (%)   |   1,974.13 |      1,939.83 | -1.74% |
| Working Set (MB)    |        461 |           464 | +0.80% |
| Build Time (ms)     |      4,754 |         4,418 | -7.06% |
| Start Time (ms)     |      8,839 |         8,698 | -1.59% |
| Published Size (KB) |     98,006 |        98,006 |  0.00% |


| load                   | ben-master | ben-batch-sql |         |
| ---------------------- | ---------- | ------------- | ------- |
| CPU Usage (%)          |          1 |             2 | +50.00% |
| Raw CPU Usage (%)      |      42.33 |         42.37 |  +0.09% |
| Working Set (MB)       |          7 |             7 |   0.00% |
| Build Time (ms)        |      3,683 |         3,676 |  -0.19% |
| Start Time (ms)        |          0 |             0 |         |
| Published Size (KB)    |     76,400 |        76,400 |   0.00% |
| First Request (ms)     |         78 |            78 |  +0.43% |
| Requests/sec           |     19,275 |        18,993 |  -1.46% |
| Requests               |    290,731 |       286,568 |  -1.43% |
| Mean latency (ms)      |      13.72 |         13.97 |  +1.82% |
| Max latency (ms)       |     124.29 |        133.24 |  +7.20% |
| Bad responses          |          0 |             0 |         |
| Socket errors          |          0 |             0 |         |
| Read throughput (MB/s) |      13.93 |         13.72 |  -1.46% |
| Latency 50th (ms)      |      12.94 |         13.11 |  +1.34% |
| Latency 75th (ms)      |      13.74 |         13.90 |  +1.21% |
| Latency 90th (ms)      |      14.64 |         14.81 |  +1.14% |
| Latency 99th (ms)      |      37.41 |         38.59 |  +3.15% |

src/BenchmarksApps/Kestrel/PlatformBenchmarks/Data/BatchUpdateString.cs
```diff
@@ -11,8 +11,8 @@ internal class BatchUpdateString
 
         public static DatabaseServer DatabaseServer;
 
-        internal static readonly string[] Ids = Enumerable.Range(0, MaxBatch).Select(i => $"@Id_{i}").ToArray();
-        internal static readonly string[] Randoms = Enumerable.Range(0, MaxBatch).Select(i => $"@Random_{i}").ToArray();
+        internal static readonly string[] Ids = Enumerable.Range(0, MaxBatch).Select(i => $"@I{i}").ToArray();
```
roji, line 14: He's at it again :rofl:

roji: More seriously, when we finally do raw SQL in Npgsql (https://github.com/npgsql/npgsql/issues/3200#issuecomment-708936165) we will start using native PostgreSQL placeholders, which are positional, so we'll have $1, $2.

roji (commented):
Very cool that multiplexing does so well here, thanks @sebastienros. Am a bit surprised the CASE doesn't perform better though.

@sebastienros BTW any specific tool you use to do two runs and get the compared results in a nice GFM table? :wink:

n-stefan:
Regarding multiplexing, any reason not to enable it for the non-platform tests also (over at TechEmpower)?

sebastienros:
@n-stefan multiplexing relies on a pre-release version of npgsql, and we prefer to stick to go-live and stable versions of libraries, at least for these benchmarks. TBH we only are using multiplexing in Platform benchmarks because @benaadams did it, but we would not have done that otherwise.

sebastienros:
@roji 

> any specific tool you use to do two runs and get the compared results in a nice GFM table?

Run crank with `--output baseline.json` then you can either:
- run crank again with `--output change.json` then run `crank compare baseline.json change.json (... anything.json, ...)`
- run crank again with `--compare baseline.json`

n-stefan:
@sebastienros I see, thanks :)

PR #1593, opened 2020-10-01
Add dotnet cover versions using EventSources package

sebastienros:


sebastienros:
fyi @BrennanConroy @halter73

PR #1598, opened 2020-10-27
Use Npgsql 5.0.0-ci.20201101T161311 in platform benchmarks

roji:
This is the latest Npgsql CI package, closer to what will actually be released as 5.0.


roji:
Perf verification: three runs currently averaged at 426812 RPS, with the new version they averaged at 421975 RPS.

PR #1599, opened 2020-10-27
Use EFCore.PG 5.0.0-rc2 and Npgsql 5.0.0-preview1 for non-platform benchmarks

roji:


roji:
Perf verification: three runs currently averaged at 273573 RPS, with the new version they averaged at 263598 RPS. This is a 3.6% regression - we know that the new non-multiplexed pool is a bit slower compared to the custom lock-free implementation in 4.1. I'll investigate turning on multiplexing with EF too.

/cc @ajcvickers

PR #1603, opened 2020-11-17
[Caching] remove one additional interface call and a branch

adamsitnik:
This is a very tiny improvement that gives... +1-2k RPS

The `Get<T>` method is an extension method that uses `IMemoryCache` interface:

https://github.com/dotnet/runtime/blob/69d61271f900afd7ae6dea7f0c5e57d43090f9dd/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs#L18-L21

By using `MemoryCache.TryGetValue` instead, we remove one interface method invocation and a branch


adamsitnik:
@sebastienros ping

PR #1605, opened 2020-11-25
Adds benchmarks for APIs

javiercn:
Adds a new set of benchmarks for common API operation templates.

### DIFF
--- a/build/regressions.config.yml
+++ b/build/regressions.config.yml
@@ -39,6 +39,29 @@ sources:
         - sebastienros
         - agocke
       template: build
+  - name: MVC 
+    table: MvcBenchmarks
+    rules:
+      - include: .
+      - include: Mvc
+        labels: 
+          - area-mvc
+      - exclude: ulib
+      - exclude: arm
+    daysToLoad: 14
+    stdevCount: 7
+    daysToSkip: 1 # allow one day of benchmarks before filing an issue, to detect false positives
+    regressions:
+      probes: 
+        - path: jobs.load.results['wrk/rps/mean']
+          threshold: 2
+      labels: 
+        - Perf
+        - perf-regression
+      owners:
+        - jacalvar
+        - pranavkm
+      template: rps      
 templates:
   rps: |
     A performance regression has been detected for the following scenarios:
--- a/build/regressions.config.yml
+++ b/build/regressions.config.yml
@@ -50,6 +50,8 @@ sources:
     daysToSkip: 1 # allow one day of benchmarks before filing an issue, to detect false positives
     regressions:
       probes: 
+        - path: jobs.load.results['wrk/rps/mean']
+          threshold: 2
         - path: jobs.load.results['bombardier/rps/mean']
           threshold: 2
       labels: 
--- a/build/regressions.config.yml
+++ b/build/regressions.config.yml
@@ -42,17 +42,15 @@ sources:
   - name: MVC 
     table: MvcBenchmarks
     rules:
-      - include: ApiCrud
+      - include: ApiCrud.*
         labels: 
           - area-mvc
-      - exclude: ulib
-      - exclude: arm
     daysToLoad: 14
     stdevCount: 7
     daysToSkip: 1 # allow one day of benchmarks before filing an issue, to detect false positives
     regressions:
       probes: 
-        - path: jobs.load.results['wrk/rps/mean']
+        - path: jobs.load.results['bombardier/rps/mean']
           threshold: 2
       labels: 
         - Perf
--- a/build/regressions.config.yml
+++ b/build/regressions.config.yml
@@ -42,8 +42,7 @@ sources:
   - name: MVC 
     table: MvcBenchmarks
     rules:
-      - include: .
-      - include: Mvc
+      - include: ApiCrud
         labels: 
           - area-mvc
       - exclude: ulib
--- a/src/BenchmarksApps/Mvc/ProductsController.cs
+++ b/src/BenchmarksApps/Mvc/ProductsController.cs
@@ -6,7 +6,7 @@ using Microsoft.AspNetCore.Mvc;
 
 namespace Mvc
 {
-    public class CRUDController : ControllerBase
+    public class ProductsController : Controller
     {
         private static readonly Product[] _fixedProductList = new Product[]
         {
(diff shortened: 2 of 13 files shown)


pranavkm (approved):
Looks fine to me.

davidfowl:
@sebastienros can you look? I assume we'll be tracking improvements to these benchmark as well?

sebastienros:
@sebastienros I need a bot to remind me when I added a comment but didn't submit the review

src/BenchmarksApps/Mvc/ProductsController.cs
```diff
@@ -0,0 +1,126 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Mvc
+{
+    public class ProductsController : ControllerBase
```
davidfowl, line 9: API controller?

javiercn: I don't want to add those behaviors just yet.

We can add more variations later.

davidfowl:
@javiercn Once @sebastienros signs off this is good. I think we also want another variation of this exact same API with "route to code" so that we have a floor.

@sebastienros I'd like us to make an power BI dashboard for each of the operations here so we can track it over time.

javiercn:
@davidfowl this is already good to go.

I reviewed it in person yesterday with Sebastien.

I'm currently working on a follow-up to add modelbinding scenarios

We also spoke about the Power BI reports and there'll be an MVC report tab there.

davidfowl:
> We also spoke about the Power BI reports and there'll be an MVC report tab there.

Great! I'm not sure it should be MVC but should capture the scenario since we plan to have other API centric benchmarks in there (its probably fine tho)

build/regressions.config.yml
```diff
@@ -39,6 +39,29 @@ sources:
...
+      probes: 
+        - path: jobs.load.results['wrk/rps/mean']
+          threshold: 2
+      labels: 
+        - Perf
+        - perf-regression
+      owners:
+        - jacalvar
+        - pranavkm
+      template: rps
```
javiercn, line 64: @sebastienros like this?

build/regressions.config.yml
```diff
@@ -39,6 +39,29 @@ sources:
...
+        labels: 
+          - area-mvc
+      - exclude: ulib
+      - exclude: arm
+    daysToLoad: 14
+    stdevCount: 7
+    daysToSkip: 1 # allow one day of benchmarks before filing an issue, to detect false positives
+    regressions:
+      probes: 
+        - path: jobs.load.results['wrk/rps/mean']
```
javiercn, line 56: @sebastienros we are using bombardier.

Should this point to some other path? Do the scripts below need any adjustment?

build/regressions.config.yml
```diff
@@ -39,6 +39,28 @@ sources:
         - sebastienros
         - agocke
       template: build
+  - name: MVC 
+    table: MvcBenchmarks
+    rules:
+      - include: ApiCrud
```
javiercn, line 45: @sebastienros 
Is it correct to assume that these are prefixes and that indicate the list of `scenarios` to include?

sebastienros: These are regular expression that will match the scenario and description arguments used in the command line. In this case you can just use a "." catch all regex that will be applied to everything in this table. It's useful for instance to add a different rule that can exclude some specific scenarios, or to add custom labels/owners. For instance you could only tag yourself for anything that has "Api.*" (just an example).

It makes me realize that the command line arguments in the azdo yaml file needs to be updated to this:

```
"args": [ "${{ s.arguments }} $(azureProfile) --profile ${{ parameters.profile }} --application.channel edge --application.framework net5.0 --variable warmup=${{ parameters.warmup }} --variable duration=${{ parameters.duration }} --application.options.collectCounters true --no-metadata --no-measurements --session $(session) --description \"${{ s.displayName }} ${{ parameters.profile }}\" --property buildId=\"$(buildId)\" --property buildNumber=\"$(buildNumber)\" --table MvcBenchmarks --sql SQL_CONNECTION_STRING --chart" ]
```

build/regressions.config.yml
```diff
@@ -39,6 +39,28 @@ sources:
...
+        labels: 
+          - area-mvc
+      - exclude: ulib
```
sebastienros, line 48: These two rules can be removed, it's an example of the explanation I gave about these rules, to exclude specific scenarios (of group of) from the bot.

build/regressions.config.yml
sebastienros: The probes is an array. In the table we have a mix of clients, wrk and bombardier, so we need to add a new line, like this:
```
- path: jobs.load.results['bombardier/rps/mean']
```

This just says that the measurements named "bombarder/rps/mean" need to be tracked. If we wanted to track latency we could add the name of the result here too.

build/regressions.config.yml
sebastienros: Looks good

sebastienros:
@javiercn something that is missing is to add an azdo job in the /build/benchmarks.yml file. I can do it if you prefer, but I believe you might also want to do it yourself in order to understand it.

In this file, you can copy the jobs named SingleFile_Linux and SingleFile_Windows, and replace the template by "mvc-scenarios.yml". Also change the dependency to be SingleFile_Linux/SingleFile_Windows and the next job to depend on yours now. These are two jobs as they will run in parallel, one on windows and one on Linux. You can also use warmup and duration of 15 to get more stable results. We care more about build time and size that raw perf for singlefile benchmarks.

build/regressions.config.yml
javiercn: I'm not sure where I need to update this command

src/BenchmarksBot/Tags.cs
```diff
@@ -11,6 +11,11 @@ public class Tags
     {
         public static readonly Dictionary<string, Tags> Scenarios = new Dictionary<string, Tags>(StringComparer.OrdinalIgnoreCase)
         {
```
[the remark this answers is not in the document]
javiercn: Removed

build/regressions.config.yml
javiercn: I've changed it to `- path: jobs.load.results['bombardier/rps/mean']`

build/regressions.config.yml
javiercn: Removed

build/regressions.config.yml
sebastienros: I fixed it, please take a look

PR #1606, opened 2020-11-26
[MVC] Add modelbinding benchmarks

javiercn:
Adds a few benchmarks for modelbinding


build/regressions.config.yml
```diff
@@ -45,6 +45,12 @@ sources:
       - include: ApiCrud.*
         labels: 
           - area-mvc
+      - include: ModelBinding.*
+        labels:
+          - area-mvc
+      - include: Mvc.*
+        labels:
+          - area-mvc
```
javiercn, line 53: @sebastienros I think this will add back the other MVC results?

DamianEdwards:
@javiercn are we still interested in adding this?

PR #1617, opened 2020-12-14
Add jwt auth variant for mvc webapi

HaoK:
Part of https://github.com/dotnet/aspnetcore/issues/27220

Add jwt/cert/auth scenario code to the mvc app and the matching yaml.  JWT is stable with the key in the app.  Cert yaml is still work in progress, they run, but are not fully implemented


HaoK:
Depends on https://github.com/dotnet/crank/pull/202/files for the cert scenarios

src/BenchmarksApps/Mvc/benchmarks.certapi.yml
```diff
@@ -0,0 +1,210 @@
+﻿# Examples:
+# --config benchmarks.certapi.yml --scenario ApiCrudListProducts --profile aspnet-perf-lin
+
+imports:
+- https://raw.githubusercontent.com/dotnet/crank/haok/crank/src/Microsoft.Crank.Jobs.HttpClient/httpclient.yml
```
HaoK, line 5: update branch to master everywhere in this file before merge

src/BenchmarksApps/Mvc/Program.cs
```diff
@@ -19,11 +23,21 @@ public static void Main(string[] args)
...
                 {
-                    webBuilder.UseStartup<Startup>();
+                    webBuilder.UseStartup<Startup>()
+                        .ConfigureKestrel(options =>
+                        {
+#if CERTAUTH
+                            options.ConfigureHttpsDefaults(opt =>
+                            {
+                                opt.ClientCertificateMode = ClientCertificateMode.RequireCertificate;
+                                opt.AllowAnyClientCertificate();
```
sebastienros, line 37: What about testing the one that we are using. I don't think that anyone using this feature would use Any certificate. Or do we know that the host path is still checked?

src/BenchmarksApps/Mvc/Startup.cs
```diff
@@ -28,6 +35,33 @@ public void ConfigureServices(IServiceCollection services)
...
+#if JWTAUTH
+            services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme).AddJwtBearer(o =>
+            {
+                o.TokenValidationParameters.ValidateActor = false;
+                o.TokenValidationParameters.ValidateLifetime = true;
+                o.TokenValidationParameters.ValidateAudience = true;
+                o.TokenValidationParameters.ValidateIssuer = true;
+                o.TokenValidationParameters.ValidIssuer = "Test";
+                o.TokenValidationParameters.ValidAudience = "test";
+                o.TokenValidationParameters.IssuerSigningKey = new SymmetricSecurityKey(Convert.FromBase64String("MFswDQYJKoZIhvcNAQEBBQADSgAwRwJAca32BtkpByiveJTwINuEerWBg2kac7sb"));
```
sebastienros, line 48: Where is this key coming from? Or how do we generate it?

src/BenchmarksApps/Mvc/Startup.cs
```diff
@@ -28,6 +35,33 @@ public void ConfigureServices(IServiceCollection services)
...
+            });
+#elif CERTAUTH
+            services.AddAuthentication(CertificateAuthenticationDefaults.AuthenticationScheme).AddCertificate(o =>
+            {
+                o.AllowedCertificateTypes = CertificateTypes.All;
+                o.RevocationFlag = X509RevocationFlag.EntireChain;
+                o.RevocationMode = X509RevocationMode.NoCheck;
+                o.ValidateCertificateUse = false;
+                o.ValidateValidityPeriod = false;
+            }).AddCertificateCache();
```
sebastienros, line 58: This is the new feature you implemented in 5.0?

src/BenchmarksApps/Mvc/benchmarks.certapi.yml
sebastienros: Worst case we can merge and do the change after (before the branch is closed)

src/BenchmarksApps/Mvc/Program.cs
sebastienros: For instance by passing the certificate file/url in an argument the same way it's done on the client

HaoK: Good point, I'll take a note for now to do a follow up PR that configures more complex validation to see how much that affects perf.

src/BenchmarksApps/Mvc/Startup.cs
HaoK: The key doesn't matter, I just generated this random one and used a jwt generator to generate the encoding/payload, so its hand crafted to match the Jwt tokens.

HaoK: I will put a comment in here though to describe the process in case we need to generate new claims or want to change the key or something in the future

sebastienros: A comment is great, thanks

src/BenchmarksApps/Mvc/Startup.cs
HaoK: Right but unless we change the load to do more than 1 request per client it will never matter, since clients aren't staying on the same connection really.  I think for the initial checkin things are fine, but we might need to look at making the load behave in more 'session' like way

HaoK:
Ok I think this is addressed for first iteration @sebastienros

PR #1619, opened 2021-01-12
Add new jwt/cert scenarios

HaoK:


build/mvc-scenarios.yml
```diff
@@ -60,6 +60,30 @@ parameters:
   - displayName: API remove existing element
     arguments: --scenario ApiCrudDeleteProduct $(crudApiJobs) --property scenario=ApiCrudDeleteProduct
 
+  # CRUD API CERT AUTH
+  - displayName: API return collection data
+    arguments: --scenario ApiCrudListProducts $(crudApiCertJobs) --property scenario=ApiCrudListProducts
```
sebastienros, line 65: The scenario properties should be unique, as they are used to distinguish the benchmarks that are based on the same scenario arguments (ApiCrudListProducts). Here for instance it could be `--property scenario=ApiCrudListProductsCertAuth`

sebastienros: Also, why would we run benchmarks on all methods? Do we expect each benchmark to be impacted differently by the auth?

Why not one method with JWT and one with Certs?
The answer we are trying to answer is "how much does auth cost?" so I assume here the cost is the same in all cases, it won't depend on the payload (in or out).

HaoK: Sure that sounds reasonable, updated

PR #1623, opened 2021-01-22
[NGINX] HTTPS-HTTP, HTTPS-HTTPS, HTTP2-HTTP1 proxy scenarios

alnikola:
Adds HTTPS-HTTP, HTTPS-HTTPS, HTTP2-HTTP1 proxy scenarios on NGINX.
Note: NGINX doesn't seem to support HTTP/2 protocol to the downstream, so only HTTP2 - proxy - HTTP1 - downstream is currently implemented.


docker/nginx/nginx.conf
```diff
@@ -29,4 +29,52 @@ http {
...
+        ssl_certificate /etc/ssl/certs/testCert.crt;
+        ssl_certificate_key /etc/ssl/private/testCert.rsa;
+
+        server_name _;
+
+        access_log off;
+        error_log stderr warn;
+
+        location / {
+            proxy_pass https://backend;
```
Tratcher, line 76: The scenario doesn't dictate if the destination should use http or https. Why did you pick https?

alnikola: It's HTTPS-HTTPS scenario. Or is it redundant?

alnikola: [This one is HTTPS-HTTP](https://github.com/aspnet/Benchmarks/blob/7c7d00e2652a50bbc514ed1c9b9a423731d95b04/docker/nginx/nginx.conf#L34).

Tratcher: This is the http2-http1 scenario. The destination could use either http or https. I just want to know why you picked https, it seems like unneeded overhead.

alnikola: Ah, sorry, I misunderstood. Yes, you are probably right. I will change it back to http.

alnikola: Done

docker/nginx/nginx.dockerfile
```diff
@@ -3,6 +3,8 @@ FROM nginx:latest
 ENV WORKER_PROCESSES auto
 ENV PROXY_PASS http://tfb-database:5000
 
+ADD https://github.com/aspnet/Benchmarks/raw/master/src/Benchmarks/testCert.crt /etc/ssl/certs/testCert.crt
```
Tratcher, line 6: Shouldn't this be pulling from the local file system?

sebastienros: It's on the filesystem if it means it came from github ... It has never failed for me, unless when GH was down, but in this case this file dockerfile can't even be cloned.

But otherwise if it's already there, I agree there is no reason to pull it again. Unless it's in a folder that docker can't access. We do set a docker context after all, that might limit it.

Maybe we should copy these certs in the nginx folder. It would be redundant, but that might make things easier too.

So many options.

sebastienros:
Seeing how this will make the list of scenarios explode, I'd rather find a solution where we can just reuse the current variables (downstreamScheme, ...) and introduce new profiles that will decide what combination of protocols to use, defaulting to http/http. Otherwise we'll get a mix of scenario names and variables different for each proxy.
So maybe having different config files is better, and have some logic switch config files in the docker file or in the yml template. It might require some new global variables to simplify even the existing yarp scenarios and have something homogenous. Like serverScheme, serverProtocol, downstreamScheme, downstreamProtocol.

I think I can try to find a solution to it since you fixed the nginx part.

docker/nginx/nginx.dockerfile
alnikola: What are the boundaries for locally accessing files in the repo from inside of a dockerfile? If it can access the repo root folder, I'd suggest to create a new 'certificates' folder there storing the certs. Otherwise, I'd prefer to download them from GitHub, unless there is a strong objection to it.

alnikola:
@sebastienros It would be much appreciated if you could help me with finding a more efficient way for defining the YARP scenarios configurations.

docker/nginx/nginx.dockerfile
Tratcher: Isn't the docker file just used to build the container? I'd expect it to have full access to the local disk, unless that container is being deployed to another machine?

alnikola: I would expect the 'container builder' to have the full access to the local filesystem, but Sebastian mentioned above that it might not be so.

PR #1624, opened 2021-01-27
Adding nginx scenarios

sebastienros:


sebastienros:
Co-authored-by: @alnikola

docker/nginx/nginx.dockerfile
```diff
@@ -1,11 +1,44 @@
...
+# ARG SERVER_SCHEME 
+ADD nginx-http.conf /etc/nginx/nginx.conf
+
+# Listening to https connections
+FROM base AS scheme-https-http
+# ARG SERVER_SCHEME
+ADD nginx-https.conf /etc/nginx/nginx.conf
+
+# Listening to h2 connections
+FROM base AS scheme-https-http2
```
alnikola, line 33: It should be HTTP/2 - plaintext HTTP 1.1.

sebastienros: This is ${SERVER_SCHEME}-${SERVER_PROTOCOL}
Nothing to do with upstream

sebastienros: http-http2 would mean h2c for instance

alnikola: I got it now. Sorry for the misunderstanding.

PR #1625, opened 2021-02-01
HTTPS and HTTP/2 scenarios on Envoy

alnikola:
This PR depends on #1624


scenarios/proxy.benchmarks.yml
```diff
@@ -203,4 +214,4 @@ profiles:
           - http://localhost:5010
       load:
         endpoints: 
-          - http://localhost:5010
+          - http://localhost:5010
```
sebastienros, line 217:
```suggestion
          - http://localhost:5010

```

docker/envoy/envoy.dockerfile
```diff
@@ -1,11 +1,52 @@
...
+
+# Listening to h2 connections proxying h2
+FROM base AS scheme-https-http2-to-https-http2
+# ARG SERVER_SCHEME
+ADD envoy-http2.yaml /etc/envoy.yaml
+
+# Listening to h2 connections proxying http 1.1
+FROM base AS scheme-https-http2-to-http-http
+# ARG SERVER_SCHEME
+ADD envoy-https-http.yaml /etc/envoy.yaml
```
sebastienros, line 48: Wrong filename?

alnikola: No. That's correct. Envoy negotiates either HTTP 1.1 or HTTP/2 using ALPN, thus it uses the same HTTPS endpoint for both of the protocols.

scenarios/proxy.benchmarks.yml
```diff
@@ -51,11 +51,17 @@ jobs:
       BranchOrCommit: master
       DockerFile: docker/envoy/envoy.dockerfile
       DockerImageName: envoy-reverseproxy
-      DockerContextDirectory: docker/envoy
+      DockerContextDirectory: .
```
sebastienros, line 54: I don't think this works. The context directory should be docker/envoy or the relative path in the docker file won't match

alnikola: Hmm. I changed it because noticed [the same change in NGINX scenario](https://github.com/aspnet/Benchmarks/blob/b0e6602510ef5196dc51532a750e7706e9d4d222/scenarios/proxy.benchmarks.yml#L38). Is it a bug?

sebastienros: Yes, that's a problem. I must have changed it when I did local tests. I am fixing it in nginx. Please do the same here

sebastienros (approved):

scenarios/proxy.benchmarks.yml
alnikola: Fixed

PR #1633, opened 2021-02-16
Add YARP scenario for NET 5.0

MihaZupan:
Intentionally leaving out `supportsServerHttps` and `supportsServerHttp2` to reduce the number of combinations ran for 5.0.
If need be we can add them in the future.

### DIFF
--- a/build/proxies-scenarios.yml
+++ b/build/proxies-scenarios.yml
@@ -23,7 +23,7 @@ parameters:
     supportsServerHttps: true
     supportsServerHttp2: true
   - displayName: YARP-net50
-    arguments: --scenario proxy-yarp        $(proxyJobs)  --property proxy=yarp --application.channel edge --application.framework net5.0
+    arguments: --scenario proxy-yarp        $(proxyJobs)  --property proxy=yarp-net50 --application.framework net5.0
   - displayName: HttpClient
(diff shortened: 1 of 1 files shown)


build/proxies-scenarios.yml
```diff
@@ -22,6 +22,8 @@ parameters:
     arguments: --scenario proxy-yarp        $(proxyJobs)  --property proxy=yarp --application.channel edge --application.framework net6.0
     supportsServerHttps: true
     supportsServerHttp2: true
+  - displayName: YARP-net50
+    arguments: --scenario proxy-yarp        $(proxyJobs)  --property proxy=yarp --application.channel edge --application.framework net5.0
```
sebastienros, line 26: channel current in this case, or don't specify it (default)

MihaZupan:
@sebastienros is there a way to test such a change, or do we let it run in CI and fix it if stuff breaks?

build/proxies-scenarios.yml
sebastienros: I would recommend to give it a different name, like yarp-net5.0 so that it appears automatically in the charts. Otherwise I will have to find a hacky way to distinguish the data coming from the two runs.

MihaZupan: Ah I see the `--property proxy=` now, thought it was going by the "Display name" first

PR #1634, opened 2021-02-16
Add custom jwt auth scenario

HaoK:
Add scenario using JsonWebToken to compare against

### DIFF
--- a/src/BenchmarksApps/Mvc/CustomJwtAuth.cs
+++ b/src/BenchmarksApps/Mvc/CustomJwtAuth.cs
@@ -1,0 +1,58 @@
+﻿using System;
+using System.Security.Claims;
+using System.Text.Encodings.Web;
+using System.Threading.Tasks;
+using JsonWebToken;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.Net.Http.Headers;
+
+namespace Mvc
+{
+    public class CustomJwtAuth : AuthenticationHandler<AuthenticationSchemeOptions>
+    {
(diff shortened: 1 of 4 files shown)


src/BenchmarksApps/Mvc/CustomJwtAuth.cs
```diff
@@ -0,0 +1,68 @@
...
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.Net.Http.Headers;
+
+namespace Mvc
+{
+    public class CustomJwtAuth : AuthenticationHandler<AuthenticationSchemeOptions>
+    {
+        private readonly SymmetricJwk _key = new SymmetricJwk(Convert.FromBase64String("MFswDQYJKoZIhvcNAQEBBQADSgAwRwJAca32BtkpByiveJTwINuEerWBg2kac7sb"));
```
sebastienros, line 15: static

src/BenchmarksApps/Mvc/CustomJwtAuth.cs
```diff
@@ -0,0 +1,68 @@
...
+        private JwtReader _reader;
```
sebastienros, line 16: Same comment here, can these be reused across requests, are they thread-safe? Should be defined in the documentation. It's important for perf to reuse them if they are meant this way.

HaoK: Sure the key and the policy are safe to reuse, I kept the reader as instance since that is likely not thread safe

sebastienros:
Merge it? It's not added in the CI.

HaoK:
its a bit stale let me make sure it still works :)

PR #1637, opened 2021-02-24
Benchmark YARP and NGINX in gRPC streaming scenario

alnikola:


scenarios/proxy.grpc.benchmarks.yml
```diff
@@ -0,0 +1,130 @@
...
+scenarios:
+  proxy-yarp-grpc:
+    downstream:
+      job: grpcAspNetCoreServer
+    application:
+      job: yarp
+    load:
+      job: grpcDotNetClient
+
+profiles:
```
sebastienros, line 53: Would it work if you imported this file instead: https://github.com/aspnet/Benchmarks/blob/master/scenarios/aspnet.profiles.yml?raw=true

In the `imports` sections. I am trying to centralize this list of servers.

scenarios/proxy.grpc.benchmarks.yml
```diff
@@ -0,0 +1,130 @@
...
+    warmup: 5
+    duration: 5
+    logLevel: none
+    requestSize: 0
+    responseSize: 0
+    enableCertAuth: false
+    latency: false
+
+jobs:
+  grpcDotNetClient:
```
sebastienros, line 21: You didn't want to import the existing yaml files that contain these definitions?

scenarios/proxy.grpc.benchmarks.yml
alnikola: I tried that at first, but then I have to define the variables (streams, protocol, etc.) in each of the scenarios. I will add more proxies besides YARP and want to avoid variables duplication.

scenarios/proxy.grpc.benchmarks.yml
alnikola: I tried, but it didn't work properly. Please, check the above comment.

PR #1640, opened 2021-02-26
Updating certificates

sebastienros:
Current one is out of date.
This one is valid for 100 years.
Key size of 4096 bits.

@Tratcher do you know if the key size impacts the traffic perf?


HaoK (approved):
Nice 4.5x rps gains :)

Tratcher:
> @Tratcher do you know if the key size impacts the traffic perf?

I know it can affect the negotiated protocol version which could get you very different results.

PR #1641, opened 2021-03-01
Add no mvc auth scenarios

HaoK:
Jwt/Cert variants that skip mvc/routing and just directly exercise auth


src/BenchmarksApps/Mvc/Startup.cs
```diff
@@ -86,10 +93,41 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILogger<
...
+                    // Deny anonymous request beyond this point.
+                    if (user == null || !user.Identities.Any(identity => identity.IsAuthenticated))
+                    {
+                        // Normally this is a call to challenge
+                        response.StatusCode = 401;
+                        return;
+                    }
+
+                    response.StatusCode = 200;
+                    response.ContentType = "text/html";
```
sebastienros, line 116: Plaintext with "Hello world"

sebastienros (commented):
Can you also add the runs to the /build file for azdo ?

HaoK:
Updated with cleanup and azdo jobs

PR #1643, opened 2021-03-04
Add Houdini EchoTodo benchmark

halter73:


scenarios/houdini.benchmarks.yml
```diff
@@ -0,0 +1,79 @@
+﻿imports:
+  - https://raw.githubusercontent.com/dotnet/crank/master/src/Microsoft.Crank.Jobs.Bombardier/bombardier.yml
+
+jobs:
+  mapaction:
+    source:
+      repository: https://github.com/dotnet/aspnetcore.git
+      branchOrCommit: halter73/benchmark-mapaction
```
pranavkm, line 8: Would these eventually point to master?

pranavkm (approved):

scenarios/houdini.benchmarks.yml
halter73: Yes.

halter73:
This is currently failing with the following error locally:

```
  Determining projects to restore...
  Restored C:\Users\halter73\AppData\Local\Temp\benchmarks-agent\benchmarks-server-34896\mrxziete.n3q\aspnetcore\src\Http\perf\benchmarkapps\MapActionApi\MapActionApi.csproj (in 1.96 sec).
C:\Users\halter73\AppData\Local\Temp\benchmarks-agent\benchmarks-server-34896\mrxziete.n3q\aspnetcore\Directory.Build.props(225,3): error MSB4019: The imported project "C:\Users\halter73\AppData\Local\Temp\benchmarks-agent\benchmarks-server-34896\mrxziete.n3q\aspnetcore\artifacts\bin\GenerateFiles\Directory.Build.props" was not found. Confirm that the expression in the Import declaration "artifacts\bin\GenerateFiles\Directory.Build.props" is correct, and that the file exists on disk. [C:\Users\halter73\AppData\Local\Temp\benchmarks-agent\benchmarks-server-34896\mrxziete.n3q\aspnetcore\src\Http\perf\benchmarkapps\MapActionApi\MapActionApi.csproj]
Exit code: 1
```

It's complaining that "aspnetcore\artifacts\bin\GenerateFiles\Directory.Build.props" was not found. I'm guessing this is an issue with mixing crank and arcade. 

@pranavkm I know there are several MVC benchmark projects that are also in the same repo. In the Houdini benchmark project I'm using a `FrameworkReference` instead of the usual `Reference` and marked it as a `IsTestAssetProject`. That's what [similar projects](https://github.com/dotnet/aspnetcore/search?q=BenchmarksTargetFramework) appear to be doing. You can see the csproj here: https://github.com/dotnet/aspnetcore/pull/30636/files#diff-a62c931a98cf45dc230f55ae69d0919e9961c58072e45afe8022a123654637a6.

sebastienros:
What about implementing plaintext and json at least, and we add them to the main Benchmarks page (trends.benchmarks.yml).

We can keep the benchmarks you made here, but I don't think we should track these for now.

sebastienros:
@halter73 correct, the solution is to add an empty Directory.Build.props file to "clear" the inherited one. Or to move your application in this repos.

sebastienros:
In the end you will have to implement the other TE benchmarks with houdini, because it will be necessary to take it seriously. But again plaintext and json only for now is totally fine to setup something. Ideally with the same "no log" option that we have in Platform if you want to have an idea of how it can compare to middleware/platform.

halter73:
I don't want to double the scenarios overnight. Our benchmark servers are already overworked! /s

Seriously though. The empty Directory.Build.props and Directory.Build.targets fixed the build issue, so thanks! The benchmark is running successfully now locally.

I'll add the plaintext and json scenarios before and verify the pipeline works before merging.

pranavkm:
AFAIK, I don't think the MVC benchmarking apps in the aspnetcore repo work at all. @javiercn and I authored a few benchmarks, but they were added to this repo.

halter73:
Once the APIs stabilize, we should move the benchmarks. Adding the Directory.Build.props and Directory.Build.targets fixed the crank build issue, but the local dev experience leaves a lot to be desired.

sebastienros:
So is having to clone aspnetcore to run the benchmarks on the agent

halter73:
@sebastienros is this good to merge now?

I wanted to test it with https://github.com/aspnet/Benchmarks/tree/halter73/houdini-test, but I can't seem to manually trigger the pipeline for any branch other than main. Is it https://github.com/aspnet/Benchmarks/blob/main/build/azure-pipelines.yml#L12 that prevents manual triggering for other branches?

sebastienros:
No that should work, I can try

sebastienros:
The issue is that there is code mirroring, and it only mirrors main :/

build/houdini-scenarios.yml
```diff
@@ -0,0 +1,37 @@
...
+  type: string
+  default: ''
+
+
+# Scenarios
+- name: scenarios 
+  type: object
+  default: 
+  - displayName: Houdini EchoTodo
+    arguments: --scenario echotodo $(houdiniJobs) --application.channel edge --application.framework net6.0
```
sebastienros, line 23: I don't want it in a separate table, too much work. Can you put it with the other MVC benchmarks instead? they are also using bombardier so it will be easier also for the charts. The other plaintext/json ones should be kept where you put them.

halter73:
> The issue is that there is code mirroring, and it only mirrors main :/

I thought so to, so I manually pushed `halter73/houdini-test` to VSTS, but it still won't let me trigger a run for that branch.

halter73:
> Can you put it with the other MVC benchmarks instead?

Done. Is it time to give it a spin?

PR #1647, opened 2021-03-14
Add gRPC proxy scenarios to dashboards

alnikola:
It also disables running HTTP/2 - HTTP/2 on NGINX because it doesn't support that.

### DIFF
--- a/build/proxies-scenarios.yml
+++ b/build/proxies-scenarios.yml
@@ -96,7 +96,7 @@ parameters:
       requiresServerHttp2: true
       requiresDownstreamHttp2: true
     - displayName: gRPC (h2 - h2)
-      arguments: --variable connections=1 --variable streams=100 --variable serverScheme=https --variable serverProtocol=grpc --variable downstreamScheme=https --variable downstreamProtocol=grpc
+      arguments: --variable connections=1 --variable streams=100 --variable serverScheme=https --variable serverProtocol=grpc --variable downstreamScheme=https --variable downstreamProtocol=grpc --property serverProtocol=grpc --property downstreamProtocol=grpc
       requiresGRPC: true
 # We need to find a way to tell the proxy that the downstream server is expecting h2c and not http/1.1
 #    - displayName: h2 - h2c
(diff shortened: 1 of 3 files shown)


build/proxies-scenarios.yml
```diff
@@ -85,6 +94,10 @@ parameters:
     - displayName: h2 - h2
       arguments: --variable connections=28 --variable serverScheme=https --variable serverProtocol=http2 --variable downstreamScheme=https --variable downstreamProtocol=http2 --property serverProtocol=h2 --property downstreamProtocol=h2
       requiresServerHttp2: true
+      requiresDownstreamHttp2: true
+    - displayName: gRPC (h2 - h2)
+      arguments: --variable connections=1 --variable streams=100 --variable serverScheme=https --variable serverProtocol=grpc --variable downstreamScheme=https --variable downstreamProtocol=grpc
```
sebastienros, line 99: I think you are missing `--property serverProtocol=grpc --property downstreamProtocol=grpc` or there will be no way to isolate those in the charts

build/proxies-scenarios.yml
```diff
@@ -96,9 +109,9 @@ steps:
   - ${{ each payload in parameters.payloads }}:
     - ${{ each protocol in parameters.protocols }}:
       # doesn't requiresServerHttps or supportsServerHttps AND doesn't requiresServerHttp2 or supportsServerHttp2
-      - ${{ if and( or( not( protocol.requiresServerHttps), s.supportsServerHttps), or( not( protocol.requiresServerHttp2), s.supportsServerHttp2) ) }}:
+      - ${{ if and( or( not( protocol.requiresServerHttps), s.supportsServerHttps), or( not( protocol.requiresServerHttp2), s.supportsServerHttp2) ), or( not( protocol.requiresDownstreamHttp2), s.supportsDownstreamHttp2) ), or( not( protocol.requiresServerGRPC), s.supportsGRPC) ) }}:
         # For .NET 5.0, only run HTTP-HTTP and H2-H2 benchmarks
-        - ${{ if or( not( eq( s.displayName, 'YARP-net50')), in( protocol.displayName, 'http - http', 'h2 - h2') ) }}:
+        - ${{ if and( not(s.requiresServerGRPC), or( not( eq( s.displayName, 'YARP-net50')), in( protocol.displayName, 'http - http', 'h2 - h2') ) ) }}:
```
sebastienros, line 114: This new condition, where it's placed, should prevent any `requiresServerGRPC` scenario from running. This line is intended as a special rule for yarp50.

sebastienros: I see you are moving it in another "if". However I still don't think this should be on this line. Or please change the comment to explain that it's for both yarp50 and also to exclude grpc ones.

build/proxies-scenarios.yml
```diff
@@ -110,3 +123,15 @@ steps:
...
+          - task: PublishToAzureServiceBus@1
+            condition: succeededOrFailed()
+            displayName: ${{ s.displayName }} ${{ protocol.displayName }}
+            inputs:
+              connectedServiceName: ${{ parameters.connection }}
+              waitForCompletion: true
+              messageBody: |
+                {
+                  "name": "crank",
+                  "args": [ "--table ProxyBenchmarks --sql SQL_CONNECTION_STRING --chart --session $(session) --profile ${{ parameters.profile }}-grpc --no-metadata --no-measurements --downstream.options.reuseBuild true --load.variables.warmup ${{ parameters.warmup }} --load.variables.duration ${{ parameters.duration }} ${{ s.arguments }} ${{ protocol.arguments }} --description \"${{ s.displayName }} ${{ protocol.displayName }} ${{ parameters.profile }}-grpc\"" ]
```
sebastienros, line 136: `${{ parameters.profile }}-grpc` the grpc suffix is not necessary, it's already in the `displayName`

sebastienros (commented):
You managed to use a single file ... but it's not getting simpler. Fortunately http3 won't be supported soon.

build/proxies-scenarios.yml
alnikola: Fixed.

build/proxies-scenarios.yml
alnikola: Moved that `if` to a different line. `else` would be better here, but I don't know how to use it.

build/proxies-scenarios.yml
alnikola: Removed.

build/proxies-scenarios.yml
```diff
@@ -96,17 +109,30 @@ steps:
...
             inputs:
               connectedServiceName: ${{ parameters.connection }}
               waitForCompletion: true
               messageBody: |
                 {
                   "name": "crank",
-                  "args": [ "--table ProxyBenchmarks --sql SQL_CONNECTION_STRING --chart --session $(session) --profile ${{ parameters.profile }} --no-metadata --no-measurements --downstream.options.reuseBuild true --load.variables.warmup ${{ parameters.warmup }} --load.variables.duration ${{ parameters.duration }} ${{ s.arguments }} ${{ payload.arguments }} ${{ protocol.arguments }} --description \"${{ s.displayName }} ${{ payload.displayName }} ${{ protocol.displayName }} ${{ parameters.profile }}\"" ]
+                  "args": [ "--table ProxyBenchmarks --sql SQL_CONNECTION_STRING --chart --session $(session) --profile ${{ parameters.profile }}-grpc --no-metadata --no-measurements --downstream.options.reuseBuild true --load.variables.warmup ${{ parameters.warmup }} --load.variables.duration ${{ parameters.duration }} ${{ s.arguments }} ${{ protocol.arguments }} --description \"${{ s.displayName }} ${{ protocol.displayName }} ${{ parameters.profile }}\"" ]
```
sebastienros, line 137: why the `-grpc` after `--profile ${{ parameters.profile }}`

alnikola: Ah, I realized that it's wrong. I need to add `-grpc` to scenario name, not profile name. The goal is to show grpc as yet another protocol like h2. Thus, I tried to keep the existing scenarios (like `YARP`), but then patch `--scenario proxy-yarp` argument with an extra `-grpc` suffix to call the right scenario.

alnikola: Did I get it right that the proxy to be shown on the dashboards is identified by `--property proxy=...` and not by `displayName`? I've added new scenarios like "YARP gRPC" with different `--scenario` values, but the same ``--property proxy` value. That also eliminates the second `if` block.

sebastienros:
I am yolo'ing it, and worst case I'll remove the new loop if that's causing an issue with azdo.

A comment I'd have about the duration 5/15 is that from experience the variability is more at the beginning of the run than the end, since jitting happens there. So if you spend 20 seconds benchmarking, it's better to spend the most time in the warmup, and measure the end that is more stable, and more representative of the real numbers.

PR #1648, opened 2021-03-16
Add SQLite

bricelam:
Where do I go from here, @sebastienros?

### DIFF
--- a/src/Benchmarks/Startup.cs
+++ b/src/Benchmarks/Startup.cs
@@ -18,6 +18,7 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.Data.Sqlite;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
@@ -64,6 +65,7 @@ namespace Benchmarks
 
             Console.WriteLine($"Database: {appSettings.Database}");
             Console.WriteLine($"ConnectionString: {appSettings.ConnectionString}");
+            Console.WriteLine($"WAL: {appSettings.WAL}");
 
             switch (appSettings.Database)
             {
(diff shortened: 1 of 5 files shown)


src/BenchmarksServer/hello_world.sql
```diff
@@ -0,0 +1,32 @@
...
+CREATE TABLE world (
+    id INTEGER NOT NULL PRIMARY KEY,
+    randomNumber INTEGER NOT NULL DEFAULT 0
+);
+
+.system echo "BEGIN;" > temp.sql
+.system pwsh /C "1..10000 | %{ 'INSERT INTO world (randomNumber) VALUES (abs(random() % 10000));' >> temp.sql }"
+.system echo "COMMIT;" >> temp.sql
+.read temp.sql
+.system pwsh /C "rm temp.sql"
```
bricelam, line 12: PowerhShell is an acceptable cross-platform procedural language for SQLite, right? 😉

sebastienros: I have to check where to move the sqlite binary file, the BenchamrkServer folder is also legacy.

bricelam: We could also create it lazily in C# if that would be easier than keeping a binary file around.

sebastienros: That might be better. So when sqlite is chosen, on the startup run the code?

sebastienros: Can't be a migration since it should work with ADO.NET/Dapper/EF which are supported here

bricelam:
Updated to lazily create the database in Startup

src/Benchmarks/Startup.cs
```diff
@@ -108,6 +109,78 @@ public void ConfigureServices(IServiceCollection services)
 
                     break;
 #endif
+
+                case DatabaseServer.Sqlite:
```
bricelam, line 113: Should I add a check for `Cache=Shared` in the connection string?

sebastienros: We know what we are doing. And we also might want to test that optionally. So no.

src/Benchmarks/Startup.cs
```diff
@@ -108,6 +109,78 @@ public void ConfigureServices(IServiceCollection services)
...
+                    using (var connection = new SqliteConnection(appSettings.ConnectionString))
+                    {
+                        if (!File.Exists(connection.DataSource))
+                        {
+                            connection.Open();
+
+                            var command = connection.CreateCommand();
+                            command.CommandText =
+                            @"
+                                PRAGMA journal_mode = 'wal';
```
sebastienros, line 123: That should not be the default ... maybe a command line argument, or ENV, so we can decide to test different values. I think the defaults should be whatever sqlite default is if we don't set it.

sebastienros: If taken from dotnet config we then have the choice for ENV or command line, there are other argument in this project, should be easy to copy.

bricelam: I was debating whether to do that or not. Will update.

sebastienros: That's something I definitely want to test, whether it's better or not. And we have different kinds of payload in this app, reads, inserts, updates ... so that should give us more real feedback.

bricelam: Note, WAL is the default for databases created by EF Core.

bricelam: Yeah, the official SQLite guidance says that it's better for concurrent access to the database where there are some writes. But it'll be nice to finally prove (or disprove) that.

PR #1655, opened 2021-04-09
Fix server address for windows

davidfowl:
cc @BrennanConroy

### DIFF
--- a/scenarios/signalr.benchmarks.yml
+++ b/scenarios/signalr.benchmarks.yml
@@ -53,7 +53,7 @@ profiles:
           - http://asp-perf-load:5001
   aspnet-perf-win:
     variables:
-      serverAddress: 10.0.0.104
+      serverAddress: 10.0.0.110
       cores: 12
     jobs:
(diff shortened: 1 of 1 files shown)


scenarios/signalr.benchmarks.yml
```diff
@@ -53,7 +53,7 @@ profiles:
           - http://asp-perf-load:5001
   aspnet-perf-win:
     variables:
-      serverAddress: 10.0.0.104
+      serverAddress: 10.0.0.110
```
BrennanConroy, line 56: Is this for local runs only?

davidfowl: What do you mean?

BrennanConroy: Does the CI use this and was just broken?

davidfowl: Yea the CI doesn't use this, these are the other machines used for test. The IP was just wrong.

PR #1662, opened 2021-04-22
Make Fortunes message required for EF Core

roji:
Makes EF Core skip the IsDBNull check

RPS: 91,423 -> 92,696 (~1.4%)

/cc @ajcvickers


ajcvickers:
@roji It's great to see optimizations like not always calling `isDBNull` for required columns having a real impact like this!

/cc @aspnet/efteam

PR #1663, opened 2021-04-26
Kestrel + fake HttpClient with HttpRequest, HttpRequestMessage

alnikola:
…esponseMessage, HttpResponse

### DIFF
--- a/src/BenchmarksApps/HttpClient/Proxy/ProxyExtensions.cs
+++ b/src/BenchmarksApps/HttpClient/Proxy/ProxyExtensions.cs
@@ -132,6 +132,7 @@ namespace Proxy
             var response = new HttpResponseMessage(System.Net.HttpStatusCode.OK);
             CopyHeaders(response.Headers, request.Headers);
             response.Content = new ByteArrayContent(responseBody);
+            response.Content.Headers.ContentLength = responseBody.Length;
             return response;
         }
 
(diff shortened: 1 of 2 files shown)


alnikola:
@Tratcher I used [this existing header copying code](https://github.com/aspnet/Benchmarks/blob/main/src/BenchmarksApps/HttpClient/Proxy/ProxyExtensions.cs) from HttpClient proxy as the template. Do you mean that this code is incorrect too?

alnikola:
Ah, got it. There is already ContentLength set by the downstream. I fix my test then.

src/BenchmarksApps/HttpClient/Proxy/ProxyExtensions.cs
```diff
@@ -126,6 +127,14 @@ public static async Task CopyProxyHttpResponse(this HttpContext context, HttpRes
             }
         }
 
+        public static HttpResponseMessage CreateResponse(HttpRequestMessage request, byte[] responseBody)
+        {
+            var response = new HttpResponseMessage(System.Net.HttpStatusCode.OK);
+            CopyHeaders(response.Headers, request.Headers);
+            response.Content = new ByteArrayContent(responseBody);
```
Tratcher, line 134: You still need to set the ContentLength.

alnikola: Done, but I don't see any impact on RPS.

alnikola: I discovered that it does increase RPS but only in one scenario - 1024B.

PR #1683, opened 2021-06-20
Update JSON platform benchmarks to use JSON source generator

layomia:
| Profile | Scenario | RPS (Before) | RPS (After) | RPS % increase | Max alloc rate (Before) (b/sec) | Max alloc rate (After) (b/sec) | Max alloc rate % decrease | Latency (Before) (ms) | Latency (After) (ms) | Latency % decrease | Start time + first request (Before) (ms) | Start time (After) | % Start time decrease |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| aspnet-citrine-lin | json | 1,232,440 | 1,265,521 | 2.684187466 | 41,672,184 | 12,429,712 | 70.17264082 | 0.57 | 0.51 | 10.52631579 | 246 | 262 | -6.504065041 |
| | multiple\_queries | 23,452 | 23,434 | -0.07675251578 | 752,562,891 | 727,026,011 | 3.393321715 | 10.92 | 11.01 | -0.8241758242 | 1725 | 1735 | -0.5797101449 |
| | single\_query | 397,556 | 402,294 | 1.191781787 | 1,084,997,824 | 1,098,367,805 | -1.232258785 | 0.66 | 0.66 | 0 | 1940 | 1831 | 5.618556701 |
| | updates | 18,276 | 18,478 | 1.105274677 | 1,072,226,685 | 1,067,972,931 | 0.3967215198 | 14 | 13.86 | 1 | 1955 | 1961 | -0.3069053708 |

We see a small increase in RPS and a big drop in max alloc rate for the `json` benchmark. There's also a small increase in RPS for the `single_query` benchmark. Otherwise, the rest of the benchmarks don't change substantially.

Benchmarks were run in 6 iterations; excluding the worst two and the best result; then taking the mean.
Benchmarks had 30s warmup.
Benchmarks run on `aspnet-citrine-lin` profile.

```
crank --config platform.benchmarks.yml --scenario json --profile aspnet-citrine-lin --variable warmup=30 --iterations 6 --exclude 2:1 --exclude-order "load:wrk/rps/mean;http/rps/mean" --application.options.collectCounters true
```

FYI @sebastienros @adamsitnik @eerhardt @steveharter @ericstj @stephentoub @jkotas @davidfowl

### DIFF
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.Json.cs
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.Json.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Buffers;
 using System.Text.Json;
 
@@ -9,7 +8,7 @@ namespace PlatformBenchmarks
 {
     public partial class BenchmarkApplication
     {
-        private readonly static uint _jsonPayloadSize = (uint)JsonSerializer.SerializeToUtf8Bytes(new JsonMessage { message = "Hello, World!" }, SerializerOptions).Length;
+        private readonly static uint _jsonPayloadSize = (uint)JsonSerializer.SerializeToUtf8Bytes(new JsonMessage { message = "Hello, World!" }, SerializerContext.JsonMessage).Length;
 
         private readonly static AsciiString _jsonPreamble =
             _http11OK +
@@ -30,7 +29,8 @@ namespace PlatformBenchmarks
             utf8JsonWriter.Reset(bodyWriter);
 
             // Body
-            JsonSerializer.Serialize<JsonMessage>(utf8JsonWriter, new JsonMessage { message = "Hello, World!" }, SerializerOptions);
+            JsonSerializer.Serialize(utf8JsonWriter, new JsonMessage { message = "Hello, World!" }, SerializerContext.JsonMessage);
+            utf8JsonWriter.Flush();
         }
     }
 }
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.Json.cs
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.Json.cs
@@ -8,7 +8,14 @@ namespace PlatformBenchmarks
 {
     public partial class BenchmarkApplication
     {
-        private readonly static uint _jsonPayloadSize = (uint)JsonSerializer.SerializeToUtf8Bytes(new JsonMessage { message = "Hello, World!" }, SerializerContext.JsonMessage).Length;
+        private readonly static uint _jsonPayloadSize = (uint)JsonSerializer.SerializeToUtf8Bytes(
+            new JsonMessage { message = "Hello, World!" },
+#if NET6_0_OR_GREATER
+            SerializerContext.JsonMessage
+#else
+            SerializerOptions
+#endif
+            ).Length;
 
         private readonly static AsciiString _jsonPreamble =
             _http11OK +
@@ -29,8 +36,15 @@ namespace PlatformBenchmarks
             utf8JsonWriter.Reset(bodyWriter);
 
             // Body
-            JsonSerializer.Serialize(utf8JsonWriter, new JsonMessage { message = "Hello, World!" }, SerializerContext.JsonMessage);
-            utf8JsonWriter.Flush();
+            JsonSerializer.Serialize(
+                utf8JsonWriter,
+                new JsonMessage { message = "Hello, World!" },
+#if NET6_0_OR_GREATER
+                SerializerContext.JsonMessage
+#else
+                SerializerOptions
+#endif
+                );
         }
     }
 }
--- a/src/BenchmarksApps/Kestrel/PlatformBenchmarks/PlatformBenchmarks.csproj
+++ b/src/BenchmarksApps/Kestrel/PlatformBenchmarks/PlatformBenchmarks.csproj
@@ -33,7 +33,6 @@
   </ItemGroup>
 
   <ItemGroup Condition="$(TargetFramework) == 'net6.0'">
-    <PackageReference Include="System.Text.Json" Version="$(MicrosoftNETCoreAppPackageVersion)" />
     <PackageReference Include="Npgsql" Version="6.0.0-preview4" />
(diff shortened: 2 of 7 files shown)


src/BenchmarksApps/Kestrel/PlatformBenchmarks/BenchmarkApplication.Json.cs
```diff
@@ -30,7 +29,8 @@ private static void Json(ref BufferWriter<WriterAdapter> writer, IBufferWriter<b
             utf8JsonWriter.Reset(bodyWriter);
 
             // Body
-            JsonSerializer.Serialize<JsonMessage>(utf8JsonWriter, new JsonMessage { message = "Hello, World!" }, SerializerOptions);
+            JsonSerializer.Serialize(utf8JsonWriter, new JsonMessage { message = "Hello, World!" }, SerializerContext.JsonMessage);
+            utf8JsonWriter.Flush();
```
layomia, line 33: The serializer should flush the writer here (and elsewhere), not the user. Will make a change in dotnet/runtime then update this PR.

```suggestion
```

adamsitnik:
@layomia I was looking forward to find out what is the difference for Caching benchmark, where so far 40% of time was spent for JSON serialization. So I run the benchmarks against your fork:

| load                   |    net5.0 |    net6.0 |   this PR |
| ---------------------- | --------- | --------- | --------- |
| CPU Usage (%)          | 20        | 21        | 30        |
| Cores usage (%)        | 556       | 597       | 842       |
| Working Set (MB)       | 37        | 38        | 37        |
| Private Memory (MB)    | 358       | 358       | 358       |
| Start Time (ms)        | 0         | 0         | 0         |
| First Request (ms)     | 66        | 67        | 62        |
| **Requests/sec**           | **243,000**   | **260,928**   | **364,224**   |
| Requests               | 3,669,151 | 3,939,804 | 5,499,468 |
| Mean latency (ms)      | 1.22      | 1.09      | 0.90      |
| Max latency (ms)       | 57.32     | 47.40     | 44.63     |
| Bad responses          | 0         | 0         | 0         |
| Socket errors          | 0         | 0         | 0         |
| Read throughput (MB/s) | 764.86    | 821.29    | 1,146.88  |
| Latency 50th (ms)      | 1.03      | 0.96      | 0.67      |
| Latency 75th (ms)      | 1.14      | 1.06      | 0.76      |
| Latency 90th (ms)      | 1.28      | 1.19      | 0.87      |
| Latency 99th (ms)      | 9.57      | 6.34      | 9.40      |

That gives us a **100k RPS GAIN** which is almost `+30%`.

cc @jeffhandley @maryamariyan

adamsitnik (approved):
This is simply most impressive. Great work @layomia !

layomia:
Thanks @adamsitnik! FWIW I set out wanting to measure the caching benchmark diff, but somehow thought it was the "multiple_queries" scenario. Happy the RPS improved, as we hoped!

src/BenchmarksApps/Kestrel/PlatformBenchmarks/PlatformBenchmarks.csproj
```diff
@@ -18,6 +18,7 @@
 
   <ItemGroup>
     <PackageReference Include="Dapper" Version="2.0.30" />
+    <PackageReference Include="System.Text.Json" Version="6.0.0-preview.7.21316.4" />
```
sebastienros, line 21: We still need to decide on the strategy to run with these changes.
Issue here is that it's pinning the STJ's version so we'll potentially miss some regressions. Also it's using non-crossgened versions which will have an impact on startup (not sure how much) and trigger some regressions.
Finally we are now running net5.0 baseline with STJ from net6.0.

Right now the best solution would be to just add some ifdefs for net6.0 and use source generators in this case.

layomia: Thanks. I've made the change to ifdef the implementation using source generation to .NET 6 and higher only. To resolve the issue with non-crossgen'ing & using a pinned version of STJ - we've settled on a [proposal](https://github.com/dotnet/designs/blob/02486c962c09113deec777cabd9a54e9acfef437/accepted/2021/InboxSourceGenerators.md) for having inbox source generators. It should be implemented soon & at that time we can have STJ version with the rest of the framework. cc @ericstj 

Are we currently running net6.0 benchmarks continuously (CI/dashboards etc)? I notice the only TFM in this project is net5.0.

src/BenchmarksApps/Kestrel/PlatformBenchmarks/PlatformBenchmarks.csproj
```diff
@@ -33,6 +33,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="$(TargetFramework) == 'net6.0'">
+    <PackageReference Include="System.Text.Json" Version="6.0.0-preview.7.21326.4" />
```
layomia, line 36:
```suggestion
    <PackageReference Include="System.Text.Json" Version="$(MicrosoftNETCoreAppPackageVersion)" />
```

src/BenchmarksApps/Kestrel/PlatformBenchmarks/PlatformBenchmarks.csproj
ericstj: With 6.0 Preview7 builds this is now the case.  No `PackageReference` is needed to use the source generator.

layomia: Awesome, I've removed the reference. I saw that the internal .NET 6 benchmark runs are using the latest Preview 7 bits, so I believe this PR is unblocked. @davidfowl, can we merge now? (Sebastien is currently OOF)

adamsitnik: @halter73 It seems that both @sebastienros and @davidfowl are offline, could you please take a look (and hopefully merge)?

sebastienros:
Not seeing the whole changeset, would this break on 3.1 or 5.0? Last time I checked the PR it would require different versions of the code.

layomia:
@sebastienros the 3.1 and 5.0 implementations are unchanged in this PR, meaning that they don't use the source generator and things shouldn't break. This is based on ifdef'ing the new code based on `NET6_0_OR_GREATER`. FWIW the generator works well with these TFMs, but we're deciding not to use it for those cases.

halter73 (approved):
LGTM

halter73:
Thanks @layomia

PR #1685, opened 2021-07-07
Escape comma-s

mangod9:
else sdk parses it as a separate param.


pranavkm (approved):
Looks like the right thing to do based on https://github.com/dotnet/msbuild/issues/2999#issuecomment-417562519.

PR #1686, opened 2021-07-08
echo benchmark for Websockets

kondratyev-nv:
Contributes to https://github.com/dotnet/runtime/issues/55293.

## Projects

* New server app - src/BenchmarksApps/Websocket
* New client app - src/WebsocketClient

I haven't added a new worker as I haven't found any information on how it's used and how to test it. If the current code is ok, I can add it in the following PR if necessary.

## Scenarios

* echo - Server responds with the same content received from a client

## Tests

* Run benchmark locally via crank. Replaced links to yaml files to my fork and run crank 

```
crank.exe --config https://raw.githubusercontent.com/kondratyev-nv/Benchmarks/main/scenarios/websocket.benchmarks.yml --scenario websocket --profile local --variable scenario=echo
```

Output
```

| application           |                           |
| --------------------- | ------------------------- |
| CPU Usage (%)         | 39                        |
| Cores usage (%)       | 156                       |
| Working Set (MB)      | 79                        |
| Private Memory (MB)   | 88                        |
| Build Time (ms)       | 31,798                    |
| Start Time (ms)       | 1,637                     |
| Published Size (KB)   | 87,593                    |
| .NET Core SDK Version | 5.0.301                   |
| ASP.NET Core Version  | 5.0.7+67acc3d             |
| .NET Runtime Version  | 5.0.0-rtm.20509.6+cca8c88 |


| warmup                |               |
| --------------------- | ------------- |
| CPU Usage (%)         | 53            |
| Cores usage (%)       | 213           |
| Working Set (MB)      | 48            |
| Private Memory (MB)   | 28            |
| Build Time (ms)       | 14,074        |
| Start Time (ms)       | 870           |
| Published Size (KB)   | 66,597        |
| .NET Core SDK Version | 5.0.301       |
| ASP.NET Core Version  | 5.0.7+67acc3d |
| .NET Runtime Version  | 5.0.7+556582d |
| Max RPS               | 41,633        |
| Requests              | 625,245       |
| Mean latency (ms)     | 2             |


| load                  |               |
| --------------------- | ------------- |
| CPU Usage (%)         | 43            |
| Cores usage (%)       | 173           |
| Working Set (MB)      | 44            |
| Private Memory (MB)   | 26            |
| Build Time (ms)       | 7,077         |
| Start Time (ms)       | 1,437         |
| Published Size (KB)   | 66,597        |
| .NET Core SDK Version | 5.0.301       |
| ASP.NET Core Version  | 5.0.7+67acc3d |
| .NET Runtime Version  | 5.0.7+556582d |
| Max RPS               | 25,665        |
| Requests              | 391,466       |
| Mean latency (ms)     | 4             |
```

P.S. Link in CONTRIBUTING.md leads to 404 😅


src/BenchmarksApps/Websocket/Startup.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                if (result.CloseStatus.HasValue)
+                {
+                    break;
+                }
+
+                receivedMessage += Encoding.UTF8.GetString(buffer, 0, result.Count);
+                if (result.EndOfMessage)
+                {
+                    await webSocket.SendAsync(Encoding.UTF8.GetBytes(receivedMessage).AsMemory(), result.MessageType, result.EndOfMessage, CancellationToken.None);
```
BrennanConroy, line 73: Why create a string then convert it back to a byte array instead of just echoing the byte array?

BrennanConroy:
+@CarnaViire

src/BenchmarksApps/Websocket/Startup.cs
kondratyev-nv: I don't mind changing it to just echoing byte array (actually, this was my initial code 😅). However, to clarify, my idea is that server behavior should be closer to some real-world server. I believe in most cases this would mean that messages have to be read up to an `EndOfMessage` flag before processing and returning a response. Another consideration is that this just an initial implementation that will be extended to other scenarios (e.g. compression), where I think this code will be useful (otherwise, it'll be re-implemented). Let me know what you think 🙂

BrennanConroy: I believe the goal of the benchmark is to measure WebSockets, so we should try to do as little as possible outside of the WebSocket code so we know we are only measuring WebSockets.

When compression is added later there shouldn't need to be any change to the echo code.

In the future if we want to add more scenarios we can always update the code.

kondratyev-nv:
For the new benchmark, I'll keep links in yaml files to my fork while this PR is not ready to be merged, so I won't need to change those every time.

kondratyev-nv:
@BrennanConroy @sebastienros Any comments on this?

src/WebsocketClient/Program.cs
```diff
@@ -0,0 +1,370 @@
...
+                {
+                    case "echo":
+                        for (var i = 0; i < _connections.Count; i++)
+                        {
+                            var id = i;
+                            // kick off a task per connection so they don't wait for other connections when sending "Echo"
+                            _ = Task.Run(async () =>
+                            {
+                                var buffer = new byte[1024 * 4];
+                                var message = Encoding.UTF8.GetBytes(DateTime.UtcNow.ToString("O"));
```
BrennanConroy, line 111: I think this message should be configurable, either choose from a table of data (this could include images etc. for when we add compression tests), or just "send 1000 bytes" so we can test different payload sizes. For starters might be good to just choose a 100 or 1000 byte message of random bytes.

BrennanConroy (commented):
Looks good overall, we'll wait for Seb to be back so he can take a look at the yml files etc.

src/WebsocketClient/Program.cs
```diff
@@ -0,0 +1,370 @@
...
+                    break;
+                }
+                else
+                {
+                    response = await _connections[id].ReceiveAsync(buffer.AsMemory(), cts.Token);
+                }
+            }
+
+            _echoTimers[id].Stop();
+            LogLatency(_echoTimers[id].Elapsed, id);
```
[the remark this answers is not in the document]
BrennanConroy: Oh I see, this isn't the main loop, disregard

src/WebsocketClient/Program.cs
kondratyev-nv: I can change that, the question is what would be the best way to configure it.
1. I can add another CLI arg to set a number of bytes for a randomly generated message. The thing is that I'm not sure if this would be useful in comparison to some just constant size (let's say 1000).
2. I expect compression to be a different scenario, and that way it'd make more sense to have a different set of payloads for that (compression won't work well with random values). That's the reason I expect option 1 not to be reused for compression scenarios.

So my suggestion would be to do it simple for now - randomly generate a constant number of bytes for a message. Let me know what you think.

BrennanConroy: > randomly generate a constant number of bytes for a message

Yes, that's what I was trying to get at.

stephentoub:
How close is this to being merged?
cc: @davidfowl

kondratyev-nv:
> How close is this to being merged?

From my end, after PR is approved all I need to do is to change references to the correct remote repo in yml files.

stephentoub:
@sebastienros?  I was hoping to use this against citrine and some of the other servers to help validate https://github.com/dotnet/runtime/pull/56282.

src/BenchmarksApps/Websocket/NuGet.config
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
```
sebastienros, line 1: This file should not be necessary. There is one in the root folder with the same sources.

docker/benchmarks/run-server.sh
```diff
@@ -76,7 +76,8 @@ if lscpu | grep -q 'AuthenticAMD'; then
   hardware=AMD64
 fi
 
-mkdir -p /mnt/benchmarks
+mount_dir=/mnt/benchmarks
```
sebastienros, line 79: This file is not used anymore since we use crank.

src/BenchmarksApps/Websocket/BenchmarkServer.csproj
```diff
@@ -0,0 +1,30 @@
...
+
+  <ItemGroup>
+    <Folder Include="wwwroot\" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Crank.EventSources" Version="0.1.0-alpha.20501.1" />
+  </ItemGroup>
+
+  <!-- These references are used when running on the Benchmarks Server -->
```
sebastienros, line 24: This is not requires anymore, crank inject it before it compiles the app.

src/BenchmarksApps/Websocket/BenchmarkServer.csproj
```diff
@@ -0,0 +1,30 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework Condition="'$(BenchmarksTargetFramework)' != ''">$(BenchmarksTargetFramework)</TargetFramework>
```
sebastienros, line 5: This is not requires anymore, crank inject it before it compiles the app.

src/BenchmarksApps/Websocket/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace BenchmarkServer
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            Console.WriteLine($"Process ID: {Process.GetCurrentProcess().Id}");
+
+            BenchmarksEventSource.MeasureAspNetVersion();
```
sebastienros, line 19: That's fine, but FYI crank now stores it automatically. I have seen it get skipped because of event pipes timing issues that I decide to let the crank server record it automatically.

src/BenchmarksDriver2/benchmarks.websocket.yml
```diff
@@ -0,0 +1,93 @@
+﻿# Examples:
```
sebastienros, line 1: This file should be removed. I don't expect it to be used.

src/WebsocketClient/BenchmarksEventSource.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System.Diagnostics.Tracing;
```
sebastienros, line 1: This file can be removed, the projects use the NuGet package already.

sebastienros:
@stephentoub Since the code is already in a PR, the yml files are available to pass to a crank command line even if it's not merged yet. So the command line that was shared already works, it's just pulling the apps from the branch. Here I updated it to point to the perf-lin machine for instance.

```
crank.exe --config https://raw.githubusercontent.com/kondratyev-nv/Benchmarks/main/scenarios/websocket.benchmarks.yml --scenario websocket --profile aspnet-perf-lin --variable scenario=echo
```

Then to use custom dlls, add `--application.options.outputFiles c:\bin\*.dll` if the "application" tier.

scenarios/websocket.benchmarks.yml
```diff
@@ -0,0 +1,24 @@
...
+
+variables:
+    serverPort: 5000
+    serverScheme: ws
+
+scenarios:
+  websocket:
+    application:
+      job: websocketServer
+    warmup:
```
sebastienros, line 17: Ideally the load generator has a warmup argument such that we only need to run the app once, and it will issue both the warmup and the actual load. Usually the same loop but with a timer that start measuring requests after the warmup delay has passed.

src/WebsocketClient/Program.cs
```diff
@@ -0,0 +1,373 @@
...
+
+        private static List<ClientWebSocket> _connections;
+        private static List<int> _requestsPerConnection;
+        private static List<List<double>> _latencyPerConnection;
+        private static List<(double sum, int count)> _latencyAverage;
+        private static Stopwatch _workTimer = new Stopwatch();
+        private static List<Stopwatch> _echoTimers;
+        private static bool _stopped;
+        private static SemaphoreSlim _lock = new SemaphoreSlim(1);
+        private static int _totalRequests;
```
sebastienros, line 29: Is there a way to display the number of errors/bad responses, even if it's 0? That helps confirm everything went well during the run.

stephentoub:
@sebastienros, I'd tried that previously (and again now), but get this:
```
The specified endpoint url 'http://asp-perf-lin:5001' for 'application' is invalid or not responsive: "No such host is known. (asp-perf-lin:5001)"
```

sebastienros:
@stephentoub Try to access it in a browser (http://asp-perf-lin:5001) you need to be in the VPN. I just tried it an it worked.

stephentoub:
Ah, somehow my VPN refuses to connect.

adityamandaleeka:
@stephentoub Is it working for you now?

Also, I just looked at your other question about throughput figures for this benchmark. @sebastienros I see client side figures for requests and requests/sec emitted by this benchmark (under "warmup" and "load"). Just to confirm, those are successful/completed requests right? Not just requests sent? In other words, is that number a good representation of the throughput of the application?

sebastienros:
@adityamandaleeka I actually added a comment on the PR to always display the number of failures. And to not display the warmup job as an independent result.

BrennanConroy:
> Just to confirm, those are successful/completed requests right? Not just requests sent?

These numbers are generated by the client code at https://github.com/aspnet/Benchmarks/pull/1686/files#diff-2b8caa6c2f3e1c51ab39e188615427641d0906941c8cb85b80aea66987fadbb9R176 for every end of message received.

I'm not sure what an unsuccessful request would look like for WebSockets, do you just mean it throws?

scenarios/websocket.benchmarks.yml
```diff
@@ -0,0 +1,21 @@
+﻿# Examples:
+# --config benchmarks.websocket.yml --scenario websocket --variable scenario=echo --profile asp-perf-lin
+
+imports:
+  - https://raw.githubusercontent.com/kondratyev-nv/Benchmarks/main/src/WebsocketClient/websocketclient.yml
```
BrennanConroy, line 5: fix links

src/BenchmarksApps/Websocket/BenchmarkServer.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
BrennanConroy, line 4:
```suggestion
    <TargetFramework>net6.0</TargetFramework>
    <TargetFramework Condition="'$(BenchmarksTargetFramework)' != ''">$(BenchmarksTargetFramework)</TargetFramework>
```

src/BenchmarksApps/Websocket/BenchmarkServer.csproj
```diff
@@ -0,0 +1,23 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="wwwroot\" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Crank.EventSources" Version="0.1.0-alpha.20501.1" />
+  </ItemGroup>
+
```
BrennanConroy, line 22:
```suggestion

<!-- These references are used when running on the Benchmarks Server -->
  <ItemGroup Condition="'$(BenchmarksTargetFramework)' != ''">
    <FrameworkReference Update="Microsoft.AspNetCore.App" RuntimeFrameworkVersion="$(MicrosoftAspNetCoreAppPackageVersion)" />
    <FrameworkReference Update="Microsoft.NETCore.App" RuntimeFrameworkVersion="$(MicrosoftNETCoreAppPackageVersion)" />
  </ItemGroup>
  
```

src/BenchmarksApps/Websocket/websocketServer.yml
```diff
@@ -0,0 +1,10 @@
+
+jobs:
+  websocketServer:
+    source:
+      repository: https://github.com/kondratyev-nv/benchmarks.git
```
BrennanConroy, line 5: link

src/WebsocketClient/websocketclient.yml
```diff
@@ -0,0 +1,19 @@
+﻿jobs:
+  websocketClient:
+    source:
+      repository: https://github.com/kondratyev-nv/benchmarks.git
```
BrennanConroy, line 4: link

BrennanConroy (approved):

src/BenchmarksApps/Websocket/BenchmarkServer.csproj
sebastienros: No, I had it removed, it's not required since last year. Some of our projects might still have it, but the agent is able to inject it.

src/BenchmarksApps/Websocket/BenchmarkServer.csproj
sebastienros: Not required either

build/websocket-scenarios.yml
```diff
@@ -0,0 +1,39 @@
...
+    condition: succeededOrFailed()
+    displayName: ${{ s.displayName }}
+    inputs:
+      connectedServiceName: ${{ parameters.connection }}
+      waitForCompletion: true
+      messageBody: |
+        {
+          "timeout": "00:5:00",
+          "name": "crank",
+          "args": [ "${{ s.arguments }} --application.options.collectCounters true --load.options.collectCounters true --warmup.variables.duration ${{ parameters.warmup }} --load.variables.duration ${{ parameters.duration }} --profile ${{ parameters.profile }} --load.channel edge --load.framework net6.0 --warmup.channel edge --warmup.framework net6.0 --application.channel edge --application.framework net6.0 --session $(session) --table WebsocketBenchmarks --sql SQL_CONNECTION_STRING --chart --no-measurements --no-metadata --warmup.options.discardResults true" ]
```
sebastienros, line 37:
```suggestion
          "args": [ "${{ s.arguments }} --application.options.collectCounters true --load.options.collectCounters true --load.variables.warmup ${{ parameters.warmup }} --load.variables.duration ${{ parameters.duration }} --profile ${{ parameters.profile }} --load.channel edge --load.framework net6.0 --application.channel edge --application.framework net6.0 --session $(session) --table WebsocketBenchmarks --sql SQL_CONNECTION_STRING --chart --no-measurements --no-metadata --warmup.options.discardResults true" ]
```

sebastienros: In this app, the warmup is handled. In the signalr client it is not, and a different job is used to simulate the warmup.

src/BenchmarksApps/Websocket/websocket.json
```diff
@@ -0,0 +1,19 @@
+{
```
sebastienros, line 1: This file should be deleted, it's not used.

sebastienros:
Merged to see if that runs successfully on the CI. Would be useful to add it to the scenarios/readme.md for visibility. If you don't do it I will.

kondratyev-nv:
I've already updated Readme in this PR - https://github.com/aspnet/Benchmarks/tree/main/scenarios#websockets-benchmarks.

PR #1692, opened 2021-08-17
Update grpc-v2-scenarios.yml

JamesNK:


sebastienros:
Had to remove the `Protocol:` part since it's breaking the yml format. I believe I could have just added quotes. Feel free to do it if you prefer.

PR #1706, opened 2021-11-10
Add initial HTTP client benchmark

CarnaViire:
Add infrastructure for HttpClient benchmarks and first simple GET scenario. Additional parameters and scenarios TBD.

Set of parameters (see readme):
 - HTTP version - 1.1 vs 2.0 vs 3.0
 - Number of clients, concurrency per client (+ HttpClient's H/1 & H/2 specific params)
 - Duration (warmup & test)

Set of metrics:
 - Time to headers (min, max, 50-75-90-99 percentile) 
 - Time to first content byte (min, max, 50-75-90-99 percentile) 
 - Time to last content byte (min, max, 50-75-90-99 percentile) 
 - Mean RPS 
 - Total requests count 
 - Exception count 
 - Bad response count 

*Note:* HTTP/3 works on linux (aspnet-perf-lin) but doesn't seem to work on windows (aspnet-perf-win), reasons yet unknown.

/cc @geoffkizer @scalablecory @ManickaP @wfurt @MihaZupan @karelz @JamesNK


src/BenchmarksApps/HttpClientBenchmarks/Servers/Kestrel/Program.cs
```diff
@@ -0,0 +1,37 @@
...
+});
+var app = builder.Build();
+
+app.MapGet("/", () => Results.Ok());
+
+app.MapGet("/get", () => "Hello World!");
+
+await app.StartAsync();
+
+Console.WriteLine("Application started"); // readyStateText
```
sebastienros, line 35: The default logging should already output that. But doesn't hurt.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+            await RunScenario();
+            Log("Scenario done");
+        });
+
+        return await rootCommand.InvokeAsync(args);
+    }
+
+    private static void Setup()
+    {
+        BenchmarksEventSource.Register("IsServerGC", Operations.First, Operations.First, "Server GC enabled", "Server GC is enabled", "");
```
sebastienros, line 56: Not sure it matters to report these. Or is there a specific reason?

scenarios/README.md
```diff
@@ -401,6 +401,41 @@ The scenario named `custom` can be used to pass any custom filter variable like
...
+- HTTP Version to use:
+  - `--variable httpVersion="1.1"` (default)
+  - `--variable httpVersion="2.0"`
+  - `--variable httpVersion="3.0"`
+- Number of clients:
+  - `--variable numberOfClients=<N>` (default: `1`)
+- Number of concurrect requests per one client:
+  - `--variable concurrencyPerClient=<N>` (default: `12`)
+- Max number of HTTP/1.1 connections per server:
+  - `--variable http11MaxConnectionsPerServer=<N>` (default: `1`)
```
sebastienros, line 430: What happens with all the defaults: httpVersion=1.1, concurrencyPerClient=12, http11MaxConnectionsPerServer=1

One HttpClient instance, 12 concurrent tasks can use it, but a single connection allowed? To me it looks like `http11MaxConnectionsPerServer` should not be set and be driven by `concurrencyPerClient` indirectly. But might be my limited knowledge of HttpClient.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+using System.Runtime;
+using Microsoft.Crank.EventSources;
+
+namespace HttpClientBenchmarks;
+
+class Program
+{
+    private static ClientOptions _options = null!;
+    private static List<HttpClient> _httpClients = new();
+    private static ConcurrentBag<long>? _headersTimes;
```
sebastienros, line 15: Do you also intend to accept sending custom headers? (e.g., `-H` multiple times)

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+            _httpClients.Add(new HttpClient(handler)
+            {
+                BaseAddress = new Uri(_options.Url!),
+                DefaultRequestVersion = _options.HttpVersion!,
+                DefaultVersionPolicy = HttpVersionPolicy.RequestVersionExact
+            });
+        }
+    }
+
+    private static async Task Warmup()
```
sebastienros, line 80: Would be nice to measure the first request independently, we track it. But this client might not care about it. We can look at it if I start using it for more benchmarks.

scenarios/README.md
CarnaViire: Concurrent tasks end up competing for a single HTTP/1 connection, yes. But that would be an honest comparison with HTTP/2 and 3, that also have 1 connection, but use it more efficiently via multiplexing different requests on a single connection.

scenarios/README.md
```diff
@@ -401,6 +401,41 @@ The scenario named `custom` can be used to pass any custom filter variable like
...
+```
+crank --config https://raw.githubusercontent.com/CarnaViire/Benchmarks/http3-benchmarks/scenarios/httpclient.benchmarks.yml --scenario httpclient-kestrel-get --profile aspnet-perf-lin
+```
+*TODO: update to `https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/httpclient.benchmarks.yml`*
+
+### Available scenarios
+
+- `httpclient-kestrel-get`: Execute GET requests to the server. Client is HttpClient, server is Kestrel (minimal API).
+
+#### Arguments
```
geoffkizer, line 419: How do I control TLS vs non-TLS?

scenarios/README.md
MihaZupan: Infinite / concurrencyPerClient may be a better default. While it's great that we have an option to test the contended scenario, I think we will for the most part be interested in the actual throughput of the system and not a single connection.

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,44 @@
...
+    http11MaxConnectionsPerServer: 1
+    http20EnableMultipleConnections: false
+    warmup: 15
+    duration: 30
+
+scenarios:
+  httpclient-kestrel-get:
+    application:
+      job: kestrel
+    load:
```
geoffkizer, line 21: It's a bit confusing to me that the "application" is kestrel and the "load" is httpClient.

The thing we want to measure here in httpClient. So why isn't the application httpClient?

What will happen if I turn on perf collection, for example? Will it collect a perf trace of kestrel or httpClient?

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+            var handler = new SocketsHttpHandler() 
+            {
+                // accept all certs
+                SslOptions = new SslClientAuthenticationOptions { RemoteCertificateValidationCallback = delegate { return true; } },
+                MaxConnectionsPerServer = _options.Http11MaxConnectionsPerServer,
+                EnableMultipleHttp2Connections = _options.Http20EnableMultipleConnections
+            };
+            _httpClients.Add(new HttpClient(handler)
+            {
+                BaseAddress = new Uri(_options.Url!),
```
MihaZupan, line 73: Using the `BaseAddress` and passing a relative Uri to `Get` / `SendAsync` will result in quite a bit of Uri overhead per request.
Unless that was intended, we shouldn't set `BaseAddress`, but create the Uri instance once and re-use it between requests.

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,44 @@
...
+    readyStateText: Application started # app should write this line to output! 
+    arguments: "--urls https://{{serverAddress}}:{{serverPort}}"
+
+  httpClient:
+    source:
+      repository: https://github.com/CarnaViire/benchmarks.git #https://github.com/aspnet/benchmarks.git
+      #localFolder: C:\Users\knatalia\dev\git\Benchmarks
+      branchOrCommit: http3-benchmarks #main
+      project: src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/HttpClient.csproj
+    isConsoleApp: true
```
geoffkizer, line 42: What does this setting do?

scenarios/README.md
geoffkizer: I agree, the basic scenario should be infinite connections for all HTTP versions.

scenarios/httpclient.benchmarks.yml
sebastienros: When you enable collection or counters you explicitly set which "job" does it, and you can set both if you want.
Something like `--load.collect true` will work.

scenarios/README.md
```diff
@@ -401,6 +401,41 @@ The scenario named `custom` can be used to pass any custom filter variable like
...
+
+- HTTP Version to use:
+  - `--variable httpVersion="1.1"` (default)
+  - `--variable httpVersion="2.0"`
+  - `--variable httpVersion="3.0"`
+- Number of clients:
+  - `--variable numberOfClients=<N>` (default: `1`)
```
geoffkizer, line 426: Since this refers to the number of HttpClient instances (I think right?) can we consistently use the term HttpClient instead of Client? E.g. numberOfHttpClients, concurrencyPerHttpClient, etc. The term "client" is vague and could mean lots of things.

scenarios/README.md
```diff
@@ -401,6 +401,41 @@ The scenario named `custom` can be used to pass any custom filter variable like
...
+- Number of concurrect requests per one client:
+  - `--variable concurrencyPerClient=<N>` (default: `12`)
```
geoffkizer, line 428: This seems like a low default. 

I assume we want to max CPU for this test. As such shouldn't this be much larger?

scenarios/httpclient.benchmarks.yml
geoffkizer: Ok, then what's the difference between "application" and "load"? Are they just arbitrary names? If so can we call them something clearer?

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+        rootCommand.Handler = CommandHandler.Create<ClientOptions>(async (options) =>
+        {
+            _options = options;
+            Log("HttpClient benchmark");
+            Log("Options: " + _options);
+
+            Setup();
+            Log("Setup done");
+
+            await Warmup();
```
geoffkizer, line 44: This seems a little weird, because it will cause the client app to basically "pause" briefly between warmup phase and the actual run. I don't think we want that. Rather, I think we want a continuous run where the first N seconds are the warmup phase and the rest is the actual run.

scenarios/httpclient.benchmarks.yml
sebastienros: They are arbitrary, but they are mapped in the profiles to specific machines. I have an issue/task to create another indirection between names and target machines, which would allow use to change the names without impacting the profiles (aspnet-perf-win, aspnet-citrine-arm, ...)

geoffkizer: Interesting. What sort of hardware will the httpClient job end up running on, then? I assume we want this to run on the "big" hardware, meaning aspnet-perf-win etc... though it's also interesting to run it on smaller boxes, like 4 CPUs or 8 CPUs.

sebastienros: You set it on each run. This is not the issue with this PR, when a crank command is run we specify the profile (target machines) in the crank arguments. And our continuous runs will do that on several machines, including Azure VMs.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+    {
+        var oneByteArray = new byte[1];
+        var drainArray = new byte[81920];
+        var stopwatch = Stopwatch.StartNew();
+        while (!token.IsCancellationRequested)
+        {
+            try
+            {
+                stopwatch.Restart();
+                using var result = await client.GetAsync("/get", HttpCompletionOption.ResponseHeadersRead, token);
```
geoffkizer, line 164: Passing the cancellation token here means we will need to register for it on every request, which adds some amount of cost, possibly nontrivial.

While it's interesting to measure the cost of passing a cancellation token here, I'm not sure we want to do that in the base scenario.

It seems like it's better to just not pass the cancellation token and instead just let the call complete normally. We will see that the cancellation token is cancelled when the loop iterates.

scenarios/httpclient.benchmarks.yml
geoffkizer: Okay, but I assume I want to use the profile "aspnet-perf-lin" or "aspnet-perf-win" typically, right? What will happen when I do that? Where will each job run and on what sort of hardware?

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+using System.Diagnostics;
+using System.Net.Security;
+using System.Runtime;
+using Microsoft.Crank.EventSources;
+
+namespace HttpClientBenchmarks;
+
+class Program
+{
+    private static ClientOptions _options = null!;
```
geoffkizer, line 13: All these static fields should be "s_" instead of just "_"

scenarios/httpclient.benchmarks.yml
sebastienros: @geoffkizer You are right, we need to invert the name until I am doing the mappings feature, because otherwise stating asp-perf-win will still run the client on a Linux box.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+        {
+            try
+            {
+                stopwatch.Restart();
+                using var result = await client.GetAsync("/get", HttpCompletionOption.ResponseHeadersRead, token);
+                var headersTime = stopwatch.ElapsedTicks;
+                if (result.IsSuccessStatusCode)
+                {
+                    var content = await result.Content.ReadAsStreamAsync();
+                    var bytesRead = await content.ReadAsync(oneByteArray, token);
```
geoffkizer, line 169: Reading only one byte here means we will always have to perform at least two ReadAsync calls.

This might not matter much in practice, but it seems like it would be better to avoid this. I would suggest we read as much as we can in the first ReadAsync call, then record the "content start" time when that returns. That call should return as soon as any data is available, so this shouldn't change how the "content start" is measured in practice.

Then, loop and ReadAsync until we get EOF.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+                    var contentStartTime = stopwatch.ElapsedTicks;
+                    while (bytesRead != 0)
+                    {
+                        bytesRead = await content.ReadAsync(drainArray, token);
+                    }
+                    var contentEndTime = stopwatch.ElapsedTicks;
+
+                    if (collectMetrics)
+                    {
+                        _headersTimes!.Add(headersTime);
```
geoffkizer, line 179: I worry about using ConcurrentBag here, and also the interlocked operations below. 

We can avoid any sort of cross-thread data sharing here by collecting stats locally for this Task, and then returning them when the test is complete and aggregating the results from the various Tasks.

geoffkizer: Also: Aren't these bags going to get big? Like really big?

geoffkizer: Additionally: I'm not sure how much I care about these metrics. A big part of what these are going to measure is *server* latency. I don't care about that here.

Stats like these are interesting for a load-generating app like WRK to capture and report, because they are measuring the *server* performance. And we don't really care if there's extra overhead in the load generating app because we can just run multiple instances of it (or run it on bigger hardware, etc).

But **we are not trying to build WRK here**. "httpClient" is the perf app. "kestrel" is the equivalent of WRK in this case -- though since this is a client scenario, it's not really a "load generator", more like a "load receiver". So if anything, we should be gathering metrics in the kestrel app, not the httpClient app.

scenarios/README.md
CarnaViire: This was mostly because we intend to compare to other things like LLHTTP, and it would be an LLHTTP connection in that case; thought, the term "HTTP client" might also apply to it, what do you think @MihaZupan @scalablecory?

scenarios/README.md
CarnaViire: Ok, since there's no way to set the number of connections strictly to concurrencyPerClient for all HTTP versions, then I see `http11MaxConnectionsPerServer=int.MaxValue` and `http20EnableMultipleConnections=true` as defaults will be closest to what you want. We would still have only one connection for HTTP/3 though.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
CarnaViire: I am not sure what we can measure on Kestrel side though.
On the client side, it might be interesting to measure e.g. time between calling SendAsync and sending the request on the wire, but it's not easy to get this info from the outside.
But still I see the value of current metrics in terms of comparison between different HTTP versions, while server latency remains the same.

Do you have some ideas on what can we additionally measure?

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
CarnaViire: I passed CT to avoid hanging forever in case something will go wrong, but that actually will not be the case inside SendAsync because HttpClient timeout would still apply. I believe I would still need to use it later when reading response content. What do you think?

scenarios/README.md
CarnaViire: We talked below how we are not building a WRK here. Would we still want to max out the CPU by default?

scenarios/httpclient.benchmarks.yml
CarnaViire: It means that the app is "ready" from the start, so Crank would not wait for a signal from the app to consider it started -- as opposing to server which needs some time to start before a load can start attacking it. In case of server or any other app which need some time to start, you would need to set a different setting "readyStateText" and then output that line from the app to signal to Crank that app is ready.
https://github.com/dotnet/crank/blob/main/src/Microsoft.Crank.Controller/benchmarks.schema.json#L154

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
MihaZupan: Even if we set the timeout to infinite, HttpClient will still pass the `_pendingRequestsCts` here, which means we will still incur cancellation-related overhead.

We could choose to ignore cancellation here entirely. 

If we're not using any HttpClient features, could we use `HttpMessageInvoker(SocketsHttpHandler)` instead?
We'd still have the option to test what the overhead of cancellation is if we wanted to.

MihaZupan: The behavior of the underlying `CancellationTokenSource` could also be different here. With `HttpClient`, a single CTS will be contended by all requests. In cases like YARP, this would never be the case as each request has dedicated CTSs.

scenarios/httpclient.benchmarks.yml
CarnaViire: I can rename them to whatever. I would just need to copy and edit the profiles from `aspnet.profiles.yml` instead of importing them. (until the mapping is implemented as @sebastienros mentioned)
Thanks @sebastienros  for pointing out that in `asp-perf-win` profile only `application: endpoints: - http://asp-perf-win:5001` is windows and `load: endpoints: - http://asp-perf-db:5001` is linux actually; I didn't look inside profiles closely enough so I wasn't aware. I will invert the machines. 

Interestingly, that might be connected to why HTTP/3 wasn't working on `asp-perf-win`

Also, @sebastienros what defines an order in which jobs are created? Is it the order here in yml in `scenarios` definition, or something else?

scenarios/README.md
CarnaViire: Since the idea of the benchmark originated from HTTP/3 which is always TLS, I didn't add the parameter here. But it makes sense, so I should add it, now or in the future. Do you think it's better to add right away in this PR?

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
CarnaViire: You mean passing them from the outside of the app? We might want to add that in the future as low pri, but in the beginning we were thinking of just hard-coding/generating them inside a scenario, possibly based on some parameter.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
stephentoub: I think it depends on what our goals are for the benchmark.  If it's to highlight best-case scenario that a highly specialized consumer like YARP will use, we could use HttpMessageInvoker.  If it's to highlight expected case for the vast majority of consumers, we should stick with HttpClient.  CancelPendingRequests is a good example a non-pay-for-play feature we should avoid adding in the future.

geoffkizer: We should start with scenarios that represent the "expected case for the vast majority of customers". This is the most valuable data since it best represents the performance of HttpClient in practice.

We can and should measure other, less common/more advanced scenarios or variations as well, later.

scenarios/README.md
geoffkizer: I think non-TLS scenarios are so basic and relevant that they should be added in this PR.

scenarios/README.md
geoffkizer: When measuring throughput (as we are here) we generally want to max out CPU on the thing being measured, i.e. httpClient. This serves several purposes: most importantly, it's basically meaningless to compare benchmark throughput results if the CPU Is not maxed. Also, if we can't max the CPU then that usually indicates some sort of contention problem.

Note that when we run ASP.NET server benchmarks (e.g. TechEmpower) we don't care whether the WRK job hits max CPU or not. In fact it's generally better if it doesn't.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
geoffkizer: It's definitely interesting but doesn't need to be included in this PR.

src/BenchmarksApps/HttpClientBenchmarks/Servers/Kestrel/Program.cs
CarnaViire: Ah right, there's `Application started. Press Ctrl+C to shut down.`. It was in the middle of other default logs so I missed that. And my `readyStateText` said something else at first, so it wasn't working, and only after I got it working I changed it back to "Application started" 😄

scenarios/httpclient.benchmarks.yml
sebastienros: If I haven't answered somewhere else, it's the order in the scenario. And `waitForExit` is blocking the next one from starting until it has finished. That's why a web app doesn't have `waitForExit: true`, once it's ready the next service starts.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,268 @@
...
+
+namespace HttpClientBenchmarks;
+
+class Program
+{
+    private static ClientOptions s_options = null!;
+    private static List<HttpClient> s_httpClients = new();
+    private static Metrics s_metrics = new();
+    private static bool s_isWarmup;
+    private static bool s_isRunning;
```
antonfirsov, line 15: Probably not a big issue, but you may want to consider some synchronization around these variables, at least by making them [`volatile`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/volatile). (Even better if you can avoid them, by making warmup a separate step.)

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,268 @@
...
+            }
+            
+            try
+            {
+                stopwatch.Restart();
+                using var result = await client.GetAsync(uri, HttpCompletionOption.ResponseHeadersRead);
+                var headersTime = stopwatch.ElapsedTicks;
+                if (result.IsSuccessStatusCode)
+                {
+                    var content = await result.Content.ReadAsStreamAsync();
```
antonfirsov, line 172:
```suggestion
                    using var content = await result.Content.ReadAsStreamAsync();
```

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
CarnaViire: They were two separate steps before, but I merged them together as per @geoffkizer's comment https://github.com/aspnet/Benchmarks/pull/1706#discussion_r746996563

Some time ago, I've discussed `volatile` and `Interlocked` and other synchronization for a flag set by one thread and read by another thread with @stephentoub -- and the bottom line was that it's not really going to help with anything.
1) `volatile` would help only with ordered assignments -- if there are two flags and I depend on their order of assignment in the other thread(s).
2) `Interlocked` is only needed if there could be different threads assigning the same variable.
So neither of them would help me see the assigned variable "faster" in the other thread. It's a boolean, so there shouldn't be a "broken" read. And I don't care that much to see the value as soon as possible - I measure the time separately inside the tasks, not globally outside them, so as soon as the value change is seen in a reasonable time and the loop is stopped, I should be fine.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,268 @@
...
+                    var bytesRead = await content.ReadAsync(drainArray);
```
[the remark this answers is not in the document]
CarnaViire: It is not?.. the stopwatch is restarted right before GetAsync, on line 167. So all the timings are measured from the beginning of request - to the headers, to the start of content, to the end of content (=time of full request), that's why it doesn't restart in the middle.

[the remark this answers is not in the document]
antonfirsov: Sorry I wasn't clear, by "starting" I meant that `var contentStartTime = stopwatch.ElapsedTicks;` is only assigned at the next line, although at this line you already read some content. Or am I missing something in your metrics logic?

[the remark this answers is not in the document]
CarnaViire: ContentStart is meant to be a time when *a first byte* of content arrives. At first, I've even read it to a 1 byte buffer. However, if more than 1 byte is already available, it makes more sense to read into a bigger buffer -- and if it's only 1 byte that arrived, having bigger buffer will not affect it, as the code shouldn't wait for the rest and just put in the buffer whatever's available. And reading to a bigger buffer right away will make ContentEnd later more fair (e.g. if content is small and arrives in one go, we will not be splitted into 2 reads as with a separate 1 byte buffer)

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,268 @@
...
+            }
+        }
+        var elapsed = durationStopwatch.Elapsed.TotalSeconds;
+        metrics.MeanRps = (metrics.SuccessRequests +  metrics.BadStatusRequests) / elapsed;
+        return metrics;
+    }
+
+    private static void Log(string message)
+    {
+        var time = DateTime.Now.ToString("hh:mm:ss.fff");
```
ManickaP, line 208: NIT: Is the usage of `Now` intentional? Doesn't it print PCT time?

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,268 @@
...
+        var fractionPart = i - Math.Truncate(i);
+
+        return (1.0 - fractionPart) * sortedValues[(int)Math.Truncate(i) - 1] + fractionPart * sortedValues[(int)Math.Ceiling(i) - 1];
+    }
+
+    private static void RegisterPercentiledMetric(string name, string shortDescription, string longDescription)
+    {
+        BenchmarksEventSource.Register(name + "/min", Operations.Min, Operations.Min, shortDescription + " - min", longDescription + " - min", "n2");
+        BenchmarksEventSource.Register(name + "/p75", Operations.Max, Operations.Max, shortDescription + " - p75", longDescription + " - 75th percentile", "n2");
+        BenchmarksEventSource.Register(name + "/p50", Operations.Max, Operations.Max, shortDescription + " - p50", longDescription + " - 50th percentile", "n2");
```
ManickaP, line 239: Swap to keep the order 50, 75?

src/BenchmarksApps/HttpClientBenchmarks/Servers/Kestrel/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+                X509Certificate2 cert = certReq.CreateSelfSigned(DateTimeOffset.UtcNow.AddMonths(-1), DateTimeOffset.UtcNow.AddMonths(1));
+                if (OperatingSystem.IsWindows())
+                {
+                    cert = new X509Certificate2(cert.Export(X509ContentType.Pfx));
+                }
+                listenOptions.UseHttps(cert);
+            }
+            listenOptions.Protocols = HttpProtocols.Http1AndHttp2AndHttp3;
+        }
+        // in case https is not used, specific protocol (http1/http2) should be passed via command line argument (--Kestrel:EndpointDefaults:Protocols) because of h2c limitations
```
ManickaP, line 33: Shouldn't we be validating that then?

ManickaP (commented):
Have you tried to get rid of all the metrics and run it without it? I.e. only collect number of processed requests and that's it. To just see how much the metrics collection affects the final numbers.

I'd also try to run it with lower parallelism than 200, to see if we're not wasting cycles on context switching too much.

Otherwise I haven't seen anything radically wrong with the benchmark that would prevent it to reach the full saturation.

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,189 @@
...
+imports:
+  - https://github.com/aspnet/Benchmarks/blob/main/scenarios/aspnet.profiles.yml?raw=true
+
+variables:
+    serverPort: 5000
+    httpVersion: "1.1"
+    useHttps: true
+    numberOfHttpClients: 1
+    concurrencyPerHttpClient: 200
+    http11MaxConnectionsPerServer: "-1"
```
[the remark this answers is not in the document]
CarnaViire: so it would not conflict with argument parsing; anyway, I've changed it to 0 as per your suggestion below

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,189 @@
...
+
+jobs:
+  kestrel:
+    source:
+      repository: https://github.com/CarnaViire/benchmarks.git #https://github.com/aspnet/benchmarks.git
+      #localFolder: C:\Users\knatalia\dev\git\Benchmarks
+      branchOrCommit: http3-benchmarks #main
+      project: src/BenchmarksApps/HttpClientBenchmarks/Servers/Kestrel/Kestrel.csproj
+    readyStateText: Application started # app should write this line to output -- default Kestrel logging does that
+    arguments: "--urls http{% if useHttps == true %}s{% endif %}://{{serverAddress}}:{{serverPort}} --Kestrel:EndpointDefaults:Protocols {% if httpVersion == '2.0' %}http2{% else %}http1{% endif %}"
```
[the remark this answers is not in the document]
CarnaViire: done

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,189 @@
...
+      job: httpClient
+      variables:
+        scenario: get
+
+jobs:
+  kestrel:
+    source:
+      repository: https://github.com/CarnaViire/benchmarks.git #https://github.com/aspnet/benchmarks.git
+      #localFolder: C:\Users\knatalia\dev\git\Benchmarks
+      branchOrCommit: http3-benchmarks #main
```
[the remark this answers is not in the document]
CarnaViire: I will change it at the very last moment before merging into main; otherwise this yml will not work as the code only exist in my branch

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,189 @@
...
+      project: src/BenchmarksApps/HttpClientBenchmarks/Servers/Kestrel/Kestrel.csproj
+    readyStateText: Application started # app should write this line to output -- default Kestrel logging does that
+    arguments: "--urls http{% if useHttps == true %}s{% endif %}://{{serverAddress}}:{{serverPort}} --Kestrel:EndpointDefaults:Protocols {% if httpVersion == '2.0' %}http2{% else %}http1{% endif %}"
+
+  httpClient:
+    source:
+      repository: https://github.com/CarnaViire/benchmarks.git #https://github.com/aspnet/benchmarks.git
+      #localFolder: C:\Users\knatalia\dev\git\Benchmarks
+      branchOrCommit: http3-benchmarks #main
```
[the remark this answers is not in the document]
CarnaViire: same as above

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,189 @@
...
+      project: src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/HttpClient.csproj
+    isConsoleApp: true
+    waitForExit: true
+    arguments: "--url http{% if useHttps == true %}s{% endif %}://{{serverAddress}}:{{serverPort}} --scenario {{scenario}} --httpVersion {{httpVersion}} --numberOfHttpClients {{numberOfHttpClients}} --concurrencyPerHttpClient {{concurrencyPerHttpClient}} --http11MaxConnectionsPerServer {{http11MaxConnectionsPerServer}} {% if http20EnableMultipleConnections %} --http20EnableMultipleConnections {% endif %} --warmup {{warmup}} --duration {{duration}}"
```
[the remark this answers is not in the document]
CarnaViire: done

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,268 @@
...
+        BenchmarksEventSource.Register("ProcessorCount", Operations.First, Operations.First, "Processor Count", "Processor Count", "n0");
+        LogMetric("ProcessorCount", Environment.ProcessorCount);
+
+        for (int i = 0; i < s_options.NumberOfHttpClients; ++i)
+        {
+            var handler = new SocketsHttpHandler() 
+            {
+                // accept all certs
+                SslOptions = new SslClientAuthenticationOptions { RemoteCertificateValidationCallback = delegate { return true; } },
+                MaxConnectionsPerServer = s_options.Http11MaxConnectionsPerServer > 0 ? s_options.Http11MaxConnectionsPerServer : int.MaxValue,
```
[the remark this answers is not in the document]
CarnaViire: done

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,268 @@
...
+        {
+            await Task.Delay(TimeSpan.FromSeconds(s_options.Warmup));
+            s_isWarmup = false;
+            Log("Completing warmup...");
+            await Task.Delay(TimeSpan.FromSeconds(s_options.Duration));
+            s_isRunning = false;
+            Log("Completing scenario...");
+        });
+
+        var cts = new CancellationTokenSource(TimeSpan.FromSeconds(s_options.Warmup + s_options.Duration) + TimeSpan.FromMinutes(10));
```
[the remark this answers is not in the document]
CarnaViire: In the end, I've found there is a `timeout` option in job's configuration so I moved it there. The logic behind it is like "long running test" logic in CI, so it's just an arbitrary big number.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,268 @@
...
+                    var contentStartTime = stopwatch.ElapsedTicks;
+                    while (bytesRead != 0)
+                    {
+                        bytesRead = await content.ReadAsync(drainArray);
+                    }
+                    var contentEndTime = stopwatch.ElapsedTicks;
+
+                    if (!isWarmup)
+                    {
+                        metrics.HeadersTimes.Add(headersTime);
```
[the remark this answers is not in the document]
CarnaViire: done

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,189 @@
...
+
+imports:
+  - https://github.com/aspnet/Benchmarks/blob/main/scenarios/aspnet.profiles.yml?raw=true
+
+variables:
+    serverPort: 5000
+    httpVersion: "1.1"
+    useHttps: true
+    numberOfHttpClients: 1
+    concurrencyPerHttpClient: 200
```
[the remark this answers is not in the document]
CarnaViire: While experimenting, I've found out that having anything other than 1 as a default is actually not convenient. E.g. `concurrencyPerHttpClient` doesn't make sense with WRK, so it will be plainly ignored, but while comparing WRK with HttpClient you want to be sure that you are benchmarking the same thing. The same is with comparing "1 client x N threads" vs "N clients x 1 thread", easier to override 1 variable instead of 2.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
CarnaViire: Changed to UtcNow

src/BenchmarksApps/HttpClientBenchmarks/Servers/Kestrel/Program.cs
CarnaViire: reworked server options, as discussed above + added validations

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,228 @@
...
+      variables:
+        connections: '{{numberOfHttpClients}}'
+        threads: '{{numberOfHttpClients}}'
+        serverScheme: 'http{% if useHttps %}s{% endif %}'
+        path: /get
+        pipeline: 0
+        # TODO: Job's default values seem to take preference over global variables, so if the following is removed, `serverPort: 5000` will be used, etc. 
+        # Also, substitution doesn't seem to recognize global variables of the same name, e.g. serverPort: '{{serverPort}}' doesn't work
+        # Leaving the copied values unless the way to substitute correctly is found (other configs seem to employ copying strategy)
+        # Overriding via console arg (e.g. `--variable serverPort=4321`) will work for *both* global and job's variables
```
CarnaViire, line 46: @sebastienros do you know if it's possible to make substitutions/overrides work from "global" variables on the top to the job's variables of the same name? As I mentioned above, `serverPort: '{{serverPort}}'` is just left out as a text, like it is not resolved, and it also doesn't seem to understand `{{variables.serverPort}}`

ManickaP (approved):
LGTM, thanks! Super happy to have this!

scenarios/httpclient.benchmarks.yml
sebastienros: You don't need to do `serverPort: '{{serverPort}}'`, variables are merged with a specific order. So not defining it on the job will just have the value from the global variables.
Then profiles and finally command line variables will override these values.

CarnaViire: Unfortunately, it does not. The order of merging is super confusing. If I remove the assignment `serverPort: 1234`, I see `serverPort: 5000` from default WRK config being used, not my `serverPort: 1234` from global variables.

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,228 @@
...
+      job: httpClient
+      variables:
+        scenario: get
+  wrk-kestrel-get:
+    server:
+      job: kestrel
+    client:
+      job: wrk
+      variables:
+        connections: '{{numberOfHttpClients}}'
```
sebastienros, line 38: Isn't it equivalent to `concurrencyPerHttpClient` instead? 
A single instance of `HttpClient` is able to pool multiple connections, which as assume is equal to `concurrencyPerHttpClient`? This would be what `connections` means for wrk.

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,228 @@
...
+# crank --config "C:\Users\knatalia\dev\git\Benchmarks\scenarios\httpclient.benchmarks.yml" --scenario httpclient-kestrel-get --profile local
+
+imports:
+  - https://github.com/aspnet/Benchmarks/blob/main/scenarios/aspnet.profiles.yml?raw=true
+  - https://raw.githubusercontent.com/dotnet/crank/main/src/Microsoft.Crank.Jobs.Wrk/wrk.yml
+
+variables:
+    warmup: 15 #seconds -- default value *copied* to WRK, see notes below
+    duration: 30 #seconds -- default value *copied* to WRK
+    serverPort: 1234 # -- default value *copied* to WRK
```
sebastienros, line 11: Can it be 5000 like other benchmarks?

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,228 @@
...
+    # HttpClient-specific parameters:
+    http11MaxConnectionsPerServer: 0 #unlimited
+    http20EnableMultipleConnections: true
+    useWinHttpHandler: false
+    useHttpMessageInvoker: false
+
+scenarios:
+  httpclient-kestrel-get:
+    server:
+      job: kestrel
```
sebastienros, line 27: Can you try to use the new `agent: load` property (same level as job) so it binds to the `load` node in the existing profiles? This way we don't need a new profile. So when I start changing the other benchmarks too we can rename them with `primary/secondary/...` names for the machines in the profiles.

scenarios/httpclient.benchmarks.yml
```diff
@@ -0,0 +1,228 @@
...
+    client:
+      job: httpClient
+      variables:
```
sebastienros, line 30: `agent: application`

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,321 @@
...
+        }
+
+        await coordinatorTask;
+        var metricsArray = await Task.WhenAll(tasks);
+        foreach (var metrics in metricsArray)
+        {
+            s_metrics.Add(metrics);
+        }
+
+        LogMetric("http/successrequests", s_metrics.SuccessRequests);
```
sebastienros, line 174: Would be helpful if it could use the same metrics as the other clients, because I can then mix the values from different clients in the charts on PowerBI.

List from wrk for instance:

```
http/rps/mean
http/requests
http/latency/mean
http/latency/max
http/requests/badresponses
http/requests/errors
http/throughput
http/latency/50, http/latency/75, http/latency/90, http/latency/99
```

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,321 @@
...
+            LogPercentiledMetric("http/contentstart", s_metrics.ContentStartTimes, TicksToMs);
+            LogPercentiledMetric("http/contentend", s_metrics.ContentEndTimes, TicksToMs);
+        }
+    }
+
+    private static async Task<Metrics> Get(HttpMessageInvoker client)
+    {
+        var uri = new Uri(s_options.Url + "/get");
+
+        var drainArray = new byte[81920];
```
sebastienros, line 191: Is this allocating for each request?

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
```diff
@@ -0,0 +1,321 @@
...
+    private static HttpRequestMessage CreateRequest(HttpMethod method, Uri uri) =>
+        new HttpRequestMessage(method, uri) { Version = s_options.HttpVersion!, VersionPolicy = HttpVersionPolicy.RequestVersionExact };
+
+    private static void Log(string message)
+    {
+        var time = DateTime.UtcNow.ToString("hh:mm:ss.fff");
+        Console.WriteLine($"[{time}] {message}");
+    }
+
+    private static double TicksToMs(double ticks) => ticks * 1000.0 / Stopwatch.Frequency;
```
sebastienros, line 268: `1000.0 / Stopwatch.Frequency` could be a static field and processed once.

sebastienros (commented):
None of my comments are a blocker for merging. You can decide to tackle them later if you prefer.

scenarios/httpclient.benchmarks.yml
CarnaViire: I don't think so. The terminology in my mind is as follows: *HTTP client* is an entity that allows you to send HTTP requests, and several *HTTP Client*s will be fully independent one from another. In that sense, `System.Net.Http.HttpClient` is an *HTTP client*, but the future LLHTTPConnection would also be an *HTTP client*, and WRK's connection is also an *HTTP client*. 

That way, `numberOfHttpClients` would be a number of guaranteedly independent entities. And `concurrencyPerHttpClient` would be a number of threads accessing a single entity, and whether it supports concurrent access and how well it supports it is up to the implementation and configuration, and could/should be benchmarked. Note that `System.Net.Http.HttpClient` can be configured in a way that it will have only 1 connection, so all the threads will be competing to access it. And even if not limited, there is no way to make the connections inside `HttpClient` fully independent.

Unfortunately, WRK doesn't seem to support accessing one connection from several threads (it was more about handling several connections in one thread). That's why I set number of threads in WRK to always be equal to number of connections.

src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs
CarnaViire: No, it's per thread (the method contains a loop of requests below)

scenarios/httpclient.benchmarks.yml
sebastienros: > Note that System.Net.Http.HttpClient can be configured in a way that it will have only 1 connection

In the PR, `numberOfHttpClients` is used for the `connections` property in wrk. When set it to `1` it will create exactly 1 connection to the server, with one wrk thread, so I assume a max concurrency level of 1. However the default value for `http11MaxConnectionsPerServer` is "unlimited". Running it with HttpClient should create `concurrencyPerHttpClient` physical connections at max. Why use the same variables in wrk when they don't represent the same thing, while leaving `concurrencyPerHttpClient` out of wrk when this impact how many connections can be created. Why not set wrk 's `connections` property to `numberOfHttpClients | times:  concurrencyPerHttpClient` (the syntax in liquid for multiplications)? Even if the threading model is different and could be a fixed value based on the numbers of cores of the machine (it's available in the profiles). TE for instances fixes it to 32 thread.

I would assume that the wrk scenario is defined to have a baseline for httpclient, but if one scenario create 200 connections with a set of variables, and the other a single one, I don't see how it's fair.

CarnaViire: I agree that currently it will be wrong to call WRK when `concurrencyPerHttpClient` is not 1. Having the value to be `numberOfHttpClients` x `concurrencyPerHttpClient` will be great actually. But I wasn't able to make it work the way you suggest (I get parser errors). I will reach out to you offline.

CarnaViire: (other way could be to have some kind of validation that concurrencyPerHttpClient=1, I wonder if it is possible with the syntax)

CarnaViire: // covered that offline, I'll do the change

CarnaViire:
JFYI. I'm OOF till the end of the week, will be back to this PR later

sebastienros:
LGTM. In another PR I'd like to move away from the local profiles and use the new alias feature, but it can wait for me to first migrate the other benchmarks.

sebastienros:
Let me know when you want me to merge (some urls needs to target 'main'). Or if you have access then squash-merge yourself.

CarnaViire:
The PR is ready to be merged @sebastienros (I don't have access)

PR #1715, opened 2022-01-13
Fix HttpClient citrine-amd machine serverAddress

MihaZupan:


sebastienros:
I believe that is intentional since in this benchmark what is called "application" is actually the thing that we don't care about, but the "load" is benchmarked and has to run on the target machine (AMD, ARM, ...).

There is a new feature in crank to map job names to custom machines in an environment so we don't need multiple profiles to handle this case, but I haven't configured this file yet.

MihaZupan:
Actually nvm, this seems to be working.
I probably confused the two when testing with the AMD machine as the destination.

PR #1722, opened 2022-02-08
[Mono] Fix llvm-jit mode

fanyang-mono:
Forgot to turn on LLVM for llvm-jit mode.

cc: @EgorBo

### DIFF
--- a/build/mono-scenarios.yml
+++ b/build/mono-scenarios.yml
@@ -36,9 +36,9 @@ parameters:
     - displayName: "Core CLR"
       arguments: --property runtime=coreclr
     - displayName: "Mono - JIT"
-      arguments: --application.useMonoRuntime jit --property runtime=mono-jit
+      arguments: --application.useMonoRuntime jit --property runtime=mono-jit --application.environmentVariables MONO_ENV_OPTIONS="--llvm"
     - displayName: "Mono - JIT preemptive"
-      arguments: --application.useMonoRuntime jit --property runtime=mono-jit-preemptive --application.environmentVariables MONO_THREADS_SUSPEND=preemptive
+      arguments: --application.useMonoRuntime jit --property runtime=mono-jit-preemptive --application.environmentVariables MONO_THREADS_SUSPEND=preemptive --application.environmentVariables MONO_ENV_OPTIONS="--llvm"
     - displayName: "Mono - LLVM JIT"
       arguments: --application.useMonoRuntime llvm-jit --property runtime=mono-llvm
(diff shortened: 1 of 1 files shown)


build/mono-scenarios.yml
```diff
@@ -36,9 +36,9 @@ parameters:
     - displayName: "Core CLR"
       arguments: --property runtime=coreclr
     - displayName: "Mono - JIT"
-      arguments: --application.useMonoRuntime jit --property runtime=mono-jit
+      arguments: --application.useMonoRuntime jit --property runtime=mono-jit --application.environmentVariables MONO_ENV_OPTIONS="--llvm"
```
sebastienros, line 39: Is that to enable LLVM?

fanyang-mono: oops, wrong line. Let me fix it.

EgorBo:
🎉

EgorBo:
image

sebastienros:
Was the regression expected, looks like from 6 to 7, even if the ENV needed to be present before that.

EgorBo:
> Was the regression expected, looks like from 6 to 7, even if the ENV needed to be present before that.

well it wasn't really expected but it's identified, a change to coreclr/libraries broke vectorization for non-LLVM Mono (and even for Mono-LLVM for some parts) - https://github.com/dotnet/runtime/issues/61484

EgorBo:
@sebastienros any idea why there are no mono runs since 2/8/2022?

sebastienros:
All LLVM benchmarks are failing. There is a generic error message. @fanyang-mono can you try to run one of the LLVM ones and see what the errors are?

EgorBo:
> All LLVM benchmarks are failing. There is a generic error message. @fanyang-mono can you try to run one of the LLVM ones and see what the errors are?

```
crank  --profile aspnet-citrine-lin --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/platform.benchmarks.yml  --scenario plaintext --application.options.displayOutput true --application.useMonoRuntime llvm-jit --property runtime=mono-llvm --application.environmentVariables MONO_ENV_OPTIONS="--llvm"
```
I see no errors

sebastienros:
You are missing `--framework 7.0 --application.options.collectCounters true` to reproduce what is actually running on the CI. However it also works for me. Any AOT one though will fail with an error that Fan should look into (dedicated mono support in the agent).
I should have more logs tomorrow morning about why this llvm-jit one is not showing up though it's working for us (it's Monday and the logs from last week got deleted last night).

PR #1724, opened 2022-02-15
HttpClient Benchmark - add response size configuration

CarnaViire:
Add response size configuration

/cc @ManickaP @MihaZupan @geoffkizer @scalablecory @wfurt @karelz

### DIFF
--- a/src/BenchmarksApps/HttpClientBenchmarks/Servers/Kestrel/Program.cs
+++ b/src/BenchmarksApps/HttpClientBenchmarks/Servers/Kestrel/Program.cs
@@ -10,6 +10,10 @@ class Program
 {
     private static ServerOptions s_options = null!;
 
+    private static byte[] s_data10b = new byte[10];
+    private static byte[] s_data10k = new byte[10 * 1024];
+    private static byte[] s_data10m = new byte[10 * 1024 * 1024];
+
     public static async Task<int> Main(string[] args)
     {
         var rootCommand = new RootCommand();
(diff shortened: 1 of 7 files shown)


src/BenchmarksApps/HttpClientBenchmarks/Servers/Kestrel/Program.cs
```diff
@@ -80,17 +94,29 @@ public static void RunKestrel()
...
         {
             await request.Body.CopyToAsync(Stream.Null);
-            return Results.Text("Server received the content");
+            return GetResponse(responseSize);
         });
 
         app.Run();
     }
 
+    private static IResult GetResponse(string responseSize)
```
MihaZupan, line 108: Instead of hardcoding 4 different options, can we match what the default downstream program does to enable arbitrary values?
https://github.com/aspnet/Benchmarks/blob/d5f47088cd6582ad35b44ffb5ef93c7c049bf822/src/Downstream/Program.cs#L86-L90

sebastienros: Right, a single buffer, initialized from the argument. Also does the content in the buffer matter? (randome seed).

CarnaViire: Yep, that was the other option in my mind. I don't feel strongly on one vs another. The content doesn't matter. I'll make the change

CarnaViire:
@ManickaP @MihaZupan PTAL 🙏

MihaZupan (approved):
This LGTM, thanks!

ManickaP (approved):
LGTM, thanks!

PR #1725, opened 2022-02-22
HttpClient benchmark - add generated headers

CarnaViire:
Add configuration to set number of generated headers with static values and number of generated headers with values changing per request.

Add flag to use `HttpClient.DefaultRequestHeaders` for static headers.

cc @ManickaP @MihaZupan

### DIFF
--- a/scenarios/README.md
+++ b/scenarios/README.md
@@ -433,6 +433,10 @@ crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenario
   - `--variable requestHeaders=none` (default)
   - `--variable requestHeaders=connectionclose`
   - `--variable requestHeaders=expectcontinue`
+- Number of generated request headers with static values:
+  - `--variable generatedStaticRequestHeadersCount=<N>` (default: `0`)
+- Number of generated request headers with values changing per each request:
+  - `--variable generatedDynamicRequestHeadersCount=<N>` (default: `0`)
 - Number of HTTP clients (HttpClient in `httpClient` job, connection in WRK):
   - `--variable numberOfHttpClients=<N>` (default: `1`)
 - Number of concurrect requests per one HTTP client -- *unsupported by WRK*:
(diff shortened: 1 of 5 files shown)


ManickaP (approved):
LGTM, thanks.
The only thing I'm wondering about is if we really need this level of configurability, e.g. whether to use default headers collection or not. On the other hand, the code for this is fairly simple so it doesn't incur high costs of maintenance. I'm just thinking aloud here, nothing serious though :smile:

CarnaViire:
DefaultRequestHeaders was what @MihaZupan actually suggested, saying that HttpClient+DefaultRequestHeaders should be faster than HttpMessageInvoker, so I was curious and wanted to be able to test that theory. Also would be handy if we would have LLHTTP later to compare to them too.

MihaZupan:
Right, `DefaultRequestHeaders` is the only case I know of where `HttpClient` can end up outperforming `HttpMessageInvoker`. It has a [more efficient copy loop](https://github.com/dotnet/runtime/blob/c4422a0a1687ddc0869565cbbc8459e1f0e11737/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HttpHeaders.cs#L528) than what you can achieve with `HttpMessageInvoker`.

And it's especially efficient for the `DynamicRequestHeadersCount = 0` case (i.e. all headers come from `DefaultRequestHeaders`) due to [this fast path we added in 7.0](https://github.com/dotnet/runtime/blob/c4422a0a1687ddc0869565cbbc8459e1f0e11737/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HttpHeaders.cs#L536-L555)

CarnaViire:
@sebastienros could you please take a look?

sebastienros:
Look or merge?

CarnaViire:
If you're ok with it then merge -- I have @ManickaP's approve above

PR #1726, opened 2022-03-07
add osr to pgo modes

AndyAyersMS:
We are getting close to enabling OSR by default in the runtime. Want to verify that things are working as expected.

cc @sebastienros


sebastienros:
The new options should appear automatically in the dashboard once it has run

PR #1728, opened 2022-04-15
Use agent names

sebastienros:
@CarnaViire Could your try this file locally and tell me if it's working like before the change? This way we don't have to copy the profile, we can just assign each job on the machine role we want.

### DIFF
--- a/scenarios/httpclient.benchmarks.yml
+++ b/scenarios/httpclient.benchmarks.yml
@@ -42,16 +42,20 @@ scenarios:
   httpclient-kestrel-get:
     server:
       job: kestrel
+      agent: load
     client:
       job: httpClient
+      agent: application
       variables:
         scenario: get
   
   httpclient-kestrel-post:
     server:
       job: kestrel
+      agent: load
     client:
       job: httpClient
+      agent: application
       variables:
         scenario: post
         requestContentSize: 1024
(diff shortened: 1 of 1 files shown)


CarnaViire:
Sorry for the delay @sebastienros and thanks for the change!
I've checked it, and there is a problem: while the machines are picked up correctly, the `serverAddress` variable is not updated, so it still has `application` machine IP, but the server is being run on `load` machine 😢

CarnaViire:
Sorry for the delay @sebastienros and thanks for the change!
I've checked it, and there is a problem: while the machines are picked up correctly, the `serverAddress` variable is not updated, so it still has `application` machine IP, but the server is being run on `load` machine 😢

sebastienros:
Can you share a command line I could try to debug?

CarnaViire:
1) the commit before the change -- works, the IP used is 10.0.0.104
crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/f544bbe8bf89aa29bf71861ff2790dee869dcc35/scenarios/httpclient.benchmarks.yml --scenario httpclient-kestrel-get --profile aspnet-perf-lin --variable warmup=3 --variable duration=3

2) the commit after the change -- doesn't work, the IP used is 10.0.0.102 (wrong)
crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/dc2188505a1dcaf2b6cfaaaf7673651e89de36ad/scenarios/httpclient.benchmarks.yml --scenario httpclient-kestrel-get --profile aspnet-perf-lin --variable warmup=3 --variable duration=3

sebastienros:
Closing since you have mastered this now

PR #1735, opened 2022-05-13
Add WebSocket compression and size flags

BrennanConroy:
I know the options work when manually running things from the command line, and I hope I wired everything up correctly for CI builds.

### DIFF
--- a/build/websocket-scenarios.yml
+++ b/build/websocket-scenarios.yml
@@ -11,32 +11,48 @@ parameters:
   type: string
   default: 'true'
 
+- name: sizes
+  type: object
+  default:
+    - displayName: 10 bytes
+      arguments: --variable size=10 --property size=10
+    - displayName: 100 bytes
+      arguments: --variable size=100 --property size=100
+    - displayName: 1000 bytes
+      arguments: --variable size=1000 --property size=1000
+    - displayName: 10000 bytes
+      arguments: --variable size=10000 --property size=10000
+
+- name: compress
+  type: object
+  default:
+    - displayName: with-compression
+      arguments: --variable compress=true --application.buildArguments /p:UseCompression=true --property compress=true
+    - displayName: without-compression
+      arguments: --variable compress=false --property compress=false
+
 # Scenarios
 - name: scenarios 
   type: object
   default: 
 
-  - displayName: Echo-10000
-    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --variable size=10000--property scenario=echo
-  - displayName: Echo-100
-    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --variable size=100 --property scenario=echo
-  - displayName: Echo-Compression-100
-    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --variable size=100 --variable compress=true --property scenario=echo --application.buildArguments \"/p:UseCompression=true\"
-  - displayName: Echo-Compression-10000
-    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --variable size=10000 --variable compress=true --property scenario=echo --application.buildArguments \"/p:UseCompression=true\"
+  - displayName: Echo
+    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --property scenario=echo
 
 steps:
 - ${{ each s in parameters.scenarios }}:
-  - task: PublishToAzureServiceBus@1
-    condition: succeededOrFailed()
-    displayName: ${{ s.displayName }}
-    inputs:
-      connectedServiceName: ${{ parameters.connection }}
-      waitForCompletion: true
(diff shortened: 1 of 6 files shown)


build/websocket-scenarios.yml
```diff
@@ -16,8 +16,14 @@ parameters:
   type: object
   default: 
 
-  - displayName: Echo
-    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --property scenario=echo
+  - displayName: Echo-10000
+    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --variable size=10000--property scenario=echo
```
sebastienros, line 20:
```suggestion
    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --variable size=10000 --property scenario=echo --property size=1000 --property compress=false
```

Space, but also scenario is a metadata in the results, used by power bi only. I assume you want to expose "size" as a different list, but we could also change the "scenario" metadata to contain the size (echo-1000). You choose (not separate is always better as we can still concatenate in PBI)

src/WebsocketClient/websocketclient.yml
```diff
@@ -16,4 +16,6 @@
       serverAddress: localhost
       serverPort: 5000
       collectLatency: true
-    arguments: "-c {{connections}} -w {{warmup}} -d {{duration}} -s {{scenario}} -u {% if serverUri == blank or serverUri == empty %}{{serverScheme}}://{{serverAddress}}:{{serverPort}}{{path}}{% else %}{{serverUri}}:{{serverPort}}{{path}}{% endif %}{% if collectLatency != blank %} --latency {{ collectLatency }}{% endif %}"
+      compress: false
+      size: 1000
+    arguments: "-c {{connections}} -w {{warmup}} -d {{duration}} -s {{scenario}} -u {% if serverUri == blank or serverUri == empty %}{{serverScheme}}://{{serverAddress}}:{{serverPort}}{{path}}{% else %}{{serverUri}}:{{serverPort}}{{path}}{% endif %}{% if collectLatency != blank %} --latency {{ collectLatency }}{% endif %} --size {{size}} --compress {{compress}}"
```
sebastienros, line 21:
```suggestion
    arguments: "-c {{connections}} -w {{warmup}} -d {{duration}} -s {{scenario}} -u {{serverScheme}}://{{serverAddress}}:{{serverPort}}{{path}}{% if collectLatency != blank %} --latency {{ collectLatency }}{% endif %} --size {{size}} --compress {{compress}}"
```

Just removing some old arguments check that is not used anymore

build/websocket-scenarios.yml
```diff
@@ -16,8 +16,14 @@ parameters:
...
+  - displayName: Echo-100
+    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --variable size=100 --property scenario=echo
```
sebastienros, line 22:
```suggestion
    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --variable size=100 --property scenario=echo --property size=100 --property compress=false
```

build/websocket-scenarios.yml
```diff
@@ -16,8 +16,14 @@ parameters:
...
+  - displayName: Echo-Compression-100
+    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --variable size=100 --variable compress=true --property scenario=echo --application.buildArguments \"/p:UseCompression=true\"
```
sebastienros, line 24:
```suggestion
    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --variable size=100 --variable compress=true --property scenario=echo --application.buildArguments \"/p:UseCompression=true\" --property size=100 --property compress=true
```

Note that `\"` is only used because this string is concatenated in the json payload later. In Yaml we could just use `"` directly here otherwise.

build/websocket-scenarios.yml
```diff
@@ -16,8 +16,14 @@ parameters:
...
+  - displayName: Echo-Compression-10000
+    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --variable size=10000 --variable compress=true --property scenario=echo --application.buildArguments \"/p:UseCompression=true\"
```
sebastienros, line 26:
```suggestion
    arguments: --scenario websocket $(websocketJobs) --variable scenario=echo --variable size=10000 --variable compress=true --property scenario=echo --application.buildArguments \"/p:UseCompression=true\" --property size=10000 --property compress=true
```

You can also create other collections (100, 1000, 1000) and (true/false) to do embedded loops to generate all command lines. Check the containers benchmarks file for instance.

sebastienros (approved):

build/websocket-scenarios.yml
BrennanConroy: Changed to use loops, please take a look :)

PR #1744, opened 2022-07-26
adding composite json scenario.

mangod9:
Adding a new dockerfile to use a asp.net R2R composite container image.


DamianEdwards:
@mangod9 is this something you'd still like added? If so, what does this enable exactly?

PR #1747, opened 2022-09-09
Include dotnet\runtime in benchmark monitor

kunalspathak:
This will let us trigger `/benchmarks TE` benchmarks on dotnet/runtime repository. Currently, it is only for windows and once we add linux agent, will extend it for linux.

E.g.

```
/benchmark runtime-win-x64 aspnet-citrine-win-ampere plaintext,json
```


kunalspathak:
Tried locally and seems to be working.

image

PR #1749, opened 2022-09-18
Add 4 yml for win/linux/x64/arm64

kunalspathak:
Note that we will need linux agent to run the build jobs in linux pipeline.

### DIFF
--- a/build/prbenchmarks.runtime.linux_x64.config.yml
+++ b/build/prbenchmarks.runtime.linux_x64.config.yml
@@ -15,8 +15,16 @@ defaults: --config https://github.com/aspnet/Benchmarks/blob/main/build/azure.pr
 profiles:
 
     aspnet-citrine-lin:
-        description: Linux 28 Cores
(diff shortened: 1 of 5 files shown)


build/prbenchmarks.runtime.linux_x64.config.yml
```diff
@@ -0,0 +1,32 @@
...
+        arguments:
+            --application.options.outputFiles ./artifacts/tests/coreclr/Linux.x64.Release/tests/Core_Root/
+
+# default arguments that are always used on crank commands
+defaults: --config https://github.com/aspnet/Benchmarks/blob/main/build/azure.profile.yml?raw=true --application.framework net7.0 --relay AZURE_RELAY 
+
+# the first value is the default if none is specified
+profiles:
+
+    aspnet-citrine-lin:
```
kunalspathak, line 17: Are there more Linux/x64 profiles that we can add?

kunalspathak:
@sebastienros

PR #1754, opened 2022-10-04
Add HttpClient benchmarks to runtime PR benchmarks

CarnaViire:
Depends on https://github.com/dotnet/crank/pull/471

It can theoretically be done without the PR above, but it would require moving and copy-pasting everything to `benchmarks.arguments`, including what was in `defaults` and in `components.arguments`. HttpClient benchmarks don't have `application` job so Crank will fail if an argument with `application` would be passed there.

Let me know what you think @sebastienros

### DIFF
--- a/build/prbenchmarks.runtime.linux_x64.config.yml
+++ b/build/prbenchmarks.runtime.linux_x64.config.yml
@@ -12,8 +12,7 @@ components:
 defaults: '--{{job}}.framework net7.0 --relay AZURE_RELAY'
 
 variables:
-    job: client
-    httpclientcommon: --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/httpclient.benchmarks.yml --server.framework net7.0 --variable useHttps=true --variable concurrencyPerHttpClient=100
+    job: application
 
 # the first value is the default if none is specified
 profiles:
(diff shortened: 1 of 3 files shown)


sebastienros:
I am not sure we should have so many possibilities in the PR bots. That could lead our devs to use this as the main benchmarking vector. The PR bot is useful for light checks, and when the submitter can't benchmark (external users for instance). I'd rather see the person who made the PR do the benchmark on their change, probably with their local builds, and share the results in the PR directly. PR benchmark is slower to run, and can add maybe unnecessary noise to the comments, which might make some devs not super happy with the potential mess.

Technically I think this PR is totally fine, so it's just a team decision to decide what should be available here. It's also setting a precedent, and other devs might then add all the benchmarks that they could think of. A compromise IMO would be to add one/two benchmarks per domain (JIT, GC, Libs, ASPNET, HttpClient, ...) just for safety. And use more uncommon ones manually. Because if you need more than one to be run, then it will also be much faster and easier to start them locally.

You tell me if you want to keep it this way, I would merge then.

CarnaViire:
Thanks for providing such a perspective! I didn't think about it that way. I will discuss it with the team, and we'll decide, but I will minimize the number of the benchmarks in the list. I think we'd still want to have one per HTTP version, but we'll see.

PR #1757, opened 2022-10-13
Add standardized profiles

CarnaViire:
1. Add file with standardized profiles `aspnet.profiles.standard.yml`
2. Update HttpClient benchmarks to use that file
3. Add aliases to existing profiles (e.g. to `ci.profile.yml` so that it could be used for HttpClient benchmarks as well)
4. Fix wrong IPs in `aspnet.profiles.yml` (asp-citrine-amd IP is 10.0.0.106 ([link](https://github.com/aspnet/Benchmarks/blob/main/build/ci.profile.yml#L111-L118)) and asp-citrine-load IP is 10.0.0.105 ([link](https://github.com/aspnet/Benchmarks/blob/main/build/ci.profile.yml#L64-L73)))

cc @sebastienros


CarnaViire:
PTAL @sebastienros

sebastienros:
Looks great and seems backward compatible so we could try switching existing scenarios to these profiles. Also reminded me about the `agent` property, which fits perfectly with the refactoring.

CarnaViire:
> seems backward compatible so we could try switching existing scenarios to these profiles

Yes, that what I was going for, I wanted to do so in the follow-up PRs after that one

sebastienros:
And then remove these backward compatibility values as other scenarios are migrated to the new names.

CarnaViire:
Shall we merge this PR if you are OK with it @sebastienros?

PR #1758, opened 2022-10-14
Removing OSR modes

sebastienros:
@egorbo

### DIFF
--- a/build/pgo-scenarios.yml
+++ b/build/pgo-scenarios.yml
@@ -44,7 +44,7 @@ parameters:
   - displayName: Fortunes EF
     arguments: --scenario fortunes_ef $(databaseJobs) --property scenario=FortunesEf --property protocol=http
 
-# PGO and OSR modes
+# PGO modes
 - name: modes
   type: object
   default:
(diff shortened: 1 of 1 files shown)


EgorBo:
@sebastienros thanks!! can you please also add just regular tier0 without PGO? we want to be able to compare those two to track our progress in improving performance of Tier0-instr

sebastienros:
what are the ENVs for that mode, and what name should it use?

EgorBo:
> what are the ENVs for that mode, and what name should it use?

@sebastienros same as you've just added except `DOTNET_TieredPGO=1` (or better to explicitly use `DOTNET_TieredPGO=0`)

sebastienros:
1- Maybe the name is not the one you want
2- Can you check if the profile with `COMPlus_JitDisablePGO` is still valid? Can't find any reference of this ENV

build/pgo-scenarios.yml
```diff
@@ -53,16 +53,14 @@ parameters:
...
-  - displayName: Dynamic PGO OSR
-    arguments: --property pgo=dynamic-osr --application.environmentVariables DOTNET_TieredPGO=1 --application.environmentVariables DOTNET_TC_QuickJitForLoops=1 --application.environmentVariables DOTNET_TC_OnStackReplacement=1
   - displayName: Full PGO
     arguments: --property pgo=full --application.environmentVariables DOTNET_TieredPGO=1 --application.environmentVariables DOTNET_TC_QuickJitForLoops=1 --application.environmentVariables DOTNET_ReadyToRun=0
-  - displayName: Full PGO OSR
-    arguments: --property pgo=full-osr --application.environmentVariables DOTNET_TieredPGO=1 --application.environmentVariables DOTNET_TC_QuickJitForLoops=1 --application.environmentVariables DOTNET_ReadyToRun=0 --application.environmentVariables DOTNET_TC_OnStackReplacement=1
+  - displayName: Tier0 Instrumented
+    arguments: --property pgo=tier0-instrumented --application.environmentVariables DOTNET_ReadyToRun=0 --application.environmentVariables DOTNET_TieredPGO=1 --application.environmentVariables DOTNET_TC_CallCounting=0
+  - displayName: Tier0 Instrumented
```
EgorBo, line 62: Maybe just `Tier0` ? or `Tier0 Uninstrumented`

build/pgo-scenarios.yml
```diff
@@ -53,16 +53,14 @@ parameters:
...
   - displayName: No PGO
     arguments: --property pgo=disabled --application.environmentVariables COMPlus_JitDisablePGO=1
-  - displayName: OSR
-    arguments: --property pgo=osr --application.environmentVariables DOTNET_TC_QuickJitForLoops=1 --application.environmentVariables DOTNET_TC_OnStackReplacement=1
   - displayName: Dynamic PGO
     arguments: --property pgo=dynamic --application.environmentVariables DOTNET_TieredPGO=1 --application.environmentVariables DOTNET_TC_QuickJitForLoops=1
-  - displayName: Dynamic PGO OSR
-    arguments: --property pgo=dynamic-osr --application.environmentVariables DOTNET_TieredPGO=1 --application.environmentVariables DOTNET_TC_QuickJitForLoops=1 --application.environmentVariables DOTNET_TC_OnStackReplacement=1
   - displayName: Full PGO
     arguments: --property pgo=full --application.environmentVariables DOTNET_TieredPGO=1 --application.environmentVariables DOTNET_TC_QuickJitForLoops=1 --application.environmentVariables DOTNET_ReadyToRun=0
```
EgorBo, line 59: `DOTNET_TC_QuickJitForLoops=1` is not needed anymore since it's enabled by default now and we don't plan to change it

EgorBo (approved):
Thanks!!

EgorBo:
> 2- Can you check if the profile with COMPlus_JitDisablePGO is still valid? Can't find any reference of this ENV

Should be still valid in dotnet/runtime (it's here just to ignore static pgo and detect cases when we ship something terribly wrong in that pgo data)

sebastienros:
Updated

EgorBo:
:shipit:

PR #1761, opened 2022-10-22
gRPC AOT scenarios

JamesNK:
🔥 🚀 💣 🌶️ 

Depends on https://github.com/grpc/grpc-dotnet/pull/1924

### DIFF
--- a/build/grpc-scenarios.yml
+++ b/build/grpc-scenarios.yml
@@ -20,14 +20,19 @@ parameters:
   default:
 
   - displayName: "ASP.NET Core - Client: Grpc.Net.Client"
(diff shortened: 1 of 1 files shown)


sebastienros:
Need to check it will work fist since we reached the limit of the yaml size on Friday and had to remove some benchmark (memory consumption when loading the whole processed file is limited to 10MB)
Will probably have to split the pipeline in two first before accepting this one. Another option would be to move some predefined strings in other crank config files so we don't have to include their content in the azdo yaml files.

JamesNK:
Might do this in the future, but the current AOT test is fine for now.

PR #1765, opened 2022-10-31
Add TieredPGO_InstrumentOnlyHotCode=0 to tier0 jobs

EgorBo:
We recently merged https://github.com/dotnet/runtime/pull/70941 to runtime
It means that `pgo=tier0-instrumented` job no longer measures instrumented code and TieredPGO_InstrumentOnlyHotCode=0 should be added to bring it back. 

image

cc @sebastienros


sebastienros:
Lesson: we should keep setting the values even when they are the default.

EgorBo:
Yep, but in this case `TieredPGO_InstrumentOnlyHotCode=0` is a new thing - it didn't exist previously 🙂

PR #1766, opened 2022-10-31
Fix default scheme for MVC apps

BrennanConroy:
Broke default scenarios when I fixed the cert auth scenarios by not having a default serverScheme.


BrennanConroy:
Bump

PR #1772, opened 2022-11-29
Update Json middleware benchmark to use source-gen

BrennanConroy:
179m bytes per second to 3m
Working Set 440MB to 62MB
RPS change unknown right now since the baseline changed between running earlier in the day and now. But running them back-to-back right now gives a ~1.1% increase in RPS

### DIFF
--- a/src/Benchmarks/Benchmarks.csproj
+++ b/src/Benchmarks/Benchmarks.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
   <!-- For local dev specify all supported TFMs they can be easily tested -->
   <PropertyGroup Condition="'$(BenchmarksTargetFramework)' == ''">
     <TargetFrameworks>net7.0</TargetFrameworks>
--- a/src/Benchmarks/Benchmarks.csproj
+++ b/src/Benchmarks/Benchmarks.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+<Project Sdk="Microsoft.NET.Sdk.Web">
   <!-- For local dev specify all supported TFMs they can be easily tested -->
   <PropertyGroup Condition="'$(BenchmarksTargetFramework)' == ''">
     <TargetFrameworks>net7.0</TargetFrameworks>
--- a/src/Benchmarks/Middleware/JsonMiddleware.cs
+++ b/src/Benchmarks/Middleware/JsonMiddleware.cs
@@ -12,6 +12,7 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Options;
 using Microsoft.AspNetCore.Http.Json;
+using System.Text.Json.Serialization.Metadata;
 
 namespace Benchmarks.Middleware
 {
@@ -21,11 +22,16 @@ namespace Benchmarks.Middleware
         private const int _bufferSize = 27;
         private readonly RequestDelegate _next;
         private readonly JsonSerializerOptions _jsonOptions;
+        private readonly JsonTypeInfo _jsonTypeInfo;
 
         public JsonMiddleware(RequestDelegate next, IOptions<JsonOptions> jsonOptions)
         {
             _next = next;
+#if NET8_0_OR_GREATER
+            _jsonTypeInfo = jsonOptions.Value.SerializerOptions.GetTypeInfo(typeof(JsonMessage));
+#else
             _jsonOptions = jsonOptions.Value.SerializerOptions;
+#endif
         }
 
         public Task Invoke(HttpContext httpContext)
@@ -35,7 +41,13 @@ namespace Benchmarks.Middleware
                 httpContext.Response.StatusCode = 200;
                 httpContext.Response.ContentLength = _bufferSize;
 
-                return httpContext.Response.WriteAsJsonAsync(new JsonMessage { message = "Hello, World!" }, _jsonOptions, httpContext.RequestAborted);
+                return httpContext.Response.WriteAsJsonAsync(new JsonMessage { message = "Hello, World!" },
+#if NET8_0_OR_GREATER
+                    _jsonTypeInfo
+#else
+                    _jsonOptions
+#endif
+                    );
             }
 
             return _next(httpContext);
@@ -47,7 +59,11 @@ namespace Benchmarks.Middleware
         public static IApplicationBuilder UseJson(this IApplicationBuilder builder) => builder.UseMiddleware<JsonMiddleware>();
     }
 
+#if NET8_0_OR_GREATER
+    [JsonSourceGenerationOptions(JsonSerializerDefaults.Web)]
+#else
     [JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase)]
+#endif
     [JsonSerializable(typeof(JsonMessage))]
     internal partial class CustomJsonContext : JsonSerializerContext
     {
--- a/src/Benchmarks/Middleware/JsonMiddleware.cs
+++ b/src/Benchmarks/Middleware/JsonMiddleware.cs
@@ -4,6 +4,8 @@
 using System;
 using System.IO;
 using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 using System.Threading.Tasks;
 using Benchmarks.Configuration;
 using Microsoft.AspNetCore.Builder;
@@ -17,21 +19,24 @@ namespace Benchmarks.Middleware
     public class JsonMiddleware
     {
         private static readonly PathString _path = new PathString(Scenarios.GetPath(s => s.Json));
-        private static readonly UTF8Encoding _encoding = new UTF8Encoding(false);
         private const int _bufferSize = 27;
         private readonly RequestDelegate _next;
+        private readonly JsonSerializerOptions _jsonOptions;
 
-        public JsonMiddleware(RequestDelegate next) => _next = next;
+        public JsonMiddleware(RequestDelegate next, IOptions<JsonOptions> jsonOptions)
+        {
+            _next = next;
+            _jsonOptions = jsonOptions.Value.SerializerOptions;
+        }
 
         public Task Invoke(HttpContext httpContext)
         {
             if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal))
             {
                 httpContext.Response.StatusCode = 200;
-                httpContext.Response.ContentType = "application/json";
                 httpContext.Response.ContentLength = _bufferSize;
 
-                return JsonSerializer.SerializeAsync<JsonMessage>(httpContext.Response.Body, new JsonMessage { message = "Hello, World!" });
+                return httpContext.Response.WriteAsJsonAsync<JsonMessage>(new JsonMessage { message = "Hello, World!" }, _jsonOptions, httpContext.RequestAborted);
             }
 
             return _next(httpContext);
@@ -41,6 +46,12 @@ namespace Benchmarks.Middleware
     public static class JsonMiddlewareExtensions
     {
         public static IApplicationBuilder UseJson(this IApplicationBuilder builder) => builder.UseMiddleware<JsonMiddleware>();
+    }
+
+    [JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase)]
+    [JsonSerializable(typeof(JsonMessage))]
+    internal partial class CustomJsonContext : JsonSerializerContext
+    {
     }
 
     public struct JsonMessage
--- a/src/Benchmarks/Middleware/JsonMiddleware.cs
+++ b/src/Benchmarks/Middleware/JsonMiddleware.cs
@@ -21,14 +21,17 @@ namespace Benchmarks.Middleware
         private static readonly PathString _path = new PathString(Scenarios.GetPath(s => s.Json));
         private const int _bufferSize = 27;
         private readonly RequestDelegate _next;
+#if NET8_0_OR_GREATER
+        private readonly JsonTypeInfo<JsonMessage> _jsonTypeInfo;
+#else
         private readonly JsonSerializerOptions _jsonOptions;
-        private readonly JsonTypeInfo _jsonTypeInfo;
+#endif
 
         public JsonMiddleware(RequestDelegate next, IOptions<JsonOptions> jsonOptions)
         {
             _next = next;
 #if NET8_0_OR_GREATER
-            _jsonTypeInfo = jsonOptions.Value.SerializerOptions.GetTypeInfo(typeof(JsonMessage));
+            _jsonTypeInfo = JsonTypeInfo.CreateJsonTypeInfo<JsonMessage>(jsonOptions.Value.SerializerOptions);
 #else
             _jsonOptions = jsonOptions.Value.SerializerOptions;
 #endif
--- a/src/Benchmarks/Middleware/JsonMiddleware.cs
+++ b/src/Benchmarks/Middleware/JsonMiddleware.cs
@@ -31,7 +31,7 @@ namespace Benchmarks.Middleware
         {
             _next = next;
 #if NET8_0_OR_GREATER
-            _jsonTypeInfo = JsonTypeInfo.CreateJsonTypeInfo<JsonMessage>(jsonOptions.Value.SerializerOptions);
+            _jsonTypeInfo = jsonOptions.Value.SerializerOptions.GetTypeInfo(typeof(JsonMessage)) as JsonTypeInfo<JsonMessage>;
 #else
             _jsonOptions = jsonOptions.Value.SerializerOptions;
 #endif
--- a/src/Benchmarks/Middleware/JsonMiddleware.cs
+++ b/src/Benchmarks/Middleware/JsonMiddleware.cs
@@ -10,9 +10,8 @@ using System.Threading.Tasks;
 using Benchmarks.Configuration;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Features;
-using System.Text.Json;
-using System.Text.Json.Serialization;
+using Microsoft.Extensions.Options;
+using Microsoft.AspNetCore.Http.Json;
 
 namespace Benchmarks.Middleware
 {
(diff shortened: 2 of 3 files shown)


src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -53,7 +56,11 @@ public Task Invoke(HttpContext httpContext)
 
                 return Task.CompletedTask;
 #else
-                return JsonSerializer.SerializeAsync<JsonMessage>(httpContext.Response.Body, new JsonMessage { message = "Hello, World!" });
+                _writer ??= new Utf8JsonWriter(httpContext.Response.Body);
+                _writer.Reset(httpContext.Response.Body);
+
+                JsonSerializer.Serialize<JsonMessage>(_writer, new JsonMessage { message = "Hello, World!" }, CustomJsonContext.Default.JsonMessage);
```
DamianEdwards, line 62: I thought `JsonSerializer.Serialize` will internally pool the `Utf8JsonWriter` instance it uses if we don't pass one?

BrennanConroy: No https://source.dot.net/#System.Text.Json/System/Text/Json/Serialization/JsonSerializer.Write.Stream.cs,391
And also, the source-gen isn't called when using the stream overloads. https://github.com/dotnet/runtime/pull/78646

DamianEdwards: So which overloads do pool? Just those that return string?

BrennanConroy: > Just those that return string?

Looks like it

src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -10,7 +10,7 @@
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 
-#if NETCOREAPP3_0 || NETCOREAPP3_1 || NETCOREAPP5_0 || NET5_0 || NET6_0_OR_GREATER
+#if NETCOREAPP3_0_OR_GREATER
```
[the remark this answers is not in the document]
BrennanConroy: @sebastienros ?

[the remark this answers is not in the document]
sebastienros: I changed other benchmarks recently to target 6 at mininum. I think we can do the same here as we haven't had to run it even for 5 since last year.

The excuse to not do it earlier is that we always talk about getting rid of this project and refactor it in more granular apps (middleware, mvc) and updating it to the latest templates. That some good amount of work because we will need to update the benchmarks configurations and CI too. But this could be done in multiple steps.

src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -40,11 +37,8 @@ public Task Invoke(HttpContext httpContext)
                     syncIOFeature.AllowSynchronousIO = true;
                 }
 
-                _writer ??= new Utf8JsonWriter(httpContext.Response.Body);
-                _writer.Reset(httpContext.Response.Body);
-
-                JsonSerializer.Serialize<JsonMessage>(_writer, new JsonMessage { message = "Hello, World!" }, CustomJsonContext.Default.JsonMessage);
-                return Task.CompletedTask;
+                return JsonSerializer.SerializeAsync<JsonMessage>(httpContext.Response.Body, new JsonMessage { message = "Hello, World!" },
```
sebastienros, line 40: Remove `.AllowSynchronousIO = true` ?

BrennanConroy: Maybe, forgot pushing to the branch would update a PR 😄 
Still need to test and see if the recent STJ change did anything

BrennanConroy: @eiriktsarpalis, I thought https://github.com/dotnet/runtime/pull/78646 was supposed to allow source-gen code to be used with `SerializeAsync`? But when trying I get the exception shown in https://github.com/dotnet/runtime/issues/71933, without creating a combined context. Are we using it wrong here, or is this another version of the same bug and source-gen still isn't usable with `SerializeAsync`?

BrennanConroy: Removing the `GenerationMode = JsonSourceGenerationMode.Serialization` argument made it work. Is that required for `SerializeAsync`?

BrennanConroy:
176 MB/sec -> 33 MB/sec with Source gen + new fast path streaming serialization.
Also got 1.006m RPS, the benchmark averages ~988k so slight improvement.

src/Benchmarks/Middleware/JsonMiddleware.cs
eiriktsarpalis: > Removing the GenerationMode = JsonSourceGenerationMode.Serialization argument made it work. Is that required for SerializeAsync?

Streaming serialization requires metadata generation. The change in dotnet/runtime#78646 simply introduces a heuristic that eventually falls back to fast-path serialization provided that the JSON payloads are small enough.

src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -66,6 +50,12 @@ public static class JsonMiddlewareExtensions
         public static IApplicationBuilder UseJson(this IApplicationBuilder builder) => builder.UseMiddleware<JsonMiddleware>();
     }
 
+    [JsonSourceGenerationOptions]
```
eiriktsarpalis, line 53: Can be removed at this point:

```suggestion
```

src/Benchmarks/Middleware/JsonMiddleware.cs
eiriktsarpalis: All serialization overloads pool `Utf8JsonWriter`, with the exception of the async streaming serialization overloads that can't use the fast path.

src/Benchmarks/Middleware/JsonMiddleware.cs
DamianEdwards: @BrennanConroy any reason we can't simply use the `WriteAsJsonAsync` extension methods here, passing in the options instance from DI resolved from the middleware constructor (to avoid request services)? I assume you'd have to add the serialization context to the options during DI setup too. Would make this slightly more "real world" as it'd use JSON web defaults then.

BrennanConroy: Took forever, but I got it working. One unfortunate thing you have to do to make it work using `WriteAsJsonAsync` is to set `JsonSerialierOptions.Encoder` to `null`.

DamianEdwards: You going to push the update so we can see? :smile:

BrennanConroy: Enjoy

src/Benchmarks/Program.cs
```diff
@@ -84,6 +85,11 @@ public static void Main(string[] args)
                         }
 #endif
                     })
+                    .ConfigureHttpJsonOptions(jsonOptions =>
+                    {
+                        jsonOptions.SerializerOptions.Encoder = null;
```
eerhardt, line 90: Why do we need this? It seems like a bug.... At least a comment explaining why.

BrennanConroy: https://source.dot.net/#System.Text.Json/System/Text/Json/Serialization/JsonSerializerContext.cs,60
The fast path logic requires very specific settings. Didn't add a comment because not sure we want to use this approach in the benchmark.

src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -50,7 +49,7 @@ public static class JsonMiddlewareExtensions
         public static IApplicationBuilder UseJson(this IApplicationBuilder builder) => builder.UseMiddleware<JsonMiddleware>();
     }
 
-    [JsonSourceGenerationOptions]
+    [JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase)]
```
DamianEdwards, line 52: Why is this needed if we're using the instance of the serializer options from DI that's already configured to use this casing by default?

src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -20,25 +22,22 @@ public class JsonMiddleware
...
 
-                var syncIOFeature = httpContext.Features.Get<IHttpBodyControlFeature>();
-                if (syncIOFeature != null)
-                {
-                    syncIOFeature.AllowSynchronousIO = true;
-                }
-
-                return JsonSerializer.SerializeAsync<JsonMessage>(httpContext.Response.Body, new JsonMessage { message = "Hello, World!" },
-                    CustomJsonContext.Default.JsonMessage);
+                return httpContext.Response.WriteAsJsonAsync<JsonMessage>(new JsonMessage { message = "Hello, World!" }, _jsonOptions, httpContext.RequestAborted);
```
DamianEdwards, line 40: You don't need to specify the type arg here do you?

src/Benchmarks/Middleware/JsonMiddleware.cs
BrennanConroy: Because we add this context to the options, and when it serializes JsonMessage this context is used which has the default of `null` naming policy. And the naming policy needs to match the main options in https://source.dot.net/#System.Text.Json/System/Text/Json/Serialization/JsonSerializerContext.cs,74

I of course could be doing something wrong, these Json APIs are way too complex.

src/Benchmarks/Program.cs
```diff
@@ -84,6 +85,11 @@ public static void Main(string[] args)
...
+                        jsonOptions.SerializerOptions.AddContext<Middleware.CustomJsonContext>();
```
eerhardt, line 91: 1. AddContext is going to be obsoleted as per https://github.com/dotnet/runtime/issues/83280.
2. I'm a bit confused at how this works today. `AddContext` [will put the `CustomJsonContext` at the end of the TypeInfoResolver](https://github.com/dotnet/runtime/blob/967250cc86b4c613015066b47b882e7424fdb590/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializerOptions.cs#L175-L179), after the "default" resolver. Doesn't this mean the reflection based type info resolver still gets used over the `CustomJsonContext`?

cc @eiriktsarpalis

BrennanConroy: This worked temporarily until AddContext was fixed to properly combine contexts. I think my local branch fixed that but never pushed it.

src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -4,34 +4,38 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.AspNetCore.Http.Json;
 
 namespace Benchmarks.Middleware
 {
     public class JsonMiddleware
     {
         private static readonly PathString _path = new PathString(Scenarios.GetPath(s => s.Json));
-        private static readonly UTF8Encoding _encoding = new UTF8Encoding(false);
         private const int _bufferSize = 27;
```
davidfowl, line 21: How much performance do we lose if we remove this?

src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -4,34 +4,38 @@
...
         public Task Invoke(HttpContext httpContext)
         {
             if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal))
             {
                 httpContext.Response.StatusCode = 200;
-                httpContext.Response.ContentType = "application/json";
                 httpContext.Response.ContentLength = _bufferSize;
 
-                return JsonSerializer.SerializeAsync<JsonMessage>(httpContext.Response.Body, new JsonMessage { message = "Hello, World!" });
+                return httpContext.Response.WriteAsJsonAsync(new JsonMessage { message = "Hello, World!" }, _jsonOptions, httpContext.RequestAborted);
```
davidfowl, line 38: Why are we passing the token?

src/Benchmarks/Middleware/JsonMiddleware.cs
eiriktsarpalis: It is not strictly necessary, since serialization will ultimately conform to whatever configuration is passed by the relevant `JsonSerializerOptions`. The benefit of declaring here though is that any naming policies will be computed at compile time and incorporated in the fast-path method.

src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -43,6 +47,12 @@ public static class JsonMiddlewareExtensions
         public static IApplicationBuilder UseJson(this IApplicationBuilder builder) => builder.UseMiddleware<JsonMiddleware>();
     }
 
+    [JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase)]
```
eiriktsarpalis, line 50: It should now be possible to do this:
```suggestion
    [JsonSourceGenerationOptions(JsonSerializerDefaults.Web)]
```

src/Benchmarks/Program.cs
eiriktsarpalis: You should use this to ensure equivalent behavior with the .NET 8 build
```suggestion
                        jsonOptions.SerializerOptions.TypeInfoResolver = JsonTypeInfoResolver.Combine(jsonOptions.SerializerOptions.TypeInfoResolver, Middleware.CustomJsonContext.Default);
```

src/Benchmarks/Middleware/JsonMiddleware.cs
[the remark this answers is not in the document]
DamianEdwards: Ideally you'd set it up in DI and not have to pass it at all here. Is that still not possible?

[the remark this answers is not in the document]
BrennanConroy: It should work with DI (you can see we get the options via DI in the ctor), but that's just extra per request work that we can amortize via the ctor.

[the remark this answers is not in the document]
eerhardt: Passing in the options, and making the serializer look up the JsonTypeInfo every request, is just extra work as well. If you just pass the JsonTypeInfo directly in, it is less work to do per request.

src/Benchmarks/Program.cs
BrennanConroy: Shouldn't it be `Combine(Middleware.CustomJsonContext.Default, jsonOptions.SerializerOptions.TypeInfoResolver);`? Or does order not matter with `Combine`?

src/Benchmarks/Middleware/JsonMiddleware.cs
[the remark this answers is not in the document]
BrennanConroy: Yep, I'll try that one out

src/Benchmarks/Program.cs
eiriktsarpalis: You're right, I messed up the order in my suggestion.

src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -4,34 +4,53 @@
...
+#else
+        private readonly JsonSerializerOptions _jsonOptions;
+#endif
 
-        public JsonMiddleware(RequestDelegate next) => _next = next;
+        public JsonMiddleware(RequestDelegate next, IOptions<JsonOptions> jsonOptions)
+        {
+            _next = next;
+#if NET8_0_OR_GREATER
+            _jsonTypeInfo = JsonTypeInfo.CreateJsonTypeInfo<JsonMessage>(jsonOptions.Value.SerializerOptions);
```
BrennanConroy, line 34: @eiriktsarpalis is `JsonTypeInfo.CreateJsonTypeInfo<JsonMessage>(jsonOptions.Value.SerializerOptions);` supposed to create a `JsonTypeInfo<T>` that doesn't seem to work?

src/Benchmarks/Middleware/JsonMiddleware.cs
BrennanConroy: We lose 20-30k RPS

src/Benchmarks/Middleware/JsonMiddleware.cs
eiriktsarpalis: That just creates a blank, mutable `JsonTypeInfo` for the given type. You should use `(JsonTypeInfo<JsonMessage>)jsonOptions.Value.SerializerOptions.GetTypeInfo(typeof(JsonMessage));` instead.

src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -4,34 +4,51 @@
...
+#else
+        private readonly JsonSerializerOptions _jsonOptions;
+#endif
 
-        public JsonMiddleware(RequestDelegate next) => _next = next;
+        public JsonMiddleware(RequestDelegate next, IOptions<JsonOptions> jsonOptions)
+        {
+            _next = next;
+#if NET8_0_OR_GREATER
+            _jsonTypeInfo = jsonOptions.Value.SerializerOptions.GetTypeInfo(typeof(JsonMessage)) as JsonTypeInfo<JsonMessage>;
```
eiriktsarpalis, line 33: The generic is guaranteed to match the type you're passing it:
```suggestion
            _jsonTypeInfo = (JsonTypeInfo<JsonMessage>)jsonOptions.Value.SerializerOptions.GetTypeInfo(typeof(JsonMessage));
```

src/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -4,34 +4,53 @@
...
                 httpContext.Response.ContentLength = _bufferSize;
 
-                return JsonSerializer.SerializeAsync<JsonMessage>(httpContext.Response.Body, new JsonMessage { message = "Hello, World!" });
+                return httpContext.Response.WriteAsJsonAsync(new JsonMessage { message = "Hello, World!" },
+#if NET8_0_OR_GREATER
+                    _jsonTypeInfo
+#else
+                    _jsonOptions
+#endif
+                    );
```
eerhardt, line 53: I think this is too complicated. It would make a ton of these changes much simpler if this was just:

```suggestion
                return httpContext.Response.WriteAsJsonAsync(new JsonMessage { message = "Hello, World!" }, CustomJsonContext.Default.JsonMessage);
```

Then get rid of the constructor changes, and the `ConfigureHttpJsonOptions` call.

src/Benchmarks/Program.cs
eerhardt: Is this still needed now that we aren't using the JsonOptions anymore?

src/Benchmarks/Benchmarks.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
eerhardt, line 1: (nit) can revert this file.

eerhardt (approved):
This looks great to me.

Do we have before and after numbers?

src/Benchmarks/Program.cs
BrennanConroy: Oh wow, we can remove it!

eerhardt: To confirm we are using "fast path" on both .NET 8 AND 7 with the latest changes, right? Just making sure we are comparing apples-to-apples if we compare them against each other.

Issue #1776, opened 2022-12-13
Stage1Aot fails


sebastienros:
```text
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HMMT69A75KHE", Request id "0HMMT69A75KHE:000000CA": An unhandled exception was thrown by the application.
      System.NotSupportedException: 'System.Text.Json.Serialization.Converters.SmallObjectWithParameterizedConstructorConverter`5[WeatherForecast, System.DateOnly, System.Int32, System.String, System.Object]' is missing native code or metadata. This can happen for code that is not compatible with trimming or AOT. Inspect and fix trimming and AOT related warnings that were generated when the app was published. For more information see https://aka.ms/nativeaot-compatibility The unsupported member type is located on type 'WeatherForecast[]'. Path: $.
       ---> System.NotSupportedException: 'System.Text.Json.Serialization.Converters.SmallObjectWithParameterizedConstructorConverter`5[WeatherForecast, System.DateOnly, System.Int32, System.String, System.Object]' is missing native code or metadata. This can happen for code that is not compatible with trimming or AOT. Inspect and fix trimming and AOT related warnings that were generated when the app was published. For more information see https://aka.ms/nativeaot-compatibility
         at System.Reflection.Runtime.General.TypeUnifier.WithVerifiedTypeHandle(RuntimeConstructedGenericTypeInfo, RuntimeTypeInfo[]) + 0x95
         at System.Text.Json.Serialization.Converters.ObjectConverterFactory.CreateConverter(Type, JsonSerializerOptions) + 0x229
         at System.Text.Json.Serialization.JsonConverterFactory.GetConverterInternal(Type, JsonSerializerOptions) + 0x18
         at System.Text.Json.JsonSerializerOptions.ExpandConverterFactory(JsonConverter, Type) + 0x30
         at System.Text.Json.Serialization.Metadata.DefaultJsonTypeInfoResolver.GetConverterForType(Type, JsonSerializerOptions, Boolean) + 0x7c
         at System.Text.Json.Serialization.Metadata.DefaultJsonTypeInfoResolver.CreateJsonTypeInfo(Type, JsonSerializerOptions) + 0x26
         at System.Text.Json.Serialization.Metadata.DefaultJsonTypeInfoResolver.GetTypeInfo(Type, JsonSerializerOptions) + 0x47
         at System.Text.Json.JsonSerializerOptions.GetTypeInfoNoCaching(Type type) + 0x39
[...]

PR #1777, opened 2022-12-14
Add serializer options to fix Goldilocks benchmark.

mitchdenny:
Addresses #1776

Added a serializer context. This should enable the app to respond with a valid JSON payload.

### DIFF
--- a/src/Goldilocks/Goldilocks.csproj
+++ b/src/Goldilocks/Goldilocks.csproj
@@ -7,8 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.OpenApi" Version="$(MicrosoftAspNetCoreAppPackageVersion)" />
-    <PackageReference Include="Swashbuckle.AspNetCore" Version="6.4.0" />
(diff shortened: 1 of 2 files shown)


src/Goldilocks/Goldilocks.csproj
```diff
@@ -7,8 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.OpenApi" Version="$(MicrosoftAspNetCoreAppPackageVersion)" />
-    <PackageReference Include="Swashbuckle.AspNetCore" Version="6.4.0" />
+    <PackageReference Include="System.Text.Json" Version="$(MicrosoftAspNetCoreAppPackageVersion)" />
```
sebastienros, line 10: It's part of net8.0, shouldn't be necessary

sebastienros:
Can you confirm there are no more bad requests with these changes?

src/Goldilocks/Goldilocks.csproj
mitchdenny: Fixed the unnecessary dependency.

mitchdenny:
> Can you confirm there are no more bad requests with these changes?

Yes, I just verified again on `8.0.100-alpha.1.22608.10`.

PR #1780, opened 2022-12-28
Creating separate MVC version of TechEmpower scenarios

sebastienros:
Starting to split the big Benchmarks app into separate ones. Goal is to make it manageable and easier to sync with TE's repos, but also be closer to what the templates create. We are still doing the changes necessary for performance when it matters but it has to be realistic changes.

Database scenarios are not yet implemented here as the current app contains a lot of them and I think we can refocus. For instance, RAW should be kept for Platform, the same way EF is not in Platform. Should we only go with EF here? That would mean MVC/Razor/EF.

If we want to compare EF/Dapper/Raw and also support other databases (in which case the aspnet app type doesn't matter) we could have a separate app (minimal) that would provide all these variations. But still use the TechEmpower scenarios. This one would not go in the TE repository so it could get messier.

### DIFF
--- a/src/BenchmarksApps/TechEmpower/Mvc/Controllers/DatabaseController.cs
+++ b/src/BenchmarksApps/TechEmpower/Mvc/Controllers/DatabaseController.cs
@@ -7,8 +7,6 @@ namespace Mvc.Controllers;
 public class DatabaseController : Controller
 {
     [HttpGet("fortunes")]
-    public Task<List<Fortune>> Fortunes([FromServices] Db db)
-    {
-        return db.LoadFortunesRows();
-    }
+    public Task<List<Fortune>> Fortunes([FromServices] Db db) 
+            => db.LoadFortunesRows();
 }
--- a/src/BenchmarksApps/TechEmpower/Mvc/Controllers/DatabaseController.cs
+++ b/src/BenchmarksApps/TechEmpower/Mvc/Controllers/DatabaseController.cs
@@ -1,0 +1,14 @@
+﻿using Microsoft.AspNetCore.Mvc;
+using Mvc.Database;
+using Mvc.Models;
+
+namespace Mvc.Controllers;
+
+public class DatabaseController : Controller
+{
+    [HttpGet("fortunes")]
+    public Task<List<Fortune>> Fortunes([FromServices] Db db)
+    {
+        return db.LoadFortunesRows();
+    }
+}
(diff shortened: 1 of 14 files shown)


src/BenchmarksApps/TechEmpower/Mvc/Controllers/DatabaseController.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Mvc.Controllers;
+
+public class DatabaseController : Controller
+{
+    [HttpGet("fortunes")]
+    public Task<List<Fortune>> Fortunes([FromServices] Db db)
+    {
+        return db.LoadFortunesRows();
+    }
```
davidfowl, line 13:
```suggestion
    public Task<List<Fortune>> Fortunes([FromServices] Db db) 
            => db.LoadFortunesRows();
```

src/BenchmarksApps/TechEmpower/Mvc/Database/ApplicationDbContext.cs
```diff
@@ -0,0 +1,16 @@
...
+public sealed class ApplicationDbContext : DbContext
+{
+    public ApplicationDbContext(DbContextOptions options)
+        : base(options)
+    {
+        ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;
+        ChangeTracker.AutoDetectChangesEnabled = false;
+    }
+
+    public DbSet<Fortune>? Fortune { get; set; }
```
davidfowl, line 15: Plural?

```suggestion
    public required DbSet<Fortune> Fortunes { get; set; }
```

src/BenchmarksApps/TechEmpower/Mvc/Database/Db.cs
```diff
@@ -0,0 +1,57 @@
...
+
+#pragma warning disable EF1001 // Using internal EF pooling APIs, can be cleaned up after 6.0.0-preview4
+
+namespace Mvc.Database;
+
+public class Db
+{
+    private readonly PooledDbContextFactory<ApplicationDbContext> _dbContextFactory;
+
+    public Db(AppSettings? appSettings)
```
davidfowl, line 15: Why is this nullable?

sebastienros: it's coming from `Configuration.Get<AppSettings>()` so it's nullable. Either this accepts a nullable one, or we handle that differently if that's undesirable to have nullable in the ctor.

davidfowl: Do this instead:

```C#
var appSettings = new AppSettings();
Configuration.Bind(appSettings);
```

sebastienros:
Replaced by https://github.com/aspnet/Benchmarks/pull/1790

PR #1781, opened 2023-01-13
Add HttpClient benchmarks to CI runs

CarnaViire:
cc @sebastienros


sebastienros:
Did you manage to use the template generation to create the changes on the pipeline?
Do you have any question about the files you had to change?

CarnaViire:
> Did you manage to use the template generation to create the changes on the pipeline?

Yes

> Do you have any question about the files you had to change?

I don't really have any questions for modifying the files, I figured what I need from grpc example 🤔

The only question is about testing, more like whether I'm missing something, or it's supposed to be like that:
It seems like the only way to *actually* test the file is to merge it and see what happens in CI 😅 I did manually test the crank commands that should be executed in the `for` loop in `steps:` section ([link](https://github.com/aspnet/Benchmarks/blob/5863e912420a050a388c0d9879a7cb1ab2f87fdc/build/httpclient-scenarios.yml#L63-L79)), but I wasn't able to actually generate the result of the `for` loop.

sebastienros:
It's creating a new SQL table, so we'll need a new PowerBI page too, is that what you want?

CarnaViire:
> It's creating a new SQL table, so we'll need a new PowerBI page too, is that what you want?

Yes, that's the goal

PR #1782, opened 2023-01-16
Update Goldilocks code/benchmark based on new AOT enabled API template.

mitchdenny:
We've got a new AOT API template. This PR changes our benchmark to use it.


src/Goldilocks/Program.cs
```diff
@@ -1,43 +1,29 @@
 using System.Text.Json.Serialization;
+using Goldilocks;
 
 var builder = WebApplication.CreateBuilder(args);
+builder.Logging.AddConsole();
```
DamianEdwards, line 5: Let's remove this line for the benchmarks.

src/Goldilocks/Todo.cs
```diff
@@ -0,0 +1,64 @@
...
+static class TodoGenerator
+{
+    private static readonly (string[] Prefixes, string[] Suffixes)[] _parts = new[]
+        {
+            (new[] { "Walk the", "Feed the" }, new[] { "dog", "cat", "goat" }),
+            (new[] { "Do the", "Put away the" }, new[] { "groceries", "dishes", "laundry" }),
+            (new[] { "Clean the" }, new[] { "bathroom", "pool", "blinds", "car" })
+        };
+
+    internal static IEnumerable<Todo> GenerateTodos(int count = 5)
```
DamianEdwards, line 23: Given this is a benchmark, perhaps we should remove the randomness here. Although it's only at application startup, the actual number of bytes in the todo titles will change from run to run which isn't ideal. Let's just make it return a hardcoded set of five todos.

src/Goldilocks/Goldilocks.csproj
```diff
@@ -4,6 +4,11 @@
     <TargetFramework>net8.0</TargetFramework>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
+    <NoDefaultLaunchSettingsFile Condition="'$(ExcludeLaunchSettings)' == 'True'">true</NoDefaultLaunchSettingsFile>
```
DamianEdwards, line 7: Delete this line, that's a template condition, not output that's intended for the output project.

src/Goldilocks/Goldilocks.csproj
```diff
@@ -4,6 +4,11 @@
...
+    <RootNamespace>Goldilocks</RootNamespace>
+    <ServerGarbageCollection>False</ServerGarbageCollection>
+    <TrimMode>full</TrimMode>
```
DamianEdwards, line 10: Do we pass in the `PublishAot` from the outside? If so, perhaps just pass `TrimMode` too when we pass `PublishAot` given we've removed it from the project file.

src/Goldilocks/Program.cs
eerhardt: I wonder if that is "fair" to report, both for app size and RPS/perf.

DamianEdwards: None of our other benchmarks have logging enabled. We could certainly capture both sets of figures if we wish to track the impact of logging, but the default logging levels include multiple messages per request which will significantly impact the RPS results.

sebastienros: In order **not** to have logging enabled, a line actually needs to be added:

```c#
builder.Logging.ClearProviders();
```

DamianEdwards: Yep, until the new "slim" builder is available.

src/Goldilocks/Todo.cs
```diff
@@ -0,0 +1,27 @@
...
+    public string? Title { get; set; }
+
+    public DateOnly? DueBy { get; set; }
+
+    public bool IsComplete { get; set; }
+}
+
+static class TodoGenerator
+{
+    internal static IEnumerable<Todo> GenerateTodos(int count = 5)
```
sebastienros, line 16: readonly static array? `AllTodos`?

src/Goldilocks/Program.cs
```diff
@@ -1,43 +1,29 @@
...
-            DateOnly.FromDateTime(DateTime.Now.AddDays(index)),
-            Random.Shared.Next(-20, 55),
-            summaries[Random.Shared.Next(summaries.Length)]
-        ))
-        .ToArray();
-    return forecast;
-});
+var todosApi = app.MapGroup("/todos");
+todosApi.MapGet("/", () => sampleTodos);
+todosApi.MapGet("/{id}", (int id) =>
```
sebastienros, line 18: There is no benchmark for that right now, do you still keep it around?

mitchdenny: Happy to remove. @DamianEdwards  -  I know you wanted this to be as similar as possible to the template. In this case it won't impact RPS, in theory it could impact published size but probably not in this particular case. So it would be safe to delete.

src/Goldilocks/Program.cs
```diff
@@ -1,43 +1,29 @@
 using System.Text.Json.Serialization;
+using Goldilocks;
 
 var builder = WebApplication.CreateBuilder(args);
+builder.Logging.ClearProviders();
```
sebastienros, line 5: I realize that by clearing all the providers (including the console) then we can't detect when the application is ready: `Application started.`

the mitigation is to add this before `app.Run()`:

```
app.Lifetime.ApplicationStarted.Register(() => Console.WriteLine("Application started. Press Ctrl+C to shut down."));
```

src/Goldilocks/Program.cs
sebastienros: Then keep it, but maybe with a comment (it won't impact size ;)) just because in a week I will ask the same question :/

Same of the clearing of the logging provider, that it's deriving from the template but only in a way that it matches what benchmarks do.

src/Goldilocks/Program.cs
mitchdenny: Oh good point I forgot how benchmarks figures out when to start injecting load ;)

sebastienros (approved):
GTG in terms of benchmarking infra

src/Goldilocks/Program.cs
DamianEdwards: Keep it, with a comment for clarity. Technically it can impact startup time and per request performance (more routes in the tree).

sebastienros:
Please use the squash merges button to keep the history clean. Thank ;)

mitchdenny:
Oh sorry about that ... can we change that to be the default via policy? That is what a lot of repos do to avoid this issue ;)

sebastienros:
@dougbu @wtgodbe can you configure this repos to only allow squash merges?

dougbu:
> @dougbu @wtgodbe can you configure this repos to only allow squash merges?

done

PR #1784, opened 2023-01-21
Use the new CreateSlimBuilder API

eerhardt:
The `dotnet new api -aot` template is getting updated to use the new slim API in https://github.com/dotnet/aspnetcore/pull/46040. Updating the benchmark to match.

NOTE: this depends on getting a build with https://github.com/dotnet/aspnetcore/pull/46040 merged. So it will fail until then.

### DIFF
--- a/scenarios/goldilocks.benchmarks.yml
+++ b/scenarios/goldilocks.benchmarks.yml
@@ -29,7 +29,6 @@ scenarios:
       buildArguments: 
         - "/p:PublishAot=true"
         - "/p:StripSymbols=true"
-        - "/p:TrimMode=full"
     load:
       job: wrk
       variables:
--- a/src/Goldilocks/Goldilocks.csproj
+++ b/src/Goldilocks/Goldilocks.csproj
@@ -6,11 +6,6 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <RootNamespace>Goldilocks</RootNamespace>
     <ServerGarbageCollection>False</ServerGarbageCollection>
-    <PublishIISAssets>false</PublishIISAssets>
   </PropertyGroup>
-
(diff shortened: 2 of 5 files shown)


eerhardt:
FYI - https://github.com/dotnet/aspnetcore/pull/46040 has been merged. I'm not sure how long it takes for a new aspnetcore build to propagate to the benchmarks.

cc @mitchdenny @sebastienros

mitchdenny:
Do we still need to clear providers with slim builder?

sebastienros:
Once a PR is merged it should take a just a few hours to get a build available in the public feeds, and picked up by crank. Crank checks the nuget feeds directly.

eerhardt:
> Do we still need to clear providers with slim builder?

I tested it a few times locally with and without clear providers. I was seeing 0.5%-1% drop in RPS when I don't clear.

With ClearProviders():

| Requests/sec           | 480,957   |                               |
| Requests               | 7,262,492 |                               |

Without:

| Requests/sec           | 478,373   |                               |
| Requests               | 7,223,112 |                               |

So for now I left it in.

sebastienros:
@eerhardt with and without ClearProviders, are there any changes in the console output, for instance wrt "Application started" message and other "lifetime" messages?

eerhardt:
No, no change in the console output.

I debugged the app, and now with CreateSlimBuilder there are no `ILoggerProvider` services registered anymore, which is why ClearProviders() removes:

https://github.com/dotnet/runtime/blob/007df054a526ed9e3dc70b43bfa330943bd4816a/src/libraries/Microsoft.Extensions.Logging/src/LoggingBuilderExtensions.cs#L48

So I guess the above numbers were just within margin of error?

PR #1786, opened 2023-01-27
Add Stage1Aot to size regression

eerhardt:
This will auto file issues when the Stage1Aot size regresses.

@sebastienros - is this all that needs to be done? How can I test these changes?

### DIFF
--- a/build/regressions.config.yml
+++ b/build/regressions.config.yml
@@ -72,6 +72,8 @@ sources:
     rules:
       - include: FortunesEF
       - include: Stage1Aot
+        owners:
+          - eerhardt
     daysToLoad: 60
     stdevCount: 5
     regressions:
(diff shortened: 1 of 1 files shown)


build/regressions.config.yml
```diff
@@ -85,6 +85,7 @@ sources:
         - area-perf
       owners:
         - sebastienros
+        - eerhardt
```
sebastienros, line 88: You can add yourself only for Stage1Aot (same for labels)

```yaml
    rules:
      - include: FortunesEF
      - include: Stage1Aot
        owners:
          - eerhardt
```

sebastienros: Does this sample count as documentation?

https://github.com/dotnet/crank/blob/main/src/Microsoft.Crank.RegressionBot/sample.config.yml#L7-L14

eerhardt: Done.

eerhardt:
Ok, does this look better?

Any thought to removing `daysToSkip`? For size, the changes don't swing too much. I don't think we need 2 days before logging the issue.

PR #1788, opened 2023-01-31
Add basic gRPC app

JamesNK:
Made a simple gRPC todo app and added crank/benchmark config. I haven't tested this yet.


eerhardt:
I have some hesitation using the "Stage 1" naming here. I know we are already using it, but doubling down on the name doesn't feel like the right thing to me.

1. "Stage 1" is a bad name long term all together. When we have a new effort in the future, we can no longer call the first stage of that new effort "Stage 1". As it will now be ambiguous.
2. In planning, "Stage 1" explicitly didn't include gRPC. gRPC was considered "Stage 2.b".

What does everything think about renaming "Stage 1" to something more descriptive and future proof? Options I can think of:

- ApiTemplate
- BasicMinimalApi
- BasicApi
- I'm open to more, if folks can think of better names

I'd also like to get rid of the "Goldilocks" name in the repo as well, since it is confusing to new people.

scenarios/goldilocks.benchmarks.yml
```diff
@@ -37,9 +48,38 @@ scenarios:
...
+      job: stage1grpcaspnetbenchmarks
+      buildArguments: 
+        - "/p:PublishAot=true"
+        - "/p:StripSymbols=true"
+    load:
+      job: h2LoadClient
+      variables:
+        presetHeaders: grpc
+        path: /goldilocks.TodoService/GetAllTodos
+        body: AAAAAAA=
```
eerhardt, line 69: What does this do?

src/BenchmarksApps/Goldilocks/Stage1Grpc/Goldilocks.csproj
```diff
@@ -0,0 +1,17 @@
...
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Protobuf Include="Protos\todo.proto" GrpcServices="Server" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Grpc.AspNetCore" Version="2.49.0" />
```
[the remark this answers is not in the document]
sebastienros: Note that this has to be published in the internal nuget feed, not all Nuget.org versions are available automatically. But I understand we can just ask for it to be pushed. I assume we just skip versions when there are not specific reasons to update (minor fixes for instance).

scenarios/goldilocks.benchmarks.yml
```diff
@@ -37,9 +48,38 @@ scenarios:
...
+        connections: 28
+        threads: 28
+        streams: 70
```
sebastienros, line 72: Someday we should do another pass at running a set of benchmarks and vary clients/streams to find the sweet spot.

sebastienros (approved):

scenarios/goldilocks.benchmarks.yml
JamesNK: Base64 encoded POST body of the HTTP request. It's an empty gRPC message, which is 5 zeros.

src/BenchmarksApps/Goldilocks/Stage1Grpc/Goldilocks.csproj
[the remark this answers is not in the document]
JamesNK: It will have to be updated manually.

The server side of gRPC isn't rapidly changing. Getting out of date is ok. When we care, we can update it.

JamesNK:
> I'd also like to get rid of the "Goldilocks" name in the repo as well, since it is confusing to new people.

What do you suggest? We're measuring multiple things with these benchmarks: publish size, memory usage, RPS, AOT vs CoreCLR, etc

eerhardt:
> What do you suggest?

I don't have all the history of what names are loaded/not allowed, and what aren't in ASP.NET. But if I had to pick one of the above by myself, I would call the existing app

`BasicMinimalApi`

And this one would be

`BasicGrpc`

Both would be under the `src/BenchmarksApps` folder directly.

JamesNK:
I've renamed and moved the apps. I haven't renamed any of the jobs or scenarios in the config files.

sebastienros:
[scenarios/readme.md](https://github.com/aspnet/Benchmarks/tree/main/scenarios#goldilocks-benchmarks) might need to be updated too

PR #1791, opened 2023-02-08
Add server GC vs workstation GC to AOT benchmarks

JamesNK:
I tested `DOTNET_gcServer=1` on my machine and it appears to override the publish setting.

### DIFF
--- a/build/nativeaot-scenarios.yml
+++ b/build/nativeaot-scenarios.yml
@@ -55,7 +55,7 @@ parameters:
 
   - displayName: Goldilocks gRPC Stage 1 (NativeAOT - Server GC)
     # workaround https://github.com/dotnet/runtime/issues/81382 by explicitly referencing a Microsoft.Dotnet.ILCompiler version
-    arguments: --scenario basicgrpcpublishaot $(goldilocksJobs) --property scenario=Stage1GrpcAot --property publish=nativeaot --application.packageReferences \"Microsoft.Dotnet.ILCompiler=$(MicrosoftNETCoreAppPackageVersion)\" --application.environmentVariables DOTNET_gcServer=1
+    arguments: --scenario basicgrpcpublishaot $(goldilocksJobs) --property scenario=Stage1GrpcAotServerGC --property publish=nativeaot --application.packageReferences \"Microsoft.Dotnet.ILCompiler=$(MicrosoftNETCoreAppPackageVersion)\" --application.environmentVariables DOTNET_gcServer=1
     condition: true
 
   - displayName: Gin Json
--- a/build/nativeaot-scenarios.yml
+++ b/build/nativeaot-scenarios.yml
@@ -18,6 +18,10 @@ parameters:
     arguments: --scenario basicminimalapivanilla $(goldilocksJobs) --property scenario=Stage1 --property publish=coreclr
     condition: true
 
+  - displayName: Goldilocks Stage 1 (CoreCLR - Server GC)
+    arguments: --scenario basicminimalapivanilla $(goldilocksJobs) --property scenario=Stage1ServerGC --property publish=coreclr --application.environmentVariables DOTNET_gcServer=1
+    condition: true
+
   - displayName: Goldilocks Stage 1 (CoreCLR - No PGO)
     arguments: --scenario basicminimalapivanilla $(goldilocksJobs) --property scenario=Stage1NoPgo --property publish=coreclr --application.environmentVariables DOTNET_TieredPGO=0
     condition: Math.round(Date.now() / 43200000) % 6 == 1 # once every 6 half-days (43200000 ms per half-day)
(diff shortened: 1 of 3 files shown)


build/nativeaot-scenarios.yml
```diff
@@ -18,6 +18,10 @@ parameters:
     arguments: --scenario basicminimalapivanilla $(goldilocksJobs) --property scenario=Stage1 --property publish=coreclr
     condition: true
 
+  - displayName: Goldilocks Stage 1 (CoreCLR - Server GC)
+    arguments: --scenario basicminimalapivanilla $(goldilocksJobs) --property scenario=Stage1ServerGC --property publish=coreclr --application.environmentVariables DOTNET_gcServer=1
```
eerhardt, line 22: > application.environmentVariables DOTNET_gcServer=1

Setting certain environment variables don't work in NativeAOT. This is because we hard-code which GC is used at publish time, and it can't be switched. A big reason for this is that we want to reduce size of the app, and having more GC code is undesirable.

See how I did it here: 
https://github.com/eerhardt/Albums/blob/349586e97a709facb78cce6970a4ddc6981170f8/AlbumsNetCore/Dockerfile#L14

So in order to make this work, we need to pass the variable at build/publish time. 

cc @Maoni0 @MichalStrehovsky

MichalStrehovsky: If server gc is not enabled at build time, we build with a smaller version of the runtime lib that doesn't have the server gc code. Supporting server gc is literally having another copy of the whole gc in the app.

If server gc is enabled at build time, it will be turned on by default, and you can use the env var to disable it.

eerhardt: I think this is a common misunderstanding. Do we have this (and other settings like this) documented somewhere?

Maoni0: clrgc.dll, which includes both Workstation and Server flavor, is 640kb. so half of that is 320kb. what's the percentage of 320kb in an AoT hello world app and min AoT web app?

JamesNK: That makes things complicated. Build arguments are specified in the scenario: https://github.com/aspnet/Benchmarks/blob/4c9ee8a45e54c01359cd48944cb61326d18ea7ab/scenarios/goldilocks.benchmarks.yml#L37-L47

The env var approach allowed benchmarks to reuse the scenario definitions. @sebastienros is there a way to pass a build argument into a scenario? Example?

sebastienros: Everything that is in the json can also be defined in command line. Collections are additive, hence this would work: `--application.buildArguments /p:ServerGarbageCollection=true`

sebastienros: Example usage in a yml file since it needs to be "json" encoded as it is passed as a json string:

https://github.com/aspnet/Benchmarks/blob/4c9ee8a45e54c01359cd48944cb61326d18ea7ab/build/singlefile-scenarios.yml#L30

JamesNK: Example I think: https://github.com/aspnet/Benchmarks/blob/4c9ee8a45e54c01359cd48944cb61326d18ea7ab/build/singlefile-scenarios.yml#L30

I'll play around with yml files and see if we can get the GC mode switch coming from a build argument.

JamesNK: We linked to the exact same line as an example 😆

MichalStrehovsky: > clrgc.dll, which includes both Workstation and Server flavor, is 640kb. so half of that is 320kb. what's the percentage of 320kb in an AoT hello world app and min AoT web app?

@Maoni0 Hello world in default configuration is 1,857,536 bytes. With ServerGarbageCollection, it goes to 2,093,568 bytes, so 12% worse.

Fun fact: out of those 1,857,536 bytes, 100 kB (almost 6%) are sorting tables for GC's vxsort.

JamesNK: @eerhardt Changed to build variable: https://github.com/aspnet/Benchmarks/commit/1c98543f549e129038ac49ad5841db8325441deb

Maoni0: @MichalStrehovsky yeah, vxsort has a lot of unrolled code. do you know how big a min web app is? I think the minimum I saw was something like 15mb? @eerhardt, do you have the latest data?

eerhardt: > do you have the latest data?

https://aka.ms/aspnet/benchmarks/internal  check the "Native AOT" page on the left.

Minimal API Windows Workstation GC: 13.58 MB
Minimal API Windows Server GC: 13.81 MB

Minimal API Linux Workstation GC: 16.73 MB
Minimal API Linux Server GC: 16.98 MB

Maoni0: thanks @eerhardt!

PR #1792, opened 2023-02-08
Track Stage1Aot

sebastienros:


build/regressions.config.yml
```diff
@@ -109,6 +109,18 @@ sources:
         - sebastienros
       template: reliability
       healthCheck: true
+  - name: Native AOT Reliability
+    table: NativeAotBenchmarks
```
eerhardt, line 113: In the `PublishedSize` regression above, I guess it is using `table: TrendBenchmarks`. But this one uses `NativeAotBenchmarks`.

Should I have used a different table above?

eerhardt (approved):

PR #1796, opened 2023-02-22
Update NativeAOT Benchmarks

eerhardt:
1. Cleanup the projects - remove Logging.ClearProviders.
2. Add Trim R2R SingleFile measurements to Stage1 and Stage1Grpc
3. EnableRequestDelegateGenerator for the Stage1 minimal API app

cc @sebastienros @JamesNK @captainsafia 

FYI - @mangod9 @janvorli

### DIFF
--- a/src/BenchmarksApps/BasicMinimalApi/Program.cs
+++ b/src/BenchmarksApps/BasicMinimalApi/Program.cs
@@ -2,7 +2,7 @@ using System.Text.Json.Serialization;
 using BasicMinimalApi;
 
 var builder = WebApplication.CreateSlimBuilder(args);
-builder.Logging.ClearProviders(); // Clearing for benchmark scenario, template has AddConsole();
+// No logging for benchmark scenario, template has AddConsole();
 
 builder.Services.ConfigureHttpJsonOptions(options =>
 {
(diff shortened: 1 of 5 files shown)


src/BenchmarksApps/BasicMinimalApi/Program.cs
```diff
@@ -2,7 +2,7 @@
 using BasicMinimalApi;
 
 var builder = WebApplication.CreateSlimBuilder(args);
-builder.Logging.ClearProviders(); // Clearing for benchmark scenario, template has AddConsole();
+// No logging for benchmark scenario, template has AddConsole();
```
sebastienros, line 5: So remove the explicit "Application started" output at the end of the app?

However ... not logging and not having logging provider is not the same in terms of perfs. Having logging provider registered has some perf impact because there will be custom logging scopes created still and some using of async locals. You should measure the differences for knowledge.

eerhardt: `ClearProviders()` calls `services.RemoveAll<ILoggerProvider>()`.

https://github.com/dotnet/runtime/blob/75eac9643f3d8952df3cffd928e7987887bb7dab/src/libraries/Microsoft.Extensions.Logging/src/LoggingBuilderExtensions.cs#L46-L50

When you call `CreateSlimBuilder`, there are no `ILoggerProvider` services registered. So there are no logging providers to remove:

image

So calling `ClearProviders()` doesn't do anything here.

Here's another way to prove it:

image

The number of services before and after calling `builder.Logging.ClearProviders()` is the same.

JamesNK (approved):

PR #1797, opened 2023-02-22
Fix BasicMinimalApi app with trimming

eerhardt:
The trimmer tool will strip parameter names of methods it doesn't think are being reflected upon. However, this means that RequestDelegateFactory doesn't work in a trimmed (not AOT) app. This is tracked in https://github.com/dotnet/linker/issues/3061.

Work around the trimming bug by passing `--keep-metadata parametername` so it doesn't trim parameter names. This allows RDF to work correctly.

Note that once the EnableRequestDelegateGenerator source generator works with the `dotnet new api` template, this workaround will no longer be needed.

FYI - @agocke @vitek-karas @captainsafia


eerhardt:
@sebastienros - can you please take a look and merge (since I don't have write permission).

PR #1798, opened 2023-02-22
Improve proxies http2 benchmarks

MihaZupan:
This adds `-h2load` variants of yarp, httpclient, envoy and nginx scenarios for proxies. I got Envoy and Nginx to work with h2.

Envoy seems to start spitting out some errors during high load on citrine, haven't figured out why. It's still a lot better than the current numbers on our dashboard.

I also removed the `--channel edge` in a few places.

@sebastienros I was able to test that the new crank scenarios work fine, but I have no idea how to validate the pipelines changes.
The idea is that I'd want the `-h2load` variant of the scenario to be used anytime the proxy needs to accept http2 (a way to swap the load job from `bombardier` to `h2LoadClient` in those cases)

### DIFF
--- a/build/proxies-scenarios.yml
+++ b/build/proxies-scenarios.yml
@@ -16,67 +16,78 @@ parameters:
   type: object
   default: 
   - displayName: YARP
-    arguments: --scenario proxy-yarp        $(proxyJobs) --property proxy=yarp --application.channel edge --application.framework net8.0
+    scenario: proxy-yarp
+    arguments: $(proxyJobs) --property proxy=yarp --application.framework net8.0
     supportsHttp: true
     supportsServerHttps: true
     supportsServerHttp2: true
     supportsDownstreamHttp2: true
     condition: 'true'
   - displayName: YARP gRPC
-    arguments: --scenario proxy-yarp-grpc        $(proxyGRPCJobs) --property proxy=yarp --application.channel edge --application.framework net8.0
+    scenario: proxy-yarp-grpc
+    arguments: $(proxyGRPCJobs) --property proxy=yarp --application.framework net8.0
     supportsGRPC: true
     condition: 'true'
   - displayName: YARP-net70
-    arguments: --scenario proxy-yarp        $(proxyJobs) --property proxy=yarp-net70 --application.framework net7.0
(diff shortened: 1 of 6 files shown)


build/proxies-scenarios.yml
```diff
@@ -180,5 +191,5 @@ steps:
                 "condition": "(${{ parameters.condition }}) && (${{ s.condition }})",
                 "timeout": "00:03:00",
                 "name": "crank",
-                "args": [ "--command-line-property --table ProxyBenchmarks --sql SQL_CONNECTION_STRING --chart --session $(session) ${{ parameters.arguments }} --no-metadata --no-measurements --downstream.options.reuseBuild true ${{ s.arguments }} ${{ payload.arguments }} ${{ protocol.arguments }} --property rate=0 --property cpu=0 --description \"${{ s.displayName }} ${{ payload.displayName }} ${{ protocol.displayName }} $(System.JobDisplayName)\"" ]
+                "args": [ "--command-line-property --table ProxyBenchmarks --sql SQL_CONNECTION_STRING --chart --session $(session) ${{ parameters.arguments }} --no-metadata --no-measurements --downstream.options.reuseBuild true --scenario ${{ s.scenario }}${{ protocol.scenarioSuffix }} ${{ s.arguments }} ${{ payload.arguments }} ${{ protocol.arguments }} --property rate=0 --property cpu=0 --description \"${{ s.displayName }} ${{ payload.displayName }} ${{ protocol.displayName }} $(System.JobDisplayName)\"" ]
```
sebastienros, line 194:
```suggestion
                "args": [ "--command-line-property --table ProxyBenchmarks --sql SQL_CONNECTION_STRING --chart --session $(session) ${{ parameters.arguments }} --no-metadata --no-measurements --downstream.options.reuseBuild true --scenario ${{ s.scenario }} ${{ protocol.scenarioSuffix }} ${{ s.arguments }} ${{ payload.arguments }} ${{ protocol.arguments }} --property rate=0 --property cpu=0 --description \"${{ s.displayName }} ${{ payload.displayName }} ${{ protocol.displayName }} $(System.JobDisplayName)\"" ]
```

PR #1800, opened 2023-02-27
Fix AOT published size regression configuration

sebastienros:


build/regressions.config.yml
```diff
@@ -71,6 +71,25 @@ sources:
...
+        - sebastienros
+      template: published-size
+      title: >-
+        Published size {% if Regressions.first.Change < 0 %}improvement{% else %}regression{% endif %}: {% for r in Regressions limit: 4 %}{{ r.CurrentResult.Scenario }}{% if forloop.last == false %}, {% endif %}{% endfor %}
+  - name: PublishedSizeAot
+    table: NativeAotBenchmarks
+    rules:
       - include: Stage1Aot
         owners:
           - eerhardt
```
eerhardt, line 95: Would it make sense to collapse the `owners` into a single place, now that these scenarios aren't sharing the same section anymore?

eerhardt (approved):
Thanks!

PR #1803, opened 2023-03-07
Set InvariantGlobalization in BasicMinimalApi

eerhardt:
The api template is updating, see https://github.com/dotnet/aspnetcore/pull/47066. This syncs that change here.

~@JamesNK - I didn't update the Grpc app because we aren't updating the grpc template. Do you think we should make a similar change to the grpc template?~


captainsafia (approved):

JamesNK:
Add it to BasicGrpc as well please

PR #1809, opened 2023-03-13
Update to match TechEmpower

benaadams:
Mostly is just dropping `AsciiString` since C# has caught up


sebastienros (approved):

DamianEdwards:
@sebastienros can we merge this? @roji you might need to rebase your PR after this goes in.

PR #1810, opened 2023-03-13
Flip NoPGO to PGO in Stage1

EgorBo:
These `No PGO` didn't make any sense - PGO is not enabled by default.

cc @sebastienros

### DIFF
--- a/build/nativeaot-scenarios.yml
+++ b/build/nativeaot-scenarios.yml
@@ -49,7 +49,7 @@ parameters:
     condition: 'true'
 
   - displayName: Goldilocks gRPC Stage 1 (CoreCLR - PGO)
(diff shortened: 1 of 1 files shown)


sebastienros:
@eerhardt you own these, better if you know what is necessary here

eerhardt (approved):
This would make sense why we were getting the same results between `Stage1` and `Stage1NoPgo`.

Note that our NativeAOT goal is "within 5% RPS of the default CoreCLR", so we won't be comparing NativeAOT to these, but it would be beneficial to get numbers on enabling PGO.

EgorBo:
> Note that our NativeAOT goal is "within 5% RPS of the default CoreCLR"

It seems to be overdone on arm already 😄 
image

eerhardt:
Yeah - we don't know why/how NativeAOT is getting better RPS numbers on Ampere Linux. If you have any insights, they would be appreciated. 😄

PR #1764, opened 2022-10-30
Use separate DB connection pools for read/update in platform update benchmark

roji:
This uses separate connection pools (data sources) for the queries and the updates in the platform updates benchmark, as suggested by @NinoFloris. Improvement isn't as significant as we hoped (and saw earlier): **3.64%**. But given the variability/unreliability of the updates benchmark, we may see a better trend on the graph after a bit of time. Note that I ran this for 60 seconds instead of the default 15.

| db                  | updates_baseline | updates_change |         |
| ------------------- | ---------------- | -------------- | ------- |
| CPU Usage (%)       |               66 |             95 | +43.94% |
| Cores usage (%)     |            1,849 |          2,657 | +43.70% |
| Working Set (MB)    |               58 |             89 | +53.45% |
| Build Time (ms)     |            1,381 |          1,371 |  -0.72% |
| Start Time (ms)     |              361 |            356 |  -1.39% |
| Published Size (KB) |          930,199 |        930,199 |   0.00% |

| application           | updates_baseline        | updates_change          |         |
| --------------------- | ----------------------- | ----------------------- | ------- |
| CPU Usage (%)         |                      60 |                      69 | +15.00% |
| Cores usage (%)       |                   1,668 |                   1,921 | +15.17% |
| Working Set (MB)      |                     624 |                     653 |  +4.65% |
| Private Memory (MB)   |                   1,268 |                   1,308 |  +3.15% |
| Build Time (ms)       |                   3,135 |                   3,176 |  +1.31% |
| Start Time (ms)       |                   1,828 |                   1,922 |  +5.14% |
| Published Size (KB)   |                 101,724 |                 101,725 |  +0.00% |
| .NET Core SDK Version | 8.0.100-alpha.1.22528.3 | 8.0.100-alpha.1.22528.3 |         |

| load                   | updates_baseline | updates_change |        |
| ---------------------- | ---------------- | -------------- | ------ |
| CPU Usage (%)          |                4 |              4 |  0.00% |
| Cores usage (%)        |              107 |            111 | +3.74% |
| Working Set (MB)       |               38 |             38 |  0.00% |
| Private Memory (MB)    |              363 |            363 |  0.00% |
| Start Time (ms)        |                0 |              0 |        |
| First Request (ms)     |               99 |            107 | +8.08% |
| Requests/sec           |           26,778 |         27,753 | +3.64% |
| Requests               |        1,609,340 |      1,667,648 | +3.62% |
[...]

PR #1816, opened 2023-03-16
Use separate DB connection pools for read/update in platform update benchmark

roji:
Picks up #1764 again, which is @NinoFloris's idea to separate read and write operations to separate pools to prevent head-of-line perf issues.

Unfortunately the volatility of the updates benchmark makes it impossible to say with clarity how this does; but it does seem to improve perf in a consistent way. I suggest merging this and tracking what it does to updates for a few days, we can decide whether to keep it or not after that.

/cc @sebastienros @ajcvickers

### DIFF
--- a/src/BenchmarksApps/TechEmpower/PlatformBenchmarks/Data/RawDb.cs
+++ b/src/BenchmarksApps/TechEmpower/PlatformBenchmarks/Data/RawDb.cs
@@ -20,7 +20,7 @@ namespace PlatformBenchmarks
             = new(new MemoryCacheOptions { ExpirationScanFrequency = TimeSpan.FromMinutes(60) });
 
 #if NET7_0_OR_GREATER
-        private readonly NpgsqlDataSource _dataSource;
+        private readonly NpgsqlDataSource _dataSource, _updateDataSource, _readOnlyDataSource;
 #else
         private readonly string _connectionString;
 #endif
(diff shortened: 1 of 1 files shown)


src/BenchmarksApps/TechEmpower/PlatformBenchmarks/Data/RawDb.cs
```diff
@@ -203,7 +210,33 @@ public async Task<World[]> LoadMultipleUpdatesRows(int count)
                     await reader.NextResultAsync();
                 }
             }
+
+            using (var connection = await _updateDataSource.OpenConnectionAsync())
+            using (var updateCmd = new NpgsqlCommand(BatchUpdateString.Query(count), connection))
```
vonzshik, line 215: Might as well just call `_updateDataSource.CreateCommand` (and the same thing above for batch).

roji: You're right. I prefer to do it separately/later though, to cleanly isolate the changes in this PR from anything else (so we hopefully get a clean perf change reading).

I've made a note and will do this later.

src/BenchmarksApps/TechEmpower/PlatformBenchmarks/Data/RawDb.cs
```diff
@@ -30,6 +30,15 @@ public RawDb(ConcurrentRandom random, AppSettings appSettings)
             _random = random;
 #if NET7_0_OR_GREATER
             _dataSource = NpgsqlDataSource.Create(appSettings.ConnectionString);
+
+            // For the update benchmark, we use two different connection pools for read/update, to avoid head-of-line
+            // perf issues.
+            var dataSourceBuilder = new NpgsqlDataSourceBuilder(appSettings.ConnectionString);
+
+            dataSourceBuilder.ConnectionStringBuilder.MaxPoolSize = 18;
```
vonzshik, line 38: Just FYI, this probably doesn't work the way you intended for it to work. Right now we pass `ConnectionStringBuilder` as is to a created data source, which caches it, That means, that by changing any of the settings after calling `Build` you're going to change the cached settings of an already built data source. Concerning `MaxPoolSize`, it works as is because we cache the actual value.

roji: Good catch @vonzshik, opened https://github.com/npgsql/npgsql/pull/5000 to track this.

roji:
There's been lots of changes recently, let's hold off on merging this for a bit, to have some stability in the dashboard.

PR #1821, opened 2023-03-23
Update to RazorSlices 0.4.0 & tweak Fortunes templates accordingly

DamianEdwards:
RazorSlices 0.4.0 will be released shortly and this is pre-emptive update to the benchmarks to react.

image

### DIFF
--- a/src/BenchmarksApps/TechEmpower/Minimal/Minimal.csproj
+++ b/src/BenchmarksApps/TechEmpower/Minimal/Minimal.csproj
@@ -11,7 +11,7 @@
   <ItemGroup>
     <PackageReference Include="Npgsql" Version="8.0.0-preview.1" />
     <PackageReference Include="Dapper" Version="2.0.123" />
-    <PackageReference Include="RazorSlices" Version="0.3.0" />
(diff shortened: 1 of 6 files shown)


sebastienros:
I mean I like it

DamianEdwards:
@sebastienros that was an oversight on my part. I'd assume we'd make the same optimization there in the spirit of the benchmarks, i.e. it's a reasonable optimization given the scenario.

DamianEdwards:
I think you and I are still in disagreement on this point :smile:

sebastienros:
Can you run the numbers on Minimal with and without the change on the template?

DamianEdwards:
> Can you run the numbers on Minimal with and without the change on the template?

Can do, but will have to wait until tomorrow.

DamianEdwards:
OK here's the full results. The impact in the minimal test is actually higher in some ways, with the optimized version being 1.3% faster on average, and only 0.99% slower than 0.3.0, vs. 2.26% slower in the unoptimized 0.4.0 version.

image

sebastienros (approved):
Makes me sad, but hey, someone has to be here and you did the work so this will me my contribution.

PR #1828, opened 2023-03-30
Adds TodosApi app (native AOT "goldilocks" stage 2)

DamianEdwards:
This adds a refactored version of the [TrimmedTodo app](https://github.com/DamianEdwards/TrimmedTodo/tree/minimalpostgres/no-jwt-swagger-validation/src/TrimmedTodo.MinimalApi.PostgreSQL) configured to align with our goals for the [native AOT "stage 2" app](https://github.com/dotnet/aspnetcore/issues/45910).

The app uses:
- Minimal APIs
- JSON serialization
- Configuration & logging
- JWT bearer authentication
- Npgsql for data access
- Exception handler middleware with problem details
- Health checks (with a check for database access and another for JWT configuration)

When published native AOT with the `-p:EnableLogging=true` property, the exe size is currently **20,135 KB** on Windows. Everything except the JWT authentication and data access works in native AOT right now, the latter will start working once the [JSON support for `IAsyncEnumerable`](https://github.com/dotnet/runtime/issues/82457) lands.

Native AOT published sizes from Crank runs:

Platform | exe size (KB)
---------- | --:
win-x64 | 28,316
linux-x64 | 31,121

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -66,24 +66,26 @@
     <MicrosoftEntityFrameworkCoreSqliteVersion50>5.0.17</MicrosoftEntityFrameworkCoreSqliteVersion50>
     <MicrosoftEntityFrameworkCoreSqliteVersion60>6.0.14</MicrosoftEntityFrameworkCoreSqliteVersion60>
     <MicrosoftEntityFrameworkCoreSqliteVersion80>8.0.0-preview.1.23111.4</MicrosoftEntityFrameworkCoreSqliteVersion80>
+
+    <MicrosoftAspNetCoreAppPackageVersion80>8.0.0-preview.3.23177.8</MicrosoftAspNetCoreAppPackageVersion80>
   </PropertyGroup>
 
-  <PropertyGroup Condition="$(TargetFramework) == 'netcoreapp2.1'">
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
     <MicrosoftAspNetCoreAppPackageVersion>2.1.3</MicrosoftAspNetCoreAppPackageVersion>
   </PropertyGroup>
-  <PropertyGroup Condition="$(TargetFramework) == 'netcoreapp2.2'">
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp2.2'">
     <MicrosoftAspNetCoreAppPackageVersion>2.2.0</MicrosoftAspNetCoreAppPackageVersion>
   </PropertyGroup>
-  <PropertyGroup Condition="$(TargetFramework) == 'netcoreapp3.0'">
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'">
     <MicrosoftAspNetCoreAppPackageVersion>3.0.0</MicrosoftAspNetCoreAppPackageVersion>
   </PropertyGroup>
-  <PropertyGroup Condition="$(TargetFramework) == 'netcoreapp3.1'">
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
     <MicrosoftAspNetCoreAppPackageVersion>3.1.0</MicrosoftAspNetCoreAppPackageVersion>
   </PropertyGroup>
-  <PropertyGroup Condition="$(TargetFramework) == 'net6.0'">
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
     <MicrosoftAspNetCoreAppPackageVersion>6.0.14</MicrosoftAspNetCoreAppPackageVersion>
   </PropertyGroup>
-  <PropertyGroup Condition="$(TargetFramework) == 'net8.0'">
-    <MicrosoftAspNetCoreAppPackageVersion>8.0.0-preview.1.23112.2</MicrosoftAspNetCoreAppPackageVersion>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
+    <MicrosoftAspNetCoreAppPackageVersion>$(MicrosoftAspNetCoreAppPackageVersion80)</MicrosoftAspNetCoreAppPackageVersion>
   </PropertyGroup>
 </Project>
--- a/scenarios/goldilocks.benchmarks.yml
+++ b/scenarios/goldilocks.benchmarks.yml
@@ -98,7 +98,6 @@ scenarios:
       buildArguments: 
         - "/p:PublishAot=true"
         - "/p:StripSymbols=true"
-        - "/p:EnableRequestDelegateGenerator=true"
     load:
       job: wrk
       variables:
--- a/scenarios/goldilocks.benchmarks.yml
+++ b/scenarios/goldilocks.benchmarks.yml
@@ -111,7 +111,7 @@ scenarios:
         - "/p:PublishAot=true"
         - "/p:StripSymbols=true"
       environmentVariables:
-        connectionString: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
+        CONNECTIONSTRINGS__TODOSDB: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
     load:
       job: wrk
       variables:
@@ -131,7 +131,7 @@ scenarios:
         - "/p:TrimMode=full"
         - "/p:EnableRequestDelegateGenerator=true"
       environmentVariables:
-        connectionString: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
+        CONNECTIONSTRINGS__TODOSDB: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
     load:
       job: wrk
       variables:
@@ -147,7 +147,7 @@ scenarios:
         - "/p:PublishAot=false"
         - "/p:EnableRequestDelegateGenerator=false"
       environmentVariables:
-        connectionString: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
+        CONNECTIONSTRINGS__TODOSDB: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
     load:
       job: wrk
       variables:
--- a/scenarios/goldilocks.benchmarks.yml
+++ b/scenarios/goldilocks.benchmarks.yml
@@ -21,6 +21,16 @@ jobs:
       serverScheme: http
       serverPort: 5000
     arguments: "--urls {{serverScheme}}://{{serverAddress}}:{{serverPort}}"
+  todosapiaspnetbenchmarks:
+    source:
+      repository: https://github.com/aspnet/benchmarks.git
+      branchOrCommit: main
+      project: src/BenchmarksApps/TodosApi/TodosApi.csproj
+    readyStateText: Application started.
+    variables:
+      serverScheme: http
+      serverPort: 5000
+    arguments: "--urls {{serverScheme}}://{{serverAddress}}:{{serverPort}}"
   basicgrpcaspnetbenchmarks:
     source:
       repository: https://github.com/aspnet/benchmarks.git
@@ -81,6 +91,47 @@ scenarios:
       variables:
         presetHeaders: json
         path: /todos
+
+  todosapipublishaot:
+    application:
+      job: todosapiaspnetbenchmarks
+      buildArguments: 
+        - "/p:PublishAot=true"
+        - "/p:StripSymbols=true"
+        - "/p:EnableRequestDelegateGenerator=true"
+    load:
+      job: wrk
+      variables:
+        presetHeaders: json
+        path: /api/todos
+
+  todosapipublishtrimr2rsinglefile:
+    application:
+      job: todosapiaspnetbenchmarks
+      buildArguments: 
+        - "/p:PublishAot=false"
+        - "/p:PublishTrimmed=true"
+        - "/p:PublishReadyToRun=true"
+        - "/p:PublishSingleFile=true"
+        - "/p:TrimMode=full"
+        - "/p:EnableRequestDelegateGenerator=true"
+    load:
+      job: wrk
+      variables:
+        presetHeaders: json
+        path: /api/todos
+
+  todosapivanilla:
+    application:
+      job: todosapiaspnetbenchmarks
+      buildArguments: 
+        - "/p:PublishAot=false"
+        - "/p:EnableRequestDelegateGenerator=false"
+    load:
+      job: wrk
+      variables:
+        presetHeaders: json
+        path: /api/todos
 
   basicgrpcpublishaot:
     application:
--- a/scenarios/goldilocks.benchmarks.yml
+++ b/scenarios/goldilocks.benchmarks.yml
@@ -52,6 +52,16 @@ jobs:
     waitForExit: false
     options:
       requiredOperatingSystem: linux
+  postgresql:
+    source:
+      repository: https://github.com/TechEmpower/FrameworkBenchmarks.git
+      branchOrCommit: master
+      dockerFile: toolset/databases/postgres/postgres.dockerfile
+      dockerImageName: postgres_te
+      dockerContextDirectory: toolset/databases/postgres
+    readyStateText: ready to accept connections
+    noClean: true
+
 
 scenarios:
 
@@ -93,11 +103,15 @@ scenarios:
         path: /todos
 
   todosapipublishaot:
+    db:
+      job: postgresql
     application:
       job: todosapiaspnetbenchmarks
       buildArguments: 
         - "/p:PublishAot=true"
         - "/p:StripSymbols=true"
+      environmentVariables:
+        connectionString: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
     load:
       job: wrk
       variables:
@@ -105,6 +119,8 @@ scenarios:
         path: /api/todos
 
   todosapipublishtrimr2rsinglefile:
+    db:
+      job: postgresql
     application:
       job: todosapiaspnetbenchmarks
       buildArguments: 
@@ -114,6 +130,8 @@ scenarios:
         - "/p:PublishSingleFile=true"
         - "/p:TrimMode=full"
         - "/p:EnableRequestDelegateGenerator=true"
+      environmentVariables:
+        connectionString: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
     load:
       job: wrk
       variables:
@@ -121,11 +139,15 @@ scenarios:
         path: /api/todos
 
   todosapivanilla:
+    db:
+      job: postgresql
     application:
       job: todosapiaspnetbenchmarks
       buildArguments: 
         - "/p:PublishAot=false"
         - "/p:EnableRequestDelegateGenerator=false"
+      environmentVariables:
+        connectionString: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
     load:
       job: wrk
       variables:
--- a/scenarios/goldilocks.benchmarks.yml
+++ b/scenarios/goldilocks.benchmarks.yml
@@ -111,7 +111,7 @@ scenarios:
         - "/p:PublishAot=true"
         - "/p:StripSymbols=true"
       environmentVariables:
-        CONNECTIONSTRINGS__TODOSDB: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
+        CONNECTIONSTRINGS__TODODB: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
     load:
       job: wrk
       variables:
@@ -131,7 +131,7 @@ scenarios:
         - "/p:TrimMode=full"
         - "/p:EnableRequestDelegateGenerator=true"
       environmentVariables:
-        CONNECTIONSTRINGS__TODOSDB: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
+        CONNECTIONSTRINGS__TODODB: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
     load:
       job: wrk
       variables:
@@ -147,7 +147,7 @@ scenarios:
         - "/p:PublishAot=false"
         - "/p:EnableRequestDelegateGenerator=false"
       environmentVariables:
-        CONNECTIONSTRINGS__TODOSDB: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
+        CONNECTIONSTRINGS__TODODB: Server={{databaseServer}};Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4
     load:
       job: wrk
       variables:
--- a/src/BenchmarksApps/TodosApi/DataExtensions.cs
+++ b/src/BenchmarksApps/TodosApi/DataExtensions.cs
@@ -86,7 +86,6 @@ internal static class DataExtensions
         }
     }
 
-
     public static Task<T?> MapSingleAsync<T>(this NpgsqlDataReader reader)
         where T : IDataReaderMapper<T>
         => MapSingleAsync(reader, T.Map);
@@ -142,10 +141,8 @@ internal static class DataExtensions
     public static Task<NpgsqlDataReader> QueryAsync(this NpgsqlCommand command, CancellationToken cancellationToken = default)
         => QueryAsync(command, CommandBehavior.Default, cancellationToken);
 
-    public static async Task<NpgsqlDataReader> QueryAsync(this NpgsqlCommand command, CommandBehavior commandBehavior, CancellationToken cancellationToken = default)
-    {
-        return await command.ExecuteReaderAsync(commandBehavior, cancellationToken);
-    }
+    public static Task<NpgsqlDataReader> QueryAsync(this NpgsqlCommand command, CommandBehavior commandBehavior, CancellationToken cancellationToken = default)
+        => command.ExecuteReaderAsync(commandBehavior, cancellationToken);
 
     public static async Task<List<T>> ToListAsync<T>(this IAsyncEnumerable<T> enumerable, int? initialCapacity = null, CancellationToken cancellationToken = default)
     {
--- a/src/BenchmarksApps/TodosApi/DataExtensions.cs
+++ b/src/BenchmarksApps/TodosApi/DataExtensions.cs
@@ -4,22 +4,6 @@ namespace Npgsql;
 
 internal static class DataExtensions
 {
-    public static async ValueTask OpenIfClosedAsync(this NpgsqlConnection connection)
-    {
-        if (connection.State == ConnectionState.Closed)
-        {
-            await connection.OpenAsync();
-        }
-    }
-
-    public static async Task<int> ExecuteAsync(this NpgsqlConnection connection, string commandText)
-    {
-        await using var cmd = connection.CreateCommand(commandText);
-
-        await connection.OpenIfClosedAsync();
-        return await cmd.ExecuteNonQueryAsync();
-    }
-
     public static async Task<int> ExecuteAsync(this NpgsqlDataSource dataSource, string commandText)
     {
         await using var cmd = dataSource.CreateCommand(commandText);
@@ -34,26 +18,10 @@ internal static class DataExtensions
         return await cmd.ExecuteScalarAsync(CancellationToken.None);
     }
 
-    public static async Task<int> ExecuteAsync(this NpgsqlConnection connection, string commandText, params NpgsqlParameter[] parameters)
-    {
-        await using var cmd = connection.CreateCommand(commandText, parameters);
-
-        await connection.OpenIfClosedAsync();
-        return await cmd.ExecuteNonQueryAsync();
-    }
-
     public static async Task<int> ExecuteAsync(this NpgsqlDataSource dataSource, string commandText, params NpgsqlParameter[] parameters)
     {
         await using var cmd = dataSource.CreateCommand(commandText, parameters);
 
-        return await cmd.ExecuteNonQueryAsync();
-    }
-
-    public static async Task<int> ExecuteAsync(this NpgsqlConnection connection, string commandText, Action<NpgsqlParameterCollection> configureParameters)
-    {
-        await using var cmd = connection.CreateCommand(commandText, configureParameters);
-
-        await connection.OpenIfClosedAsync();
         return await cmd.ExecuteNonQueryAsync();
     }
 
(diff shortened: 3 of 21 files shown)


src/BenchmarksApps/TodosApi/appsettings.json
```diff
@@ -0,0 +1,19 @@
...
+    }
+  },
+  "AllowedHosts": "*",
+  "Authentication": {
+    "Schemes": {
+      "Bearer": {
+        "ValidAudiences": [
+          "http://localhost:5054"
+        ],
+        "ValidIssuer": "dotnet-user-jwts"
```
eerhardt, line 15: (nit) do we want `dotnet-user-jwts` in the production appsettings.json?  or just in development.json?

src/BenchmarksApps/TodosApi/TodosApi.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <ServerGarbageCollection>false</ServerGarbageCollection>
+    <InvariantGlobalization>true</InvariantGlobalization>
+    <UserSecretsId>b8ffb8d3-b768-460b-ac1f-ef267c954c85</UserSecretsId>
+    <EnableRequestDelegateGenerator>true</EnableRequestDelegateGenerator>
```
eerhardt, line 10: You shouldn't need this anymore...

src/BenchmarksApps/TodosApi/TodosApi.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <PublishRelease>true</PublishRelease>
```
eerhardt, line 11: This isn't needed. It is the default for 8.0 projects.

src/BenchmarksApps/TodosApi/DataExtensions.cs
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    public static async Task<int> ExecuteAsync(this NpgsqlDataSource dataSource, string commandText, params NpgsqlParameter[] parameters)
+    {
+        await using var cmd = dataSource.CreateCommand(commandText, parameters);
+
+        return await cmd.ExecuteNonQueryAsync();
+    }
+
+    public static async Task<int> ExecuteAsync(this NpgsqlConnection connection, string commandText, Action<NpgsqlParameterCollection> configureParameters)
```
eerhardt, line 52: I didn't think we were using `NpgsqlConnection` anymore. It doesn't look like a bunch of these overloads are used. Can we delete any methods that aren't being used?

At least all the `NpgsqlConnection` methods can be removed since we use DataSource API now.

src/BenchmarksApps/TodosApi/appsettings.json
DamianEdwards: It's mostly for convenience for this test/benchmark app. If we don't use `dotnet-user-jwts` we need to create another thing that produces JWTs with a different issuer name.

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: It is? I thought it was going to be set in templates rather than defaulted based on TFM, hmmm.

src/BenchmarksApps/TodosApi/Program.cs
```diff
@@ -0,0 +1,73 @@
+using Microsoft.Extensions.Logging.Configuration;
+using Npgsql;
+using TodosApi;
+
+var builder = WebApplication.CreateSlimBuilder(args);
+
+#if ENABLE_LOGGING
```
eerhardt, line 7: Seems weird that `ENABLE_LOGGING` turns off appsettings.json, which is needed for AddJwtBearer right?

src/BenchmarksApps/TodosApi/TodosApi.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <ServerGarbageCollection>false</ServerGarbageCollection>
+    <InvariantGlobalization>true</InvariantGlobalization>
+    <UserSecretsId>b8ffb8d3-b768-460b-ac1f-ef267c954c85</UserSecretsId>
+    <PublishAot>true</PublishAot>
+    <EnableLogging Condition="$(Configuration.StartsWith('Debug'))">true</EnableLogging>
+    <DefineConstants Condition=" '$(EnableLogging)' == 'true' ">$(DefineConstants);ENABLE_LOGGING</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="8.0.0-preview.3.23177.8" />
```
eerhardt, line 16: is there a way to keep this version up to date with daily builds?

cc @sebastienros

src/BenchmarksApps/TodosApi/Program.cs
DamianEdwards: Just fixed that :)

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: We can use a wildcard, e.g. `8.0.0-preview.*`?

DamianEdwards: Seems there's a `$(MicrosoftAspNetCoreAppPackageVersion)` property defined, I'll try using that.

DamianEdwards: Oh it's hard-coded to a preview.1 version 😱

eerhardt: I think that might get replaced by crank.... @sebastienros ?

src/BenchmarksApps/TodosApi/TodosApi.csproj
eerhardt: https://github.com/dotnet/sdk/pull/29155

src/BenchmarksApps/TodosApi/appsettings.json
```diff
@@ -0,0 +1,19 @@
...
+      "Default": "Information",
+      "Microsoft.AspNetCore": "Warning"
+    }
+  },
+  "AllowedHosts": "*",
+  "Authentication": {
+    "Schemes": {
+      "Bearer": {
+        "ValidAudiences": [
+          "http://localhost:5054"
```
eerhardt, line 13: Is there a way that `5054` is always used? I thought that was in the launchsettings.json or something?

src/BenchmarksApps/TodosApi/appsettings.Development.json
```diff
@@ -0,0 +1,18 @@
...
+  "Authentication": {
+    "Schemes": {
+      "Bearer": {
+        "ValidAudiences": [
+          "http://localhost:5054"
+        ],
+        "ValidIssuer": "dotnet-user-jwts"
+      }
+    }
+  }
```
eerhardt, line 17: Would it make sense to not duplicate this section, since it is in the appsettings.json?

src/BenchmarksApps/TodosApi/TodosApi.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <ServerGarbageCollection>false</ServerGarbageCollection>
+    <InvariantGlobalization>true</InvariantGlobalization>
+    <UserSecretsId>b8ffb8d3-b768-460b-ac1f-ef267c954c85</UserSecretsId>
+    <PublishAot>true</PublishAot>
```
eerhardt, line 10: In the BasicMinimalApi app we pass this during the build. I kind of like having it in the .csproj though, for ease of development (it's what we are trying to achieve).

Do we plan on trying non-AOT scenarios in the benchmarks?

src/BenchmarksApps/TodosApi/Program.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    app.UseExceptionHandler();
+}
+
+app.MapHealthChecks("/health");
+
+app.MapTodoApi();
+
+// Enables testing request exception handling behavior
+app.MapGet("/throw", (HttpContext httpContext) => throw new InvalidOperationException("You hit the throw endpoint"));
```
eerhardt, line 69:
```suggestion
app.MapGet("/throw", () => throw new InvalidOperationException("You hit the throw endpoint"));
```

is that needed?

src/BenchmarksApps/TodosApi/Program.cs
```diff
@@ -0,0 +1,76 @@
...
+
+// Configure authentication & authorization
+builder.Services.AddAuthentication()
+    .AddJwtBearer(JwtConfiguration.ConfigureJwtBearer(builder));
+
+builder.Services.AddAuthorization();
+
+// Configure data access
+var connectionString = builder.Configuration.GetConnectionString("TodoDb")
+    ?? builder.Configuration["CONNECTION_STRING"]
```
eerhardt, line 33: Why do we check this one? But the exception only mentions `ConnectionStrings:TodoDb`

src/BenchmarksApps/TodosApi/DatabaseHealthCheck.cs
```diff
@@ -0,0 +1,24 @@
...
+    private readonly NpgsqlDataSource _dataSource;
+
+    public DatabaseHealthCheck(NpgsqlDataSource dataSource)
+    {
+        _dataSource = dataSource;
+    }
+
+    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+    {
+        var result = await _dataSource.ExecuteScalarAsync("SELECT COUNT(id) FROM public.todos");
```
eerhardt, line 17: (nit) `Select top 1` ? Do we really need the total count to ensure the DB is healthy?

eerhardt (commented):
Looking good!

Just need the yaml stuff and the NuGet package version figured out and I think we are golden

src/BenchmarksApps/TodosApi/appsettings.json
DamianEdwards: Right, I copied it from launchsettings.json for this app, but I should likely add http://localhost:5000 to cover the published case too.

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: I added it for the same reason you mention, i.e. having it on during development. We'll still be measuring non-AOT in benchmarks to calculate the delta between CoreCLR vs. native AOT.

src/BenchmarksApps/TodosApi/Program.cs
DamianEdwards: I was getting a compilation error without it, which might be a RDG issue?

src/BenchmarksApps/TodosApi/Program.cs
DamianEdwards: Yeah that's a carry over from TrimmedTodo. We can just simplify and go with the one key.

src/BenchmarksApps/TodosApi/DatabaseHealthCheck.cs
DamianEdwards: Nope, but it's indexed so it doesn't really matter. It's really just a nominal check to ensure the table is there. Happy to go with whatever folks think is the best approach for a DB health check here.

eerhardt: https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-7.0#database-probe

suggests `select 1`. If we want to ensure the table is there, maybe 

```
SELECT id FROM public.todos LIMIT 1
```

?

eerhardt: cc @roji

DamianEdwards: I updated it to simply perform a `LIMIT 1` query against the relevant table

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: I set this back to `false` for now as the data access doesn't work due to the JSON issue.

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: Yeah I've noticed the projects aren't always set up for ease of local use so I made some tweaks to improve that as part of this PR now, i.e. ensuring there are appropriate package versions set.

src/BenchmarksApps/TodosApi/Program.cs
DamianEdwards: https://github.com/dotnet/aspnetcore/issues/47275

src/BenchmarksApps/TodosApi/TodosApi.csproj
sebastienros: You can have a default version set for local development, and fallback to $(MicrosoftAspNetCoreAppPackageVersion) if it's defined. Crank will set it on builds. There is the same one for the runtime.

build/dependencies.props
```diff
@@ -66,24 +66,26 @@
...
     <MicrosoftAspNetCoreAppPackageVersion>3.1.0</MicrosoftAspNetCoreAppPackageVersion>
   </PropertyGroup>
-  <PropertyGroup Condition="$(TargetFramework) == 'net6.0'">
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
     <MicrosoftAspNetCoreAppPackageVersion>6.0.14</MicrosoftAspNetCoreAppPackageVersion>
   </PropertyGroup>
-  <PropertyGroup Condition="$(TargetFramework) == 'net8.0'">
-    <MicrosoftAspNetCoreAppPackageVersion>8.0.0-preview.1.23112.2</MicrosoftAspNetCoreAppPackageVersion>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
+    <MicrosoftAspNetCoreAppPackageVersion>$(MicrosoftAspNetCoreAppPackageVersion80)</MicrosoftAspNetCoreAppPackageVersion>
```
sebastienros, line 89: Works, but why treat net8 differently with a separate constant?

DamianEdwards: I added a net8 variant so it could be referenced elsewhere when a `TargetFramework` is not passed in from the outside (MSBuild order or evaluation, etc.).

src/BenchmarksApps/TodosApi/DatabaseHealthCheck.cs
```diff
@@ -0,0 +1,24 @@
...
+    private readonly NpgsqlDataSource _dataSource;
+
+    public DatabaseHealthCheck(NpgsqlDataSource dataSource)
+    {
+        _dataSource = dataSource;
+    }
+
+    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+    {
+        var result = await _dataSource.ExecuteScalarAsync("SELECT id FROM public.todos LIMIT 1");
```
vonzshik, line 17: Shouldn't it be wrapped in try/catch?

src/BenchmarksApps/TodosApi/TodosApi.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <ServerGarbageCollection>false</ServerGarbageCollection>
```
vonzshik, line 7: Any particular reason server garbage collection is disabled?

src/BenchmarksApps/TodosApi/DataExtensions.cs
```diff
@@ -0,0 +1,192 @@
+﻿using System.Data;
+
+namespace Npgsql;
+
+internal static class DataExtensions
+{
+    public static async Task<int> ExecuteAsync(this NpgsqlDataSource dataSource, string commandText)
```
vonzshik, line 7: Any particular reason `CancellationToken` isn't passed anywhere here?

src/BenchmarksApps/TodosApi/DatabaseHealthCheck.cs
DamianEdwards: Good point!

src/BenchmarksApps/TodosApi/DataExtensions.cs
DamianEdwards: Not really, just 'cos I'm lazy. It's good to add them though.

vonzshik: Yeah, passing `CancellationToken` can be quite a pain

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: These projects associated with our native AOT work are using Workstation GC until some on-going work in Server GC is comlpeted later in .NET 8

vonzshik: I see. Is it possible to find an issue about that on-going work?

src/BenchmarksApps/TodosApi/DataExtensions.cs
DamianEdwards: I updated it to accept and pass `CancellationToken`

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: I humbly suggest we shouldn't be relying on anything local being manually configured to get a good dev experience. The solution/projects should be openable locally after cloning without getting warnings.

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: dotnet/runtime#79658

scenarios/goldilocks.benchmarks.yml
```diff
@@ -82,6 +92,47 @@ scenarios:
         presetHeaders: json
         path: /todos
 
+  todosapipublishaot:
+    application:
+      job: todosapiaspnetbenchmarks
+      buildArguments: 
+        - "/p:PublishAot=true"
+        - "/p:StripSymbols=true"
+        - "/p:EnableRequestDelegateGenerator=true"
```
eerhardt, line 101:
```suggestion
```

Don't need this anymore for PublishAot.

src/BenchmarksApps/TodosApi/DatabaseHealthCheck.cs
```diff
@@ -0,0 +1,34 @@
...
+            result = await _dataSource.ExecuteScalarAsync("SELECT id FROM public.todos LIMIT 1", cancellationToken);
+        }
+        catch (Exception ex)
+        {
+            exception = ex;
+        }
+        
+        return result switch
+        {
+            null or int => HealthCheckResult.Healthy("Database health verified successfully"),
```
eerhardt, line 30: How does `null` indicate healthy?  no records?

src/BenchmarksApps/TodosApi/TodoApi.cs
```diff
@@ -0,0 +1,106 @@
...
+namespace Microsoft.AspNetCore.Routing;
+
+internal static class TodoApi
+{
+    public static RouteGroupBuilder MapTodoApi(this IEndpointRouteBuilder routes)
+    {
+        var group = routes.MapGroup("/api/todos");
+
+        group.MapGet("/", (NpgsqlDataSource db, CancellationToken ct) => db.QueryAsync<Todo>("SELECT * FROM Todos", ct))
+            .WithName("GetAllTodos");
```
eerhardt, line 15: What does the `WithName` do? Do we need it?

eerhardt (approved):
LGTM. This looks like a great starting app to work off.

src/BenchmarksApps/TodosApi/DatabaseHealthCheck.cs
DamianEdwards: Yep, query succeeded but no records

src/BenchmarksApps/TodosApi/TodoApi.cs
DamianEdwards: It sets the route name and endpoint name metadata which can be used for route resolution, URL generation, Open API metadata, etc. Not strictly required in this app at the moment as we aren't doing any of those things.

src/BenchmarksApps/TodosApi/TodosApi.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <InvariantGlobalization>true</InvariantGlobalization>
+    <UserSecretsId>b8ffb8d3-b768-460b-ac1f-ef267c954c85</UserSecretsId>
+    <PublishAot>false</PublishAot>
+    <EnableLogging Condition="$(Configuration.StartsWith('Debug'))">true</EnableLogging>
+    <DefineConstants Condition=" '$(EnableLogging)' == 'true' ">$(DefineConstants);ENABLE_LOGGING</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="8.0.0-preview.*" />
+    <PackageReference Include="Npgsql" Version="8.0.0-preview.2" />
```
captainsafia, line 17: Is it intentionally that we are pinning this to preview2 instead of the latest preview?

captainsafia (approved):

src/BenchmarksApps/TodosApi/TodosApi.csproj
vonzshik: That issue doesn't have that good of a description, but it seems to be about improving GC (adjust heap count on load, better handle burst load) and has nothing to do with native aot. So I fail to see why we're not going to use Server GC.

src/BenchmarksApps/TodosApi/DatabaseHealthCheck.cs
roji: Yeah, for pure health checks a `SELECT 1` would be the usual thing, no need to even access any particular table. That's different from a check whether your schema exists, which isn't a periodic health check thing but more of a startup thing.

Probably doesn't matter a great deal in any case here.

JamesNK:
Interested in seeing RPS with workstation GC on APIs that do some work (auth and/or data access)

src/BenchmarksApps/TodosApi/DatabaseHealthCheck.cs
```diff
@@ -0,0 +1,34 @@
...
+            _ => HealthCheckResult.Unhealthy("Error occurred when checking database health", exception: exception)
+        };
```
NinoFloris, line 32: This check should probably be more like this?

```suggestion
        return exception switch
        {
            null => HealthCheckResult.Healthy("Database health verified successfully"),
            _ => HealthCheckResult.Unhealthy("Error occurred when checking database health", exception: exception)
        };
```

roji (approved):

src/BenchmarksApps/TodosApi/TodosApi.csproj
vonzshik: I'm afraid I still don't follow. Since ideally we want to test with server GC only (as it's intended for server apps), to me it makes sense to already start testing with it, so we can notice any possible regressions. A good example here is npgsql, which isn't trimming safe yet, but we still use the latest preview.

DamianEdwards: We do test with server GC as well. That's defined in the `goldilocks.benchmarks.yml` file.

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: @sebastienros are you suggesting we should change all the package version references in the app project files to hard-code a default version (can include a wildcard) but prefer `$(MicrosoftAspNetCoreAppPackageVersion)` if it's set? It's a shame we can't easily make that more elegant in MSBuild but it will certainly achieve what I think we're looking for.

src/BenchmarksApps/TodosApi/TodosApi.csproj
vonzshik: You probably meant `nativeaot-scenarios.yml` which indeed enables server GC. Which makes even less sense as to why disable it here. Just so people running it locally will see pretty numbers?

DamianEdwards:
Numbers from a manual run on `aspnet-perf` environment:

| db                  | todosapi_vanilla | todosapi_trimmedr2r |         | todosapi_aot |         | todosapi_aot_win |          |
| ------------------- | ---------------- | ------------------- | ------- | ------------ | ------- | ---------------- | -------- |
| CPU Usage (%)       |               55 |                   0 |         |            0 |         |                1 |  -98.18% |
| Cores usage (%)     |              660 |                   6 | -99.09% |            6 | -99.09% |               13 |  -98.03% |
| Working Set (MB)    |              457 |                  22 | -95.19% |           23 | -94.97% |               23 |  -94.97% |
| Build Time (ms)     |            1,462 |               1,433 |  -1.98% |        1,438 |  -1.64% |           10,887 | +644.66% |
| Start Time (ms)     |            6,667 |               6,642 |  -0.37% |        6,564 |  -1.54% |            6,471 |   -2.94% |
| Published Size (KB) |          369,845 |             369,845 |   0.00% |      369,845 |   0.00% |          369,845 |    0.00% |


| application                             | todosapi_vanilla          | todosapi_trimmedr2r       |             | todosapi_aot              |              | todosapi_aot_win          |              |
| --------------------------------------- | ------------------------- | ------------------------- | ----------- | ------------------------- | ------------ | ------------------------- | ------------ |
| CPU Usage (%)                           |                        66 |                        89 |     +34.85% |                        73 |      +10.61% |                        25 |      -62.12% |
| Cores usage (%)                         |                       798 |                     1,067 |     +33.71% |                       877 |       +9.90% |                       305 |      -61.78% |
| Working Set (MB)                        |                       143 |                        92 |     -35.66% |                        64 |      -55.24% |                        64 |      -55.24% |
| Private Memory (MB)                     |                       404 |                       354 |     -12.38% |                       303 |      -25.00% |                        48 |      -88.12% |
| Build Time (ms)                         |                     2,193 |                    19,467 |    +787.69% |                    41,259 |   +1,781.40% |                    42,156 |   +1,822.30% |
| Start Time (ms)                         |                       250 |                       169 |     -32.40% |                        57 |      -77.20% |                       128 |      -48.80% |
| Published Size (KB)                     |                    96,256 |                    55,266 |     -42.58% |                   126,380 |      +31.30% |                   159,640 |      +65.85% |
| Symbols Size (KB)                       |                        56 |                        71 |     +26.79% |                    95,259 | +170,005.36% |                   131,324 | +234,407.14% |
| .NET Core SDK Version                   | 8.0.100-preview.3.23178.7 | 8.0.100-preview.3.23178.7 |             | 8.0.100-preview.3.23178.7 |              | 8.0.100-preview.3.23178.7 |              |
| Max CPU Usage (%)                       |                        66 |                        89 |     +34.53% |                         0 |              |                         0 |              |
| Max Working Set (MB)                    |                       150 |                        96 |     -35.96% |                         0 |              |                         0 |              |
| Max GC Heap Size (MB)                   |                        32 |                        11 |     -64.85% |                         0 |              |                         0 |              |
| Size of committed memory by the GC (MB) |                        46 |                        17 |     -62.99% |                         0 |              |                         0 |              |
| Max Number of Gen 0 GCs / sec           |                    146.00 |                     54.00 |     -63.01% |                      0.00 |              |                      0.00 |              |
| Max Number of Gen 1 GCs / sec           |                     84.00 |                      3.00 |     -96.43% |                      0.00 |              |                      0.00 |              |
| Max Number of Gen 2 GCs / sec           |                     16.00 |                      1.00 |     -93.75% |                      0.00 |              |                      0.00 |              |
| Max Time in GC (%)                      |                     71.00 |                      9.00 |     -87.32% |                      0.00 |              |                      0.00 |              |
| Max Gen 0 Size (B)                      |                   851,600 |                   368,736 |     -56.70% |                         0 |              |                         0 |              |
| Max Gen 1 Size (B)                      |                 9,666,784 |                 1,080,224 |     -88.83% |                         0 |              |                         0 |              |
| Max Gen 2 Size (B)                      |                27,504,240 |                 4,934,360 |     -82.06% |                         0 |              |                         0 |              |
| Max LOH Size (B)                        |                    87,464 |                         0 |             |                         0 |              |                         0 |              |
| Max POH Size (B)                        |                 1,182,944 |                 1,144,816 |      -3.22% |                         0 |              |                         0 |              |
| Max Allocation Rate (B/sec)             |               766,292,824 |               330,571,720 |     -56.86% |                         0 |              |                         0 |              |
| Max GC Heap Fragmentation               |                        50 |                        36 |     -27.54% |                         0 |              |                         0 |              |
| # of Assemblies Loaded                  |                       115 |                        98 |     -14.78% |                         0 |              |                         0 |              |
| Max Exceptions (#/s)                    |                       470 |                   119,722 | +25,372.77% |                         0 |              |                         0 |              |
| Max Lock Contention (#/s)               |                        55 |                        27 |     -50.91% |                         0 |              |                         0 |              |
| Max ThreadPool Threads Count            |                        26 |                        24 |      -7.69% |                         0 |              |                         0 |              |
| Max ThreadPool Queue Length             |                       140 |                       215 |     +53.57% |                         0 |              |                         0 |              |
| Max ThreadPool Items (#/s)              |                   104,825 |                    33,330 |     -68.20% |                         0 |              |                         0 |              |
| Max Active Timers                       |                       232 |                         2 |     -99.14% |                         0 |              |                         0 |              |
| IL Jitted (B)                           |                   566,912 |                   317,286 |     -44.03% |                         0 |              |                         0 |              |
| Methods Jitted                          |                     6,052 |                     2,857 |     -52.79% |                         0 |              |                         0 |              |


| load                   | todosapi_vanilla | todosapi_trimmedr2r |          | todosapi_aot |         | todosapi_aot_win |          |
| ---------------------- | ---------------- | ------------------- | -------- | ------------ | ------- | ---------------- | -------- |
| CPU Usage (%)          |               25 |                  15 |  -40.00% |           32 | +28.00% |               24 |   -4.00% |
| Cores usage (%)        |              295 |                 183 |  -37.97% |          385 | +30.51% |              290 |   -1.69% |
| Working Set (MB)       |               58 |                  48 |  -17.24% |           49 | -15.52% |               58 |    0.00% |
| Private Memory (MB)    |              358 |                 358 |    0.00% |          358 |   0.00% |              358 |    0.00% |
| Build Time (ms)        |            9,129 |                   0 |          |            0 |         |            9,313 |   +2.02% |
| Start Time (ms)        |                0 |                   0 |          |            0 |         |                0 |          |
| Published Size (KB)    |           73,161 |                   0 |          |            0 |         |           73,161 |    0.00% |
| Symbols Size (KB)      |                0 |                   0 |          |            0 |         |                0 |          |
| .NET Core SDK Version  |          7.0.202 |                   0 |          |            0 |         |          7.0.202 |          |
| First Request (ms)     |              166 |                 100 |  -39.76% |           57 | -65.66% |               70 |  -57.83% |
| Requests/sec           |           64,310 |              29,647 |  -53.90% |      106,569 | +65.71% |           70,403 |   +9.47% |
| Requests               |          969,473 |             446,886 |  -53.90% |    1,606,633 | +65.72% |        1,062,730 |   +9.62% |
| Mean latency (ms)      |             3.97 |                8.58 | +116.12% |         2.39 | -39.80% |             5.09 |  +28.21% |
| Max latency (ms)       |            24.11 |               20.11 |  -16.59% |        20.88 | -13.40% |            66.99 | +177.85% |
| Bad responses          |                0 |             446,886 |      +∞% |    1,606,633 |     +∞% |        1,062,730 |      +∞% |
| Socket errors          |                0 |                   0 |          |            0 |         |                0 |          |
| Read throughput (MB/s) |            32.69 |               11.31 |  -65.40% |        40.65 | +24.35% |            24.78 |  -24.20% |
| Latency 50th (ms)      |             4.57 |                8.47 |  +85.34% |         1.92 | -57.99% |             2.55 |  -44.20% |
| Latency 75th (ms)      |             5.88 |                9.51 |  +61.73% |         3.36 | -42.86% |             8.81 |  +49.83% |
| Latency 90th (ms)      |             6.79 |               10.11 |  +48.90% |         3.71 | -45.36% |            13.82 | +103.53% |
| Latency 99th (ms)      |             8.57 |               12.65 |  +47.61% |         5.22 | -39.09% |            21.23 | +147.72% |

vonzshik:
Judging by requests being equal to bad responses, it doesn't seem to work.

DamianEdwards:
>Judging by requests being equal to bad responses, it doesn't seem to work.

For the non-vanilla runs you're right 😞  Hopefully something straightforward.

DamianEdwards:
@vonzshik ah that's expected actually! Right now JSON serialization doesn't work with `IAsyncEnumerable` when trimming is enabled. That's going to be fixed by dotnet/runtime#82457

DamianEdwards:
The JSON support for `IAsyncEnumerable` is actually in the latest runtime and installer builds, but Npgsql is currently broken until [this runtime fix](https://github.com/dotnet/runtime/commit/3231e4fbfbef268b1647bf6f695e6b42cfb2117d) lands in a daily SDK installer build. Once that happens this app and the crank runs should hopefully start working for the timmed and native AOT scenarios.

vonzshik:
> Once that happens this app and the crank runs should hopefully start working for the timmed and native AOT scenarios.

If that was indeed the case, wouldn't non-trimmed also fail? + judging by the first published results (on powerbi), crank does use the runtime with the fix, but there are still errors.

DamianEdwards:
@vonzshik there are two issues. The first is that when trimming or native AOT is enabled, the JSON source generator doesn't support `IAsyncEnumerable`. The second is the runtime issue with impacting Npgsql itself which I understand is fixed in latest runtime builds. I'll dig in a bit more tomorrow to get a clearer picture of which versions are being used and to confirm the actual exceptions occurring in the runs.

DamianEdwards:
Seems there's an issue with the runtime support for JSON serializing `IAsyncEnumerable` that was added in preview.3. Issue at dotnet/runtime#84260

vonzshik:
I suppose we change to IEnnumerable for now and hopefully everything works?

src/BenchmarksApps/TodosApi/TodosApi.csproj
sebastienros: I think you can keep have something like 
```
<MicrosoftAspNetCoreAuthenticationJwtBearerVersion>8.0.0-preview.3.23177.8</MicrosoftAspNetCoreAuthenticationJwtBearerVersion>
<MicrosoftAspNetCoreAuthenticationJwtBearerVersion Condition="'$(MicrosoftAspNetCoreAppPackageVersion)' != ''">$(MicrosoftAspNetCoreAppPackageVersion)</MicrosoftAspNetCoreAuthenticationJwtBearerVersion>

```

This way it will build locally, but use the version aligned with aspnet when using crank.

DamianEdwards:
@vonzshik I'm inclined to leave it as is for now and focus on getting the underlying issue fixed (seems it's actually ASP.NET Core that needs a code change to fix this).

vonzshik:
Well then, let's pray that there are no other issues and fixing that isn't going to take long...

PR #1832, opened 2023-04-06
Added all the pertinent GC Scenarios

mrsharm:
Added all the pertinent GC Scenarios and added documentation for the ASP.NET Dashboard.

CC: @sebastienros

### DIFF
--- a/build/gc-scenarios.yml
+++ b/build/gc-scenarios.yml
@@ -16,36 +16,36 @@ parameters:
 
   # Server Runs.
   - displayName: 0gb Server
-    arguments: '--scenario 0gb --application.environmentVariables DOTNET_gcServer=1 --application.environmentVariables \"DOTNET_GCHeapCount={{ cores | minus: 2 }}\" --property scenario=0gb'
+    arguments: '--scenario 0gb --application.environmentVariables DOTNET_gcServer=1 --application.environmentVariables \"DOTNET_GCHeapCount={{ cores | minus: 2 }}\" --property scenario=0gb-server'
     condition: 'true'
 
   - displayName: 2gb Server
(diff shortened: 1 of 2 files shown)


build/gc-scenarios.yml
```diff
@@ -14,8 +14,38 @@ parameters:
...
+    arguments: '--scenario 2gb --application.environmentVariables DOTNET_gcServer=1 --application.environmentVariables \"DOTNET_GCHeapCount={{ cores | minus: 2 }}\" --property scenario=2gb-pinning'
+    condition: 'true'
+
+  - displayName: 20gb Server
+    arguments: '--scenario 20gb --application.environmentVariables DOTNET_gcServer=1 --application.environmentVariables \"DOTNET_GCHeapCount={{ cores | minus: 2 }}\" --property scenario=20gb'
+    condition: 'true'
+
+  # Workstation Runs.
+  - displayName: 0gb Workstation
+    arguments: '--scenario 0gb --application.environmentVariables DOTNET_gcServer=0 --property scenario=0gb --application.variables.tc 2'
```
sebastienros, line 36: The `--scenario 0gb` argument is to tell crank which scenario from the config file to run.
The `--property scenario=0gb` is to store a `{ "scenario": "0gb" }` element in the database, that we use to "display" the result for this line in POwerBI.

You can have multiple runs that use `--scenario 0gb` and then add some variation (like GC server/Workstation) but you can have multiple runs that have the same set of custom properties. In this case you have multiple runs with `--property scenario=0gb` and no other properties to differentiate them. How do you want PowerBI to show each of them separately then?

There are several options: 
- Use a distinct property named scenario, like `0gb-server`, `0gb-workstation`, ... and these will appear on the list of available scenario properties on the right.
- Add another property named `gc=workstation` or `gc=server`, and then change powerbi to show the list of available scenario properties, and also a list of available gc properties. The risk is that maybe you will add scenarios that don't have a gc property, and the UI will hide some of them without users understanding why if they don't know that a specific scenario has no gc property

PR #1833, opened 2023-04-12
Add validation to TodosApi & workaround JSON unspeakable issue

DamianEdwards:
This makes the stage 2 TodosApi project work properly when published native AOT by working around the JSON unspeakble types issue for now. It also adds a validation filter based on `IValidatableObject` which is much easier to make native AOT friendly.

Published exe size on win-x64 is 29,528,064 bytes with these changes.

### DIFF
--- a/src/BenchmarksApps/TodosApi/ProducesResponseTypeMetadata.cs
+++ b/src/BenchmarksApps/TodosApi/ProducesResponseTypeMetadata.cs
@@ -1,0 +1,89 @@
+﻿using Microsoft.AspNetCore.Http.Metadata;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.Http;
+
+internal sealed class ProducesResponseTypeMetadata : IProducesResponseTypeMetadata
+{
+    private readonly IEnumerable<string> _contentTypes;
+
+    public ProducesResponseTypeMetadata(int statusCode)
+        : this(typeof(void), statusCode, Enumerable.Empty<string>())
+    {
+    }
+
+    public ProducesResponseTypeMetadata(Type type, int statusCode)
+    {
+        Type = type ?? throw new ArgumentNullException(nameof(type));
+        StatusCode = statusCode;
+        _contentTypes = Enumerable.Empty<string>();
+    }
+
+    public ProducesResponseTypeMetadata(int statusCode, string contentType)
+    {
+        StatusCode = statusCode;
+        MediaTypeHeaderValue.Parse(contentType);
+        _contentTypes = GetContentTypes(contentType, Array.Empty<string>());
+    }
+
+    public ProducesResponseTypeMetadata(Type type, int statusCode, string contentType, params string[] additionalContentTypes)
+    {
+        ArgumentNullException.ThrowIfNull(contentType);
+
+        Type = type ?? throw new ArgumentNullException(nameof(type));
+        StatusCode = statusCode;
+
+        MediaTypeHeaderValue.Parse(contentType);
+        for (var i = 0; i < additionalContentTypes.Length; i++)
+        {
+            MediaTypeHeaderValue.Parse(additionalContentTypes[i]);
+        }
+
+        _contentTypes = GetContentTypes(contentType, additionalContentTypes);
+    }
+
+    // Only for internal use where validation is unnecessary.
+    private ProducesResponseTypeMetadata(Type? type, int statusCode, IEnumerable<string> contentTypes)
+    {
+
+        Type = type;
+        StatusCode = statusCode;
+        _contentTypes = contentTypes;
+    }
+
+    /// <summary>
+    /// Gets or sets the type of the value returned by an action.
+    /// </summary>
+    public Type? Type { get; set; }
+
+    /// <summary>
+    /// Gets or sets the HTTP status code of the response.
+    /// </summary>
+    public int StatusCode { get; set; }
+
+    public IEnumerable<string> ContentTypes => _contentTypes;
+
+    //internal static ProducesResponseTypeMetadata CreateUnvalidated(Type? type, int statusCode, IEnumerable<string> contentTypes) => new(type, statusCode, contentTypes);
+
+    private static List<string> GetContentTypes(string contentType, string[] additionalContentTypes)
+    {
+        var contentTypes = new List<string>(additionalContentTypes.Length + 1);
+        ValidateContentType(contentType);
+        contentTypes.Add(contentType);
+        foreach (var type in additionalContentTypes)
+        {
+            ValidateContentType(type);
+            contentTypes.Add(type);
+        }
+
+        return contentTypes;
+
+        static void ValidateContentType(string type)
+        {
+            if (type.Contains('*', StringComparison.OrdinalIgnoreCase))
+            {
+                throw new InvalidOperationException($"Could not parse '{type}'. Content types with wildcards are not supported.");
+            }
+        }
+    }
+}
--- a/src/BenchmarksApps/TodosApi/ProducesResponseTypeMetadata.cs
+++ b/src/BenchmarksApps/TodosApi/ProducesResponseTypeMetadata.cs
@@ -7,25 +7,6 @@ internal sealed class ProducesResponseTypeMetadata : IProducesResponseTypeMetada
 {
     private readonly IEnumerable<string> _contentTypes;
 
-    public ProducesResponseTypeMetadata(int statusCode)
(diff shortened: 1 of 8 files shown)


src/BenchmarksApps/TodosApi/ValidationFilter.cs
```diff
@@ -0,0 +1,158 @@
...
+            }
+        }
+
+        return result;
+    }
+}
+
+internal class ValidationMetadata(int[] map)
+{
+#pragma warning disable CA1822 // Mark members as static (This seems to be a bug with the analyzer not supporting primary constructors)
```
eerhardt, line 109: Did you log a bug for this?

DamianEdwards: Nope but will do.

src/BenchmarksApps/TodosApi/TodosApi.csproj
```diff
@@ -6,8 +6,9 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <ServerGarbageCollection>false</ServerGarbageCollection>
     <InvariantGlobalization>true</InvariantGlobalization>
+    <LangVersion>preview</LangVersion>
     <UserSecretsId>b8ffb8d3-b768-460b-ac1f-ef267c954c85</UserSecretsId>
-    <PublishAot>false</PublishAot>
+    <PublishAot>true</PublishAot>
```
eerhardt, line 11: The crank benchmarks don't assume this is set to `false` do they?

DamianEdwards: Ummm, good question. I know the force it on for the native aot run, but will need to check if they force it off for the normal runs.

src/BenchmarksApps/TodosApi/TodoApi.cs
```diff
@@ -11,13 +11,20 @@ public static RouteGroupBuilder MapTodoApi(this IEndpointRouteBuilder routes)
     {
         var group = routes.MapGroup("/api/todos");
 
-        group.MapGet("/", (NpgsqlDataSource db, CancellationToken ct) => db.QueryAsync<Todo>("SELECT * FROM Todos", ct))
+        group.AddValidationFilter();
+
+        // BUG: Having to call ToListAsync() on query results until JSON support for unspeakable types is resolved
```
eerhardt, line 16:
```suggestion
        // BUG: Having to call ToListAsync() on query results until JSON support for unspeakable types (https://github.com/dotnet/aspnetcore/issues/47548) is resolved
```

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: All good, it passes `/p:PublishAot=false` for the vanilla scenario

DamianEdwards: https://github.com/aspnet/Benchmarks/blob/1e59fc99c0f9a98e37c2108b7a7d0077f2d90a2a/scenarios/goldilocks.benchmarks.yml#L141-L148

src/BenchmarksApps/TodosApi/ProducesResponseTypeMetadata.cs
```diff
@@ -0,0 +1,89 @@
...
+
+    /// <summary>
+    /// Gets or sets the HTTP status code of the response.
+    /// </summary>
+    public int StatusCode { get; set; }
+
+    public IEnumerable<string> ContentTypes => _contentTypes;
+
+    //internal static ProducesResponseTypeMetadata CreateUnvalidated(Type? type, int statusCode, IEnumerable<string> contentTypes) => new(type, statusCode, contentTypes);
+
```
eerhardt, line 67:
```suggestion
```

Is this needed?

src/BenchmarksApps/TodosApi/ProducesResponseTypeMetadata.cs
```diff
@@ -0,0 +1,89 @@
...
+        MediaTypeHeaderValue.Parse(contentType);
+        for (var i = 0; i < additionalContentTypes.Length; i++)
+        {
+            MediaTypeHeaderValue.Parse(additionalContentTypes[i]);
+        }
+
+        _contentTypes = GetContentTypes(contentType, additionalContentTypes);
+    }
+
+    // Only for internal use where validation is unnecessary.
```
eerhardt, line 45: This is a weird comment. What does "Only for internal" mean? This is `private`.

src/BenchmarksApps/TodosApi/ValidationFilter.cs
```diff
@@ -0,0 +1,166 @@
...
+            }
+        }
+
+        return result;
+    }
+}
+
+internal class ValidationMetadata(int[] map)
+{
+#pragma warning disable CA1822 // Mark members as static: BUG https://github.com/dotnet/roslyn/issues/67783
```
eerhardt, line 116: (nit)

```suggestion
#pragma warning disable CA1822 // Mark members as static: BUG https://github.com/dotnet/roslyn-analyzers/issues/6573
```

src/BenchmarksApps/TodosApi/ValidationFilter.cs
```diff
@@ -0,0 +1,166 @@
...
+                    var problem = TypedResults.ValidationProblem(errors);
+                    return ValueTask.FromResult((object?)problem);
+                }
+            }
+        }
+        return next(context);
+    }
+
+    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code",
+        Justification = "Instance type is statically represented in generic argument declared as dynamically accessing public properties.")]
```
eerhardt, line 84: We should note here (because I'm sure people will look at this code and assume this suppression is always valid) that only the properties on `TTarget` are being preserved by the trimmer. Once you recurse and inspect properties of those property types, you are broken again.

src/BenchmarksApps/TodosApi/ValidationFilter.cs
```diff
@@ -0,0 +1,166 @@
...
+                    var errors = MapToFinalErrorsResult(workingErrors);
+                    var problem = TypedResults.ValidationProblem(errors);
+                    return ValueTask.FromResult((object?)problem);
+                }
+            }
+        }
+        return next(context);
+    }
+
+    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code",
```
eerhardt, line 83:
```suggestion
    [UnconditionalSuppressMessage("Trimming", "IL2026:RequiresUnreferencedCode",
```

This is shorter, and conveys the same amount of information.

src/BenchmarksApps/TodosApi/ProducesResponseTypeMetadata.cs
DamianEdwards: This type was copied from dotnet/aspnetcore as it's internal there.

src/BenchmarksApps/TodosApi/ValidationFilter.cs
DamianEdwards: There's no recursion going on here though, and System.CM.DA doesn't recurse today. This isn't library code, it's an internal type in an app, so I'm not quite sure I fully understand your threshold for how suppressions should be used yet.

src/BenchmarksApps/TodosApi/ProducesResponseTypeMetadata.cs
DamianEdwards: Removed it

src/BenchmarksApps/TodosApi/ValidationFilter.cs
eerhardt: It's more about the wording of the `Justification`.  I think suppressing for this app is fine.

eerhardt:
```suggestion
        Justification = "Instance type is statically represented in generic argument declared as dynamically accessing public properties. This ensures the properties on TTarget are preserved. Note that recursive properties of those properties are still not preserved, but this validation doesn't recurse.")]
```

DamianEdwards: I updated the justification to be more descriptive. LMK if you think it needs more.

PR #1835, opened 2023-04-19
Fix NativeAOT Stage2 on Windows

eerhardt:
Stage2 on Windows is currently failing `Could not find an agent named 'db'.` This is because we aren't adding a db profile for Windows. Adding the db profile used by Linux.

### DIFF
--- a/build/benchmarks-ci-01.yml
+++ b/build/benchmarks-ci-01.yml
@@ -269,7 +269,7 @@ jobs:
   - template: nativeaot-scenarios.yml
     parameters:
       connection: ASP.NET Benchmarks Queue 3
-      arguments: "$(ciProfile) --profile intel-win-app --profile intel-lin-load "
+      arguments: "$(ciProfile) --profile intel-win-app --profile intel-lin-load --profile amd-lin2-db "
       
 # GROUP 8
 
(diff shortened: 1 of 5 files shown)


build/benchmarks-ci-01.yml
```diff
@@ -269,7 +269,7 @@ jobs:
   - template: nativeaot-scenarios.yml
     parameters:
       connection: ASP.NET Benchmarks Queue 3
-      arguments: "$(ciProfile) --profile intel-win-app --profile intel-lin-load "
+      arguments: "$(ciProfile) --profile intel-win-app --profile intel-lin-load --profile amd-lin2-db "
```
eerhardt, line 272: @sebastienros - one inconsistency I noticed was that 

* `Intel Windows` leg uses: `--profile intel-lin-load`
* `Intel Linux` leg uses: `--profile amd-lin2-load`

What are the implications of using a different load profile? For the `Fortunes` benchmark, I see we are using the same load profile. Literally everything in the command is different besides 

image

DamianEdwards: Not sure, I'll leave to @sebastienros to reply to this PR.

sebastienros (approved):

build/benchmarks-ci-01.yml
eerhardt: @sebastienros explained it to me offline. this is fine because the load isn't the bottleneck on the app.

PR #1836, opened 2023-04-21
TodosApi updates

DamianEdwards:
- Adds configuration binding in preparation for configuration binding source generator
- Configured project so that OpenAPI document can be produced via a build target but OpenAPI code is removed at publish time via a runtime feature switch (this could potentially be integrated into `Microsoft.AspNetCore.OpenApi`)
  - e.g. in PowerShell: `$env:ASPNETCORE_ENVIRONMENT='Build' ; dotnet build -p:OpenApiGenerateDocumentsOnBuild=true ; $env:ASPNETCORE_ENVIRONMENT=$null`
- Added launchSettings.json files to all the projects

Seems to have no performance impact:

| load                   | todosapi_aot | todosapi_updates |         |
| ---------------------- | ------------ | ---------------- | ------- |
| CPU Usage (%)          |           12 |               13 |  +8.33% |
| Cores usage (%)        |          349 |              368 |  +5.44% |
| Working Set (MB)       |           48 |               48 |   0.00% |
| Private Memory (MB)    |          358 |              358 |   0.00% |
| Start Time (ms)        |            0 |                0 |         |
| First Request (ms)     |           75 |               89 | +18.67% |
| Requests/sec           |       72,701 |           74,347 |  +2.26% |
| Requests               |    1,096,144 |        1,121,051 |  +2.27% |
| Mean latency (ms)      |         4.22 |             4.09 |  -3.08% |
| Max latency (ms)       |        33.41 |            37.74 | +12.96% |
| Bad responses          |            0 |                0 |         |
| Socket errors          |            0 |                0 |         |
| Read throughput (MB/s) |        34.81 |            35.59 |  +2.24% |
| Latency 50th (ms)      |         1.79 |             1.83 |  +2.23% |
| Latency 75th (ms)      |         7.11 |             6.80 |  -4.36% |
| Latency 90th (ms)      |        10.29 |             9.85 |  -4.28% |
| Latency 99th (ms)      |        12.89 |            12.68 |  -1.63% |


src/BenchmarksApps/TodosApi/TodosApi.csproj
```diff
@@ -9,12 +9,27 @@
     <LangVersion>preview</LangVersion>
     <UserSecretsId>b8ffb8d3-b768-460b-ac1f-ef267c954c85</UserSecretsId>
     <PublishAot>true</PublishAot>
-    <EnableLogging Condition="$(Configuration.StartsWith('Debug'))">true</EnableLogging>
+    <EnableConfigurationBindingGenerator>false</EnableConfigurationBindingGenerator>
```
eerhardt, line 12: Why are we disabling this?

src/BenchmarksApps/TodosApi/TodosApi.csproj
```diff
@@ -9,12 +9,27 @@
...
-
+  
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="8.0.0-preview.*" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.0-preview.*" />
     <PackageReference Include="Npgsql" Version="8.0.0-preview.2" />
+    <Content Update="appSettings.Development.json" CopyToPublishDirectory="false" />
+  </ItemGroup>
+
+  <PropertyGroup Condition=" '$(EnableOpenApi)' == 'true' ">
```
eerhardt, line 26: I'm a bit torn on this. We explicitly cut Open API from our .NET 8 target. So I wonder if this is just clutter.

src/BenchmarksApps/TodosApi/Program.cs
```diff
@@ -22,20 +27,21 @@
...
-        If running locally, set the connection string in user secrets for key 'ConnectionStrings:TodoDb'.
-        If running after deployment, set the connection string via the environment variable 'CONNECTIONSTRINGS__TODODB'.
-        """);
-builder.Services.AddSingleton(_ => new NpgsqlSlimDataSourceBuilder(connectionString).Build());
+builder.Services.AddSingleton(sp =>
+{
+    var appSettings = sp.GetRequiredService<IOptions<AppSettings>>().Value;
+    return new NpgsqlSlimDataSourceBuilder(appSettings.ConnectionString).Build();
+});
+builder.Services.AddHostedService<DatabaseInitializer>();
```
eerhardt, line 44: Random thought:

I kind of liked how it was done before - called inline at the beginning of the app. Disadvantages of making this a "hosted service" that I see are:

1. It is only needed at startup. Holding onto the object for the lifetime of the app just to call a "no-op" Stop method seems wasteful.
2. There are ordering concerns such that this "initialization" might not have happened by the time some other piece of the app needs it. Doing it inline makes it explicit when during startup this happens.

src/BenchmarksApps/BasicMinimalApi/Properties/launchSettings.json
```diff
@@ -0,0 +1,12 @@
+{
+  "profiles": {
```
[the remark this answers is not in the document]
DamianEdwards: Hmm I don't know where that came from actually 😕

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: 'Cos it's currently broken (it literally generates code with a compilation bug)

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: Yeah I'm exploring if there's a flow here where OpenAPI is enabled just so that we can do build-time document generation, but not in the app when published (and thus not impacted by native AOT). It very likely needs some more iteration and arguably could be done in a different place other than this app, but this is the one that's tracked right now.

src/BenchmarksApps/TodosApi/Program.cs
DamianEdwards: I moved it here in an attempt to have it not run when the build action that generates the OpenAPI document runs. It turns out the way that works is more intrusive than I imagined and it actually fully starts the application (during build!) but with a fake `IServer` and `IHostLifetime` so moving it to a hosted service wasn't enough on its own. I was trying to find the right abstraction to hook into to ensure the DB initialization happens **before** the server starts, but **not** if the app is started by an external tool doing host inspection (e.g. EF migrations, OpenAPI generation, etc.) but this might not be the right/ideal solution. It might actually not be possible given the way all the pieces interact right now.

src/BenchmarksApps/BasicMinimalApi/Properties/launchSettings.json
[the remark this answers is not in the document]
eerhardt: I think it comes from the template. If I create a new web api project in VS, I get:

image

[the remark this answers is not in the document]
DamianEdwards: Ah OK then. I guess the one generated by VS when you have none doesn't include that property then.

src/BenchmarksApps/TodosApi/TodosApi.openapi.json
```diff
@@ -8,7 +8,7 @@
     "/throw": {
       "get": {
         "tags": [
-          "TodosApi"
+          "TodosApi, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null"
```
davidfowl: This seems not ideal

DamianEdwards: Oh interesting. I'm assuming you mean the document name seemingly being the full assembly name? This was produced by our own package (MS.Ext.ApiDescription.Server) but not sure why it changed. It's possible the previous version was produced by the Swashbuckle CLI and there's a subtle difference somewhere.

DamianEdwards: Wait, is this from an old commit? The latest doesn't have this, perhaps it was the Swashbuckle CLI that added it?

src/BenchmarksApps/TodosApi/DatabaseInitializer.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+
+    private async Task Initialize(CancellationToken cancellationToken = default)
+    {
+        // NOTE: Npgsql removes the password from the connection string
+        if (_logger.IsEnabled(LogLevel.Information))
+        {
+            _logger.LogInformation("Ensuring database exists and is up to date at connection string '{connectionString}'", _db.ConnectionString);
+        }
```
davidfowl, line 42: This doesn't seem like a good thing to log in general...

src/BenchmarksApps/TodosApi/TodosApi.openapi.json
DamianEdwards: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.Cli/HostingApplication.cs#L68-L72

src/BenchmarksApps/TodosApi/DatabaseInitializer.cs
DamianEdwards: This used to include code that sanitized the connection string but it turns out Npgsql does that already. But as you say it might be better to just not log it at all.

src/BenchmarksApps/TodosApi/HostEnvironmentExtensions.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Microsoft.Extensions.Hosting;
+
+internal static class HostEnvironmentExtensions
+{
+    public static bool IsBuild(this IHostEnvironment hostEnvironment) => hostEnvironment.IsEnvironment("Build");
```
eerhardt, line 5: Where does this `"Build"` environment get defined/set?

DamianEdwards: You have to set it manually via the `ASPNETCORE_ENVIRONMENT` env var. So e.g. in PowerShell to generate the swagger file you'd do this:

```PowerShell
> $env:ASPNETCORE_ENVIRONMENT='Build' ; dotnet build -p:OpenApiGenerateDocumentsOnBuild=true ; $env:ASPNETCORE_ENVIRONMENT=$null
```

src/BenchmarksApps/TodosApi/appsettings.json
```diff
@@ -5,7 +5,6 @@
       "Microsoft.AspNetCore": "Warning"
     }
   },
-  "AllowedHosts": "*",
```
eerhardt, line 8: I thought we wanted this in our appsettings file. https://github.com/dotnet/aspnetcore/issues/47862

src/BenchmarksApps/TodosApi/TodosApi.http
```diff
@@ -1,6 +1,6 @@
 @TodosApi_HostAddress = http://localhost:5054
 # Uncomment following line if sending requests to published app
-#@TodosApi_HostAddress = http://localhost:5000
+@TodosApi_HostAddress = http://localhost:5000
```
eerhardt, line 3:
```suggestion
#@TodosApi_HostAddress = http://localhost:5000
```

If we are going to leave the `# Uncomment following line if sending requests to published app` line in, we might as well leave this line commented out.

src/BenchmarksApps/TodosApi/TodosApi.csproj
```diff
@@ -9,12 +9,31 @@
...
+    <PackageReference Include="Microsoft.Extensions.ApiDescription.Server" Version="8.0.0-preview.4.*">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
   </ItemGroup>
+
+  <ItemGroup>
+    <Content Update="appSettings.Development.json" CopyToPublishDirectory="false" />
+    <!-- Workaround for https://github.com/dotnet/aspnetcore/issues/47941 -->
+    <IlcArg Include="--nopreinitstatics" />
```
eerhardt, line 35: Can we delete this now that https://github.com/dotnet/runtime/pull/85506 is fixed?

src/BenchmarksApps/TodosApi/DatabaseConfiguration.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public static IServiceCollection AddDatabase(this IServiceCollection services)
+    {
+        services.AddSingleton(static sp =>
+        {
+            var appSettings = sp.GetRequiredService<IOptions<AppSettings>>().Value;
+            var hostEnvironment = sp.GetRequiredService<IHostEnvironment>();
+            var db = hostEnvironment.IsBuild()
+            ? default!
+            : new NpgsqlSlimDataSourceBuilder(appSettings.ConnectionString).Build();
```
eerhardt, line 17: (nit) whitespacing to make more readable.

```suggestion
            var db = hostEnvironment.IsBuild()
                ? default!
                : new NpgsqlSlimDataSourceBuilder(appSettings.ConnectionString).Build();
```

eerhardt (approved):

src/BenchmarksApps/TodosApi/TodosApi.csproj
DamianEdwards: Has that fixed flowed to installers yet?

eerhardt: I don't believe so, but we don't need it to. The benchmarks use the latest daily build of runtime.

DamianEdwards: It has not (talking about `main` here). While this would run on the perf infrastructure 'cos it pulls the latest individual runtime installers, it would be very difficult to run this on dev machines until the fix flows to the `main` SDK installer, so I think we shouldn't be removing workarounds until that happens.

eerhardt: It's coming in https://github.com/dotnet/installer/pull/16275. Hasn't been merged yet.

PR #1837, opened 2023-04-24
Fix NativeAOT Linux runs

eerhardt:
Getting error "Could not find a profile named 'intel-db-app'."

Fix by adding an intel-db-app profile.

### DIFF
--- a/build/ci.profile.yml
+++ b/build/ci.profile.yml
@@ -62,6 +62,17 @@ profiles:
           - warmup
           - secondary
 
+  intel-db-app:
+    variables:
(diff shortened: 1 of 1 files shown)


DamianEdwards:
Was this never working, or did this break due to some other change?

eerhardt:
> Was this never working, or did this break due to some other change?

It was working up until https://github.com/aspnet/Benchmarks/commit/ee94e0a4e63afa80415aa37bd05f950c88b2bc80, which refactored which machines these jobs were run on.

PR #1840, opened 2023-04-26
Add OptimizationPreference=Speed NativeAOT runs

MichalStrehovsky:
This is now a documented option: https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/optimizing

I've put it on the same schedule as PGO runs since it's a similar thing but I don't know the motivation behind it. Should I tweak the modulus remainder to a different number than 1?

Cc @sebastienros @eerhardt @DamianEdwards @JamesNK

### DIFF
--- a/build/nativeaot-scenarios.yml
+++ b/build/nativeaot-scenarios.yml
@@ -102,6 +102,11 @@ parameters:
     arguments: --scenario basicgrpcpublishaot $(goldilocksJobs) --property scenario=Stage1GrpcAotServerGC --property publish=nativeaot --application.packageReferences \"Microsoft.Dotnet.ILCompiler=$(MicrosoftNETCoreAppPackageVersion)\" --application.buildArguments \"/p:ServerGarbageCollection=true\"
     condition: 'true'
 
+  - displayName: Goldilocks gRPC Stage 1 (NativeAOT - Server GC)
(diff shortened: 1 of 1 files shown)


sebastienros:
@MichalStrehovsky 

> I don't know the motivation behind it

This pipeline is executed 2 times per day, but we don't need to run some of these jobs that often in order to limit how long each run takes. This property will send the command to crank only if the condition is true, in this case by specifying how many half-days it should run, every 6th one in this case. And then some of them run on the first 6th, the second one. This way the native-aot job will usually take the same amount of time even though not the same jobs run on each iteration.

eerhardt:
> Should I tweak the modulus remainder to a different number than 1?

I think that would make sense. If we put the "OptimizationPreference=Speed" on a different day (but still every 6 half days).

eerhardt:
> And then some of them run on the first 6th, the second one

note that in the current code in main, there are 4 of these conditions.  3 of them are `== 1` and 1 of them is `== 2`. Maybe we should take this opportunity to split them up.

sebastienros:
@eerhardt sure, you don't have to for the ones that you really want to track as much as possible. But since we don't do changes more often than twice a day it doesn't really make sense to not do it.
Makes even more sense for the runs that take longer, like the native-aot ones where the build is slower.

eerhardt (approved):
Thanks! Let me know if you don't have permissions to merge, and I can.

MichalStrehovsky:
Yep, I don't have permissions to merge this. Thanks!

PR #1845, opened 2023-05-02
Fixed some issues with some GC scenarios

mrsharm:
There were a couple of issues with the configuration that are now fixed such as:

1. The right environment variable for the 2gb server case.
2. Unique name for the 2gb-pinning case.

CC: @sebastienros


sebastienros (approved):

sebastienros:
Feel free to merge. If you can't then you need to join aspnet-push or aspnet-team.

PR #1846, opened 2023-05-02
Depend on nanorm.npgsql

DamianEdwards:
Replace data extension methods with dependency on Nanorm.Npgsql


eerhardt:
Do we have precedent for depending on "non-product" libraries in our benchmarks? Or is this the first one?

sebastienros:
Dapper, we used to depend on NewtonSoft.Json. So yes, as long as these are tools from MS employees  ;)

DamianEdwards:
> as long as these are tools from MS employees ;)

Just to be clear, this is a joke 😄

sebastienros:
Oops sorry, forgot the smiley. Everything I say should have a smiley at the end ... which is a problem in itself.

eerhardt (approved):

PR #1847, opened 2023-05-22
Update Npgsql & RazorSlices versions

DamianEdwards:


sebastienros:
Have you tried to run a benchmark from this branch to verify that the versions are available in the internal feed?

DamianEdwards:
I mirrored the new versions myself. It doesn't even restore locally if they're not in the feed :)

src/BenchmarksApps/TechEmpower/BlazorUnited/BlazorUnited.csproj
```diff
@@ -10,7 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="Dapper" Version="2.0.123" />
-    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="8.0.0-preview.3" />
-    <PackageReference Include="Microsoft.EntityFrameworkCore.Relational" Version="8.0.0-preview.3.23174.2" />
+    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="8.0.0-preview.4" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Relational" Version="8.0.0-preview.4.*" />
```
roji, line 14: Maybe better to pin? Not sure if we have the daily build feed, but if we do at some point this starts getting messy real quick..

roji (approved):

PR #1848, opened 2023-05-23
Remove manual YARP PGO runs

MihaZupan:
No need for these anymore now that PGO is on by default.

### DIFF
--- a/build/proxies-scenarios.yml
+++ b/build/proxies-scenarios.yml
@@ -36,14 +36,6 @@ parameters:
     supportsServerHttp2: true
     supportsDownstreamHttp2: true
     condition: Math.round(Date.now() / 43200000) % 4 == 0 # once every 4 half-days
-  - displayName: YARP-PGO
-    scenario: proxy-yarp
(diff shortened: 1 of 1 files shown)


sebastienros (approved):
Should I also delete the old results from the db then?

sebastienros:
I found a few benchmarks being slower even though overall it's a positive change, haven't filed an issue yet. Adding @EgorBo for awareness.

sebastienros:
Maybe the difference is between dynamic PGO and full PGO.

MihaZupan:
Is ASP.NET also R2R'd?
For reference, with YARP scenarios here, the difference from this trade-off appear to be ~0, only Kestrel with HTTP/2 is slightly worse off.

EgorBo:
> Is ASP.NET also R2R'd?

Should be so, yes

MihaZupan:
Thanks, Egor.

> Should I also delete the old results from the db then?

I think we can, not much point in tracking that difference from YARP's perspective.

PR #1856, opened 2023-06-05
Syntax error

sebastienros:


src/BenchmarksApps/TechEmpower/PlatformBenchmarks/Program.cs
```diff
@@ -19,6 +19,8 @@ public class Program
 
         public static async Task Main(string[] args)
         {
+            foo
```
DamianEdwards, line 22: Yikes. What were we saying about a smoke test action 😬

DamianEdwards: I'm assuming this is a test

sebastienros:
This is testing the workflow, yes. But it's not working, maybe it's disabled, will check with Will. I might have to use azdo.

PR #1860, opened 2023-06-08
Switch NativeAOT scenarios to use GCDynamicAdaptationMode by default.

eerhardt:
Also adding "Workstation GC" runs every 3 days for comparisons.

Note that once we have an SDK with https://github.com/dotnet/sdk/pull/33043, we can move the default into the .csproj and not have to set the environment variable everywhere.

cc @mangod9

### DIFF
--- a/build/nativeaot-scenarios.yml
+++ b/build/nativeaot-scenarios.yml
@@ -15,96 +15,111 @@ parameters:
   default:
 
   - displayName: Goldilocks Stage 1 (CoreCLR)
-    arguments: --scenario basicminimalapivanilla $(goldilocksJobs) --property scenario=Stage1 --property publish=coreclr
+    arguments: --scenario basicminimalapivanilla $(goldilocksJobs) --property scenario=Stage1 --property publish=coreclr --application.environmentVariables DOTNET_GCDynamicAdaptationMode=1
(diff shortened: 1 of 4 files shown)


build/nativeaot-scenarios.yml
```diff
@@ -15,96 +15,111 @@ parameters:
   default:
 
   - displayName: Goldilocks Stage 1 (CoreCLR)
```
sebastienros, line 17: So `Stage1` now has a different GC mode? And before that both `Stage1ServerGC` and `Stage1` where server GC?

sebastienros: Just saw the changes in the csproj, it was actually Workstation implicitly. So fine.

sebastienros (approved):

build/nativeaot-scenarios.yml
eerhardt: right - we are testing the 3 GC modes:

* ServerGC + DynamicAdapationMode (what we want to be the "default")
* ServerGC
* WorkstationGC (only run every 3 days)

It is a bit of a mess right now that we need to set the env var everywhere. But we can refactor that when the MSBuild property flows.

PR #1854, opened 2023-05-31
Add TE Performance Tests Using Composite Images

ivdiazsa:
This PR is for issue #1849. It adds a new Middleware Dockerfile, which targets the newly released composite images, and enables some scenarios to run in TE using them in containers.

PR #1861, opened 2023-06-12
Add an Alpine Flavor for TE Performance Tests

ivdiazsa:
This PR follows up on PR #1854. Now that the composite images are up and running, it is ideal to be able to compare them to non-composites on equal grounds. So, this PR enables tests using the non-composite Alpine image so that we can compare and analyze against the composites, which are also built on Alpine.


ivdiazsa:
@sebastienros Hi Sebastien! Here's the PR to enable the normal Alpine tests.

PR #1862, opened 2023-06-12
Add startup event

eerhardt:
- Add a new EventSource event that fires when the benchmark application has started.

### DIFF
--- a/src/BenchmarksApps/BasicMinimalApi/BasicMinimalApi.csproj
+++ b/src/BenchmarksApps/BasicMinimalApi/BasicMinimalApi.csproj
@@ -8,4 +8,8 @@
     <InvariantGlobalization>true</InvariantGlobalization>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Include="..\Common\StartupExtensions.cs" />
+  </ItemGroup>
+  
 </Project>
--- a/src/BenchmarksApps/BasicMinimalApi/Program.cs
+++ b/src/BenchmarksApps/BasicMinimalApi/Program.cs
@@ -20,7 +20,7 @@ todosApi.MapGet("/{id}", (int id) =>
         ? Results.Ok(todo)
         : Results.NotFound());
 
-app.Lifetime.ApplicationStarted.Register(() => Console.WriteLine("Application started. Press Ctrl+C to shut down."));
+app.RegisterStartup();
 app.Run();
 
 [JsonSerializable(typeof(Todo[]))]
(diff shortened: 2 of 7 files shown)


sebastienros:
Is the goal to allow for the crank server to listen to this event?

I was thinking of adding a custom metadata from the crank server directly since it already has the logic to detect when each job is started, ready and then stopped.

Then the clients would also add another event for intermediate state, like during warmup/load transition.

eerhardt:
@Maoni0 was asking for an event to determine when the app has started. I assume it is to correlate with other EventSource events.

@Maoni0 - can you describe your goals?

Maoni0:
since you care about the "Start time" in the output of running these benchmarks, we'd like a way to investigate regressions there so we'd like an event that indicates "this is the end of what you considered as startup".

sebastienros:
Makes totally sense and is matching my suggestions for which event we should record.

A flaw however with my suggestion is that you want to see that in the app's trace, os if the event was recorded in the results, of if it was emitted by the crank server then you wouldn't see it.

@eerhardt do you know if it's possible to write an event source on an external process? I already have an event pipe with it, and read events from it, can it also write to it, and then perfiew/dotnet trace would be able to record it too?

eerhardt:
> do you know if it's possible to write an event source on an external process?

I do not know if that is possible, but I would be surprised if it was. My understanding is that event sources come from within the app. Maybe @noahfalk would know?

Maoni0:
I was thinking of using perfview to collect it. @brianrob told me I can just add /Providers:*BenchmarksEventSource to the perfview cmdline. although we'll need to figure out what to do about linux there. 

right, it would be to correlate with runtime/OS events.

mrsharm:
Just tried this PR and concluded that this will most definitely help us when we go about diagnosing perf issues. As a concrete example, for [this](https://github.com/dotnet/runtime/issues/94076) issue we'd greatly benefit from knowing specifically when the StartUp event was fired to understand what's causing the higher start up time for DATAS. 

Our process of diagnosing these types of regressions would be simplified as: 
1. After enabling the BenchmarksEventSource, we are able to get the Relative Timestamp of when the App started directly from the trace for the regressed case. 
image
2. We'd simply look up the frames pertinent to the GC from the CPU Stacks constrained to the specific relative start time. 
3. Make conclusions based on repeating 1. and 2. for the baseline and noting the differences.

What would be the next steps here? Can we reopen this PR? 

CC: @Maoni0

sebastienros:
I thought about it some time ago and I think this would make sense to be integrated directly into the crank agent. This way all apps would get the event for free since we already have a way to detect when the app is started. Can you look into it? We can get on a call if you need me to show you how.

mrsharm:
As a heads up - the new event that demarcates the start is now available in crank. The only prerequisite is that you need to enable the "Benchmark" provider i.e., set `Providers:Benchmarks` in the collect arguments.

image

PR #1866, opened 2023-06-15
Add Next.js implementation of the TechEmpower Fortunes benchmark

DamianEdwards:
This is an attempt to add an implementation of the TechEmpower Fortunes benchmark in Next.js. This is interesting to us given Blazor's updates in .NET 8 to support SSR, etc.

Some notes about this implementation:

- I started by following the [Next.js docs](https://nextjs.org/docs) as linked to from the "Getting Started" link on their home page
- The app is using the new "App Router" in Next.js 13, rather than the (legacy?) "Pages Router"
- The app is downgraded to use Next.js 13.4.0 rather than latest due to [this issue](https://github.com/vercel/next.js/issues/48173#issuecomment-1584585902)
- It's using the [node-postgres PostgreSQL node client](https://node-postgres.com/) with a shared [pool](https://node-postgres.com/apis/pool)
- The HTML table on the */fortunes* page doesn't exactly match the [example response in the requirements](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#fortunes) as not having a `<tbody>` element causes issues with Next.js partial rendering, i.e. an error is raised in the browser that the app could not be "hydrated" due to an expected `<tr>` not being present.
- The response HTML for */fortunes* includes a number of extra elements not specified in the templates, including a [`<meta>` viewport element in the `<head>`](https://github.com/vercel/next.js/discussions/17277), and all the `<script>` elements at the end of the page required to enable Next.js client-side rendering. I could not find a way to disable either of these.
- The app is using a custom server in order to enable node clustering (scaling cross multiple CPUs) in production. This was adapted from the [official custom server example here](https://github.com/vercel/next.js/tree/canary/examples/custom-server)
- When running the app locally on my development desktop (Intel Core i9-12900K), after building with `npm run build` and running with `npm run start`, with Postgres in a Docker container, and load generated by Bombardier, I see ~1,120 requests per second:
   ``` shell
    $ C:\tools\bombardier-windows-amd64.exe http://localhost:3000/fortunes -d 5s -c 100
    Bombarding http://localhost:3000/fortunes for 5s using 100 connection(s)
    [================================================] 5s
    Done!
    Statistics        Avg      Stdev        Max
    Reqs/sec      1114.53     371.69    2442.34
    Latency       88.92ms    13.40ms   158.84ms
    HTTP codes:
      1xx - 0, 2xx - 5683, 3xx - 0, 4xx - 0, 5xx - 0
      others - 0
    Throughput:     9.52MB/s
   ```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ bld/
 [Bb]in/
 [Oo]bj/
 [Ll]og/
+dist/
 
 # Visual Studio 2015 cache/options directory
 .vs/
--- a/src/BenchmarksApps/TechEmpower/nextjs/.eslintrc.json
+++ b/src/BenchmarksApps/TechEmpower/nextjs/.eslintrc.json
@@ -1,0 +1,3 @@
+{
+  "extends": "next/core-web-vitals"
+}
--- a/src/BenchmarksApps/TechEmpower/nextjs/.gitignore
+++ b/src/BenchmarksApps/TechEmpower/nextjs/.gitignore
@@ -1,0 +1,35 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
+/.pnp
+.pnp.js
+
+# testing
+/coverage
+
+# next.js
+/.next/
+/out/
+
+# production
+/build
+
+# misc
+.DS_Store
+*.pem
+
+# debug
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# local env files
+.env*.local
+
+# vercel
+.vercel
+
+# typescript
+*.tsbuildinfo
+next-env.d.ts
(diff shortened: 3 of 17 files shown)


DamianEdwards:
Updated as part of investigating the performance:

- Added ability to run with no database query via environment variable `NO_DB` to a JS falsey value
- Added ability to set number of workers via environment variable `WORKER_COUNT`
- Added ability to set Postgres pool max client count via environment variable `DB_MAX_CLIENTS` & changed its default to `os.cpus().length`

Ran numbers again with no database and reduced worker count and performance is still under 1,000 RPS.

DamianEdwards:
New update: I noticed that Next.js is always sending `Connection: close` response headers. This will dramatically impact request latency. I'm trying to figure out what's going on but none of the obvious things seem to impact it yet, e.g. setting environment variable `KEEP_ALIVE_TIMEOUT` which is read by the *server.js* file generated by the standalone build. This only seems to be happening in non-dev so it appears to be something related to the custom server.

PR #1871, opened 2023-06-21
Update package refs for preview.5

DamianEdwards:
Update package refs in benchmarks apps for preview.5


sebastienros:
Shouldn't these always match AspNetCore or NetCoreApp versions, if set by crank?

DamianEdwards:
They've always been this way for these newer projects. How should they be to get the behavior you're describing?

sebastienros:
`$(MicrosoftAspNetCoreAppPackageVersion)` is set by crank. This will be mapped to the aspnet version that is picked up.

PR #1876, opened 2023-07-13
Switch over to using the built-in ProducesResponseTypeMetadata.

mitchdenny:
Benchmarks have been broken since the 7th because the stage 2 template defines its own `ProducesResponseTypeMetadata` type which conflicts with the new built-in one. This change switches over to using the built-in one.


DamianEdwards:
Do you know what the minimum SDK version is that's required to get the new built-in type?

eerhardt (approved):
Thanks!

mitchdenny:
@DamianEdwards you'll need preview 7. I tested with the latest, but one from late last week _might_ be recent enough to have it.

PR #1878, opened 2023-07-19
Fix compiler warning

eerhardt:
BenchmarkApplication is sealed, so declaring a new protected member doesn't make sense.


eerhardt:
@sebastienros - can I get a quick review here?

sebastienros (approved):

PR #1879, opened 2023-07-20
Fix package name casing

eerhardt:
The correct casing is DotNet. Although NuGet package name are case-insenstive, https://github.com/NuGet/Home/issues/12757 broke using the wrong case.


eerhardt:
Merging to unblock the AOT perf legs.

PR #1881, opened 2023-08-02
Add more AOT benchmarks to regression bot.

eerhardt:
This should cover:
- Stage1Aot
- Stage1GrpcAot
- Stage2Aot

### DIFF
--- a/build/regressions.config.yml
+++ b/build/regressions.config.yml
@@ -90,7 +90,7 @@ sources:
   - name: PublishedSizeAot
     table: NativeAotBenchmarks
     rules:
-      - include: Stage*Aot
+      - include: Stage.*Aot
     daysToLoad: 60
     stdevCount: 5
     regressions:
@@ -132,7 +132,7 @@ sources:
   - name: Native AOT Reliability
     table: NativeAotBenchmarks
     rules:
-      - include: Stage*Aot
+      - include: Stage.*Aot
     regressions:
       labels: 
         - Perf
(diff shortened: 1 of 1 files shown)


build/regressions.config.yml
```diff
@@ -90,7 +90,7 @@ sources:
   - name: PublishedSizeAot
     table: NativeAotBenchmarks
     rules:
-      - include: Stage1Aot
+      - include: Stage*Aot
```
sebastienros, line 93: What does `*` mean? ;)

eerhardt: Same thing it means here:

https://github.com/aspnet/Benchmarks/blob/fc052c63779c4faa17cd5bbc13c79509a6b6ad3d/build/regressions.config.yml#L53

sebastienros: Exactly, so I don't think that's what you want. This field is a regular expression. Other example: `ulib|fasthttp|actix|gin|express|node|netty`

eerhardt: Does this not work, like a normal person would expect?

eerhardt: oh, so it isn't a wildcard. but a regex. I need a dot.

sebastienros: `Stage.Aot`, `Stage(1|2)Aot`, `Stage1Aot|Stage2Aot`, `Stage\dAot`, `Stage.*Aot` ...

eerhardt: Better?

sebastienros (approved):

build/regressions.config.yml
sebastienros: my suggestions won't work with grpc obviously, bad me

PR #1889, opened 2023-08-15
Add scenario for POSTing to minimal form

captainsafia:


src/BenchmarksApps/Mvc/benchmarks.mvcjson.yml
```diff
@@ -182,3 +182,17 @@ scenarios:
...
+
+  # MapAction EchoForm
+  MapActionEchoTodoForm:
+    application:
+      job: mapaction
+    load:
+      job: bombardier
+      variables:
+        verb: POST
+        headers:
```
sebastienros, line 194: There should be a `customHeaders` variable that is free-form, but the way you did this works too. It's just redifining the set of reusable headers and point to this new key.

sebastienros (approved):

sebastienros:
I'll let you click the merge button as an exercise.

src/BenchmarksApps/Mvc/benchmarks.mvcjson.yml
captainsafia: Good point -- I ended up simplifying this further and removing the dependency on the external body file.

PR #1892, opened 2023-08-15
Remove old TFMs

sebastienros:
Starts at net6.0
Built locally with 6.0 and 7.0


DamianEdwards:
Nice. Is this the only places we're using TFMs <`.net6.0`?

sebastienros:
@DamianEdwards no reference of net5 anywhere else. Just removed the nightly feeds too.

PR #1893, opened 2023-08-18
Update POST /EchoTodoForm endpoint

captainsafia:
Support .NET 7 and .NET 8 runtimes and remove dependency on complex form mapping.

### DIFF
--- a/src/BenchmarksApps/MapAction/Program.cs
+++ b/src/BenchmarksApps/MapAction/Program.cs
@@ -20,8 +20,12 @@ using var host = Host.CreateDefaultBuilder(args)
                 Todo EchoTodo([FromBody] Todo todo) => todo;
                 endpoints.MapPost("/EchoTodo", (Func<Todo, Todo>)EchoTodo);
                 // Intentionally disable anti-forgery for this endpoint so we can evaluate just the form-processing portions
-                endpoints.MapPost("/EchoTodoForm", ([FromForm] Todo todo) => todo)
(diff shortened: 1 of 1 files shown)


src/BenchmarksApps/MapAction/Program.cs
```diff
@@ -20,8 +20,12 @@
                 Todo EchoTodo([FromBody] Todo todo) => todo;
                 endpoints.MapPost("/EchoTodo", (Func<Todo, Todo>)EchoTodo);
                 // Intentionally disable anti-forgery for this endpoint so we can evaluate just the form-processing portions
-                endpoints.MapPost("/EchoTodoForm", ([FromForm] Todo todo) => todo)
```
BrennanConroy, line 23: Are we going to add a form binding benchmark later?

BrennanConroy (approved):

src/BenchmarksApps/MapAction/Program.cs
captainsafia: Yeah, as a different benchmark. I think when all is said and done we'll have the following scenarios:

- POST to IFormCollection
- POST to [FromForm] with complex type
- POST to IFormCollection with anti-forgery enabled

The should allow us to measure each component of form-processing somewhat independently:

- Reading the form body into an IFormCollection
- Running the anti-forgery checks
- Binding from the form body to a complex type

sebastienros:
So the bug is now fixed in aspnetcore?

PR #1924, opened 2023-10-30
Fix IIS benchmarks

BrennanConroy:


build/dependencies.props
```diff
@@ -24,5 +24,9 @@
     <MicrosoftEntityFrameworkCoreSqliteVersion70>7.0.10</MicrosoftEntityFrameworkCoreSqliteVersion70>
     <MicrosoftEntityFrameworkCoreSqliteVersion80>8.0.0-rc.2.23480.1</MicrosoftEntityFrameworkCoreSqliteVersion80>
 
+    <MicrosoftIISTestingVersion60>6.0.0-*</MicrosoftIISTestingVersion60>
+    <MicrosoftIISTestingVersion70>7.0.0-*</MicrosoftIISTestingVersion70>
+    <MicrosoftIISTestingVersion80>8.0.0-preview.5.23252.25</MicrosoftIISTestingVersion80>
```
BrennanConroy, line 29: This has to be pinned to a version before we stopped packaging Microsoft.AspNetCore.Testing. 9.0 will be fixed by https://github.com/dotnet/aspnetcore/pull/51713 so we can update to 9.0.0-* when that is merged.

src/Benchmarks/Program.cs
```diff
@@ -125,10 +125,18 @@ public static void Main(string[] args)
             }
             else if (String.Equals(Server, "IISInProcess", StringComparison.OrdinalIgnoreCase))
             {
+                // if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable("ASPNETCORE_IIS_VERSION")))
```
BrennanConroy, line 128: This relies on an 8.0 feature which isn't available in the old pinned version we currently have (see previous comment). We can uncomment once we've unpinned the version.

sebastienros (approved):

PR #1930, opened 2023-11-15
update Dapper and enable Dapper.AOT

mgravell:
With .NET 8 shipped, Dapper.AOT can also ship, yay!

This change:

- updates the `Dapper` package (although I believe it won't actually be used any more, although I haven't checked the outputs)
- adds the `Dapper.AOT` package which contains generator support and a small runtime library akin to Dapper (no ref-emit)
- adds `[DapperAot]` usage to enable the generator (this step is opt-in)
  - ~~explicitly turns AOT in one broken scenario (will fix external lib ASAP)~~
- adds `[CacheCommand]` usage to allow `DbCommand` reuse
- enables interceptor support in the build SDK
- suppresses a warning about a particular usage (alternative options cited in comments, but left "as is" for comparison purposes)

### DIFF
--- a/src/BenchmarksApps/TechEmpower/BlazorSSR/BlazorSSR.csproj
+++ b/src/BenchmarksApps/TechEmpower/BlazorSSR/BlazorSSR.csproj
@@ -10,7 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="Dapper" Version="2.1.21" />
-    <PackageReference Include="Dapper.AOT" Version="1.0.2" />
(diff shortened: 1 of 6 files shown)


DamianEdwards:
Wondering if there's any value in keeping both Dapper and Dapper.AOT and testing them both? Or do you think we should just move to Dapper.AOT now?

mgravell:
@DamianEdwards interesting question; I'm happy with either; I need to remember how to kick off test runs against branches - the data might help steer us there, although if it is useful I'm fine with keeping both - it just might complicate reporting

PR #1936, opened 2024-01-10
Benchmarks for SslStream

rzikm:
This PR adds benchmarks for SslStreams. Supported scenarios are:

- read-write - measures throughput of a single SslStream connection
- handshakes - measures the number of handshakes per second.


src/BenchmarksApps/SslStream/Client/Program.cs
```diff
@@ -0,0 +1,236 @@
...
+            var sendBuffer = new byte[bufferSize];
+            try
+            {
+                while (s_isRunning)
+                {
+                    await stream.WriteAsync(sendBuffer, cancellationToken);
+                    Interlocked.Add(ref s_bytesWritten, bufferSize);
+                }
+            }
+            catch (OperationCanceledException)
```
CarnaViire, line 112: I don't think it's a good idea to use OCE to cancel the last operation 🤔 You are measuring op/s, so I believe it should count the last one as well. Otherwise by not waiting for the last op to finish, you are superficially lowering the result -- because you are adding (part of the) time, but not the bytes. I think it doesn't matter that much that the benchmark itself will run a little bit longer that the asked e.g. 15s, it's better to get more accurate results.

src/BenchmarksApps/SslStream/Server/Program.cs
```diff
@@ -0,0 +1,177 @@
...
+        using var sock = new Socket(SocketType.Stream, ProtocolType.Tcp);
+        sock.Bind(new IPEndPoint(IPAddress.IPv6Any, options.Port));
+        sock.Listen();
+
+        Log($"Listening on {sock.LocalEndPoint}.");
+
+        CancellationTokenSource ctrlC = new CancellationTokenSource();
+        Console.CancelKeyPress += (s, e) =>
+        {
+            Log("Ctrl+C pressed.");
```
CarnaViire, line 34: 👀

scenarios/sslstream.benchmarks.yml
```diff
@@ -0,0 +1,99 @@
...
+jobs:
+  sslServer:
+    source:
+      # repository: https://github.com/aspnet/benchmarks.git
+      # branchOrCommit: main
+      localFolder: C:/source/aspnet/benchmarks/src/BenchmarksApps/SslStream
+      # project: src/BenchmarksApps/SslStream/Server/SslStreamServer.csproj
+      project: Server/SslStreamServer.csproj
+    isConsoleApp: true
+    # readyStateText: Application started # app should write this line to output -- default Kestrel logging does that
```
CarnaViire, line 86:
```suggestion
    readyStateText: Listening on # app should write this line to output
```
We should wait until the server starts listening, before starting the client. `readyStateText` is [a signal for that](https://github.com/dotnet/crank/blob/main/src/Microsoft.Crank.Controller/benchmarks.schema.json#L102). `isConsoleApp` should be `false` (default value), because "console app", as opposed to a "server", is [considered "ready" right away](https://github.com/dotnet/crank/blob/main/src/Microsoft.Crank.Controller/benchmarks.schema.json#L154), without waiting for a signal.

scenarios/sslstream.benchmarks.yml
```diff
@@ -0,0 +1,99 @@
...
+
+imports:
+  - https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/aspnet.profiles.standard.yml
+
+variables:
+  # common parameters
+  serverPort: 5000
+  sendBufferSize: 32768 # bytes
+  receiveBufferSize: 32768 # bytes
+  sslProtocols: "None" # <Default|None|Ssl2|Ssl3|Tls|Tls11|Tls12|Tls13>
```
CarnaViire, line 12: What happens if it's "None"? Does it refuse to connect, or is it not encrypted? Is this multi-value?

src/BenchmarksApps/SslStream/Client/ClientOptions.cs
```diff
@@ -0,0 +1,78 @@
...
+    public string Hostname { get; set; } = null!;
+    public CertificateSource CertificateSource { get; set; }
+    public X509Certificate2? ClientCertificate { get; set; } = null!;
+    public string? TlsHostName { get; set; }
+    public Scenario Scenario { get; set; }
+    public TimeSpan Warmup { get; set; }
+    public TimeSpan Duration { get; set; }
+}
+
+public class OptionsBinder : BinderBase<ClientOptions>
```
CarnaViire, line 30: I think I had to use the explicit binder in HttpClient benchmarks, because I had to manually parse the headers parameter from `string[]` into the list of (name, value) pairs... I think it should be possible to use an automatic binder in your case.

src/BenchmarksApps/SslStream/Client/Program.cs
```diff
@@ -0,0 +1,236 @@
...
+        Volatile.Write(ref s_bytesWritten, 0);
+
+        await Task.Delay(options.Duration);
+        s_isRunning = false;
+        Log("Completing scenario...");
+
+        cts.Cancel();
+
+        await writeTask;
+        await readTask;
```
CarnaViire, line 167:
```suggestion
        Task.WaitAll(writeTask, readTask);
```
Maybe?

scenarios/sslstream.benchmarks.yml
```diff
@@ -0,0 +1,99 @@
...
+  certSource: "Context" # <Callback|Certificate|Context>
```
CarnaViire, line 13: Is "certificate source" a term that is commonly used? 😅 because this option was confusing me, for quite some time I thought it was about where the _benchmark app_ should get the cert 😅

scenarios/sslstream.benchmarks.yml
```diff
@@ -0,0 +1,99 @@
...
+      job: sslClient
+      agent: main
+      variables:
+        scenario: Handshake
+        host: "{{secondaryAddress}}"
+
+jobs:
+  sslServer:
+    source:
+      # repository: https://github.com/aspnet/benchmarks.git
```
CarnaViire, line 80: Let's maybe have the code here that is in "ready for merge" state, just so we won't forget and accidentally merge the file with your local folders? 😄

scenarios/sslstream.benchmarks.yml
CarnaViire: (should default be "Default"? 😄)

scenarios/sslstream.benchmarks.yml
rzikm: We use something similar in S.N.S unit tests

https://github.com/dotnet/runtime/blob/77d643d8678d0cae1144557f106bb693a1cbca25/src/libraries/System.Net.Security/tests/FunctionalTests/SslStreamMutualAuthenticationTest.cs#L36-L41

I am open for suggestions.

src/BenchmarksApps/SslStream/Client/ClientOptions.cs
rzikm: as part of the binding, the code loads up the X509Certificate (path + optional password), so that the instance is already part of the resulting ClientOptions.

scenarios/sslstream.benchmarks.yml
rzikm: The naming is a bit unfortunate and has historical baggage, but Default is actually `Ssl3 | Tls` and is deprecated. I will see if I can remove it from the list.

https://github.com/dotnet/runtime/blob/77d643d8678d0cae1144557f106bb693a1cbca25/src/libraries/System.Net.Primitives/src/System/Net/SecureProtocols/SslEnumTypes.cs#L10-L26

rzikm: the None (= 0) value lets OS decide (i.e. highest supported TLS version)

CarnaViire: We don't have to use the actual SslProtocols enum here. It might be beneficial to limit the values to the ones we/endusers actually use. I.e. use a new enum that is later mapped to SslProtocols (esp given that you have an explicit binder anyway). And I'd suggest then not to use the confusing naming like "None" and "Default" 😅 So only have a small, like, 2-3 value enum here.

(btw on a side note, i don't think that letting OS decide stuff is a good thing from benchmarking perspective, it's not reproducible enough 😅 it's better to set everything explicitly 🤔)

src/BenchmarksApps/SslStream/Server/Program.cs
rzikm: Not sure what you are trying to say :D

scenarios/sslstream.benchmarks.yml
CarnaViire: I was thinking about it, and maybe something like "CertSelection"/"CertSelectionType"?

and this is only for the server now? you don't pass it to the client, but I see there's an option to do that in the future. then it might make sense to separate the variables by name, because the selection can be different between client and server.

```suggestion
  ServerCertSelection: "CertContext" # <Callback|Property|CertContext>
  ClientCertSelection: "CertContext" # <Callback|Collection|CertContext>
```

src/BenchmarksApps/SslStream/Server/Program.cs
```diff
@@ -0,0 +1,177 @@
...
+                Log($"Negotiated unknown protocol: {stream.NegotiatedApplicationProtocol}");
+            }
+        }
+        catch (AuthenticationException e)
+        {
+            Log($"Authentication failed: {e}");
+        }
+        catch (IOException e) when (e.InnerException is SocketException)
+        {
+            // client closed the connection, ignore this
```
CarnaViire, line 126: I think it will be better to count and report the number of errors (maybe even separately per kind). I believe, currently we've made the client to always finish "gracefully" in normal circumstances 🤔 am I missing something?

Also, even if we decide to ignore the last socket exception -- it should be the _last_ exception _only_ -- exceptions in the middle are wrong and should be reported to metrics.

_Just logging_ is not enough -- when the benchmark completes successfully (i.e. there were no unhandled exceptions), you normally don't see server's (and client's) logs, only the metrics. (esp. given that the end-goal is to add this as a power bi graph)

src/BenchmarksApps/SslStream/Server/Program.cs
```diff
@@ -0,0 +1,177 @@
...
+                // done everything
+                return;
+            }
+            if (stream.NegotiatedApplicationProtocol.Protocol.Span.SequenceEqual("ReadWrite"u8))
+            {
+                await ReadWriteScenario(stream, options, cancellationToken).ConfigureAwait(false);
+            }
+            else
+            {
+                Log($"Negotiated unknown protocol: {stream.NegotiatedApplicationProtocol}");
```
CarnaViire, line 117: Expanding the comment on the socket exception -- it made me think about unhandled exceptions -- I think we should treat some errors as "unrecoverable", meaning the benchmark itself must be set up incorrectly for us to get such an error. So we need to log, but should not prevent the exception from bubbling up. Then the benchmark will be aborted, and the error together with the logs will be reported in crank output.

E.g. if we suddenly negotiated an unknown protocol, it looks like something went wrong and we should not proceed further :)

src/BenchmarksApps/SslStream/Server/Program.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        var sslOptions = CreateSslServerAuthenticationOptions(options);
+        using var sock = new Socket(SocketType.Stream, ProtocolType.Tcp);
+        sock.Bind(new IPEndPoint(IPAddress.IPv6Any, options.Port));
+        sock.Listen();
+
+        Log($"Listening on {sock.LocalEndPoint}.");
+
+        CancellationTokenSource ctrlC = new CancellationTokenSource();
+        Console.CancelKeyPress += (s, e) =>
```
CarnaViire, line 32: Re: 👀:

It's just I didn't understand why we need this at all, so it looked like something leftover from debugging. Now I think I see the idea behind it. I'm still not fully convinced this needs to be this way (and I'm not sure crank uses the same event to kill the server?)

I'm not 100% sure, but I believe, the way the crank will work is to wait until the client exits, and only then kill the server... So, in my view, the only active operation at this point should be listening on a socket, and I think it shouldn't need a complicated termination, so just crank killing the process will do the trick? But I might be wrong, so LMK if I'm missing something

src/BenchmarksApps/SslStream/Server/Program.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        byte[] recvBuffer = new byte[options.ReceiveBufferSize];
+
+        CancellationTokenSource cts = new CancellationTokenSource();
+        using var reg = cancellationToken.Register(() => cts.Cancel());
+
+        var writeTask = Task.Run(() => WritingTask(stream, options.SendBufferSize, cts.Token));
+
+        var recvBufer = new byte[options.ReceiveBufferSize];
```
CarnaViire, line 86: should we start reading at all, if `options.ReceiveBufferSize` is 0? Maybe we need to do the same check as with the writing task, and then wait on both of the tasks?

src/BenchmarksApps/SslStream/Server/Program.cs
```diff
@@ -0,0 +1,177 @@
...
+    static async Task ProcessClient(Socket socket, ServerOptions options, SslServerAuthenticationOptions sslOptions, CancellationToken cancellationToken)
+    {
+        try
+        {
+            using var stream = await EstablishSslStreamAsync(socket, sslOptions, cancellationToken).ConfigureAwait(false);
+
+            if (stream.NegotiatedApplicationProtocol.Protocol.Span.SequenceEqual("Handshake"u8))
+            {
+                // done everything
+                return;
```
CarnaViire, line 109: Do I understand correctly that disposing the SslStream will also close the socket?

src/BenchmarksApps/SslStream/Client/Program.cs
```diff
@@ -0,0 +1,225 @@
...
+
+        var writeTask = Task.Run(() => WritingTask(stream, options.SendBufferSize));
+        var readTask = Task.Run(() => ReadingTask(stream, options.ReceiveBufferSize));
+
+        await Task.Delay(options.Warmup).ConfigureAwait(false);
+        Log("Completing warmup...");
+
+        Stopwatch sw = Stopwatch.StartNew();
+        Volatile.Write(ref s_counters.BytesRead, 0);
+        Volatile.Write(ref s_counters.BytesWritten, 0);
```
CarnaViire, line 150: The way the two of them are not set together atomically (and also, most possibly, during some operation that currently is running) worries me a bit 😅

in HttpClient benchmarks, it's explicitly done [between iterations](https://github.com/aspnet/Benchmarks/blob/444b7bf3b253f85ef44e1efb89824fbdc5426352/src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/Program.cs#L303-L314), so I wonder whether we might want to do something like that here?

CarnaViire:
cc @liveans -- FYI, we are discussing here some ideas and concerns regarding a benchmark implementation (e.g. running and finishing a scenario, exception handling, etc), some of the comments might be useful

src/BenchmarksApps/SslStream/Server/Program.cs
rzikm: Yes, the SslStream is set to dispose the inner NetworkStream which in turn is set to own the Socket. see `EstablishSslStreamAsync`

src/BenchmarksApps/SslStream/Server/Program.cs
rzikm: It probably is not necessary at this point anymore, it was useful during debugging. On the other hand, I don't think it hurts or complicates the code much.

src/BenchmarksApps/SslStream/Server/Program.cs
rzikm: We can't easilly bubble it up out of Main because we hand off incomming connections to a task on threadpool, but I can add a metric for errors.

src/BenchmarksApps/SslStream/Client/Program.cs
```diff
@@ -0,0 +1,249 @@
...
+#if NET8_0_OR_GREATER
+            AllowTlsResume = options.AllowTlsResume,
+#endif
+            EnabledSslProtocols = options.EnabledSslProtocols,
+            CertificateRevocationCheckMode = options.CertificateRevocationCheckMode,
+        };
+
+        if (options.ClientCertificate != null)
+        {
+            switch (options.CertificateSelection)
```
wfurt, line 203: I'm not sure if we need all this but I'm fine if we keep it. I would expect the differences would be minimal. I would expect the biggest difference would be certificate or not.

scenarios/sslstream.benchmarks.yml
```diff
@@ -0,0 +1,95 @@
...
+
+imports:
+  - https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/aspnet.profiles.standard.yml
+
+variables:
+  # common parameters
+  serverPort: 5000
+  sendBufferSize: 32768 # bytes
+  receiveBufferSize: 32768 # bytes
+  tlsVersion: "1.3" # <1.2|1.3>
```
wfurt, line 12: should ve leave it to None as default? It may make it less predictable but it would track default use nicely IMHO

src/BenchmarksApps/SslStream/Client/Program.cs
```diff
@@ -0,0 +1,249 @@
...
+                return;
+            }
+
+            BenchmarksEventSource.Register("sslstream/read/mean", Operations.Avg, Operations.Avg, "Mean bytes read per second.", "Bytes per second - mean", "n2");
+
+            var recvBuffer = new byte[bufferSize];
+            Stopwatch sw = Stopwatch.StartNew();
+            bool isWarmup = true;
+
+            while (s_isRunning)
```
wfurt, line 150: what is typical duration? We should make sure the read/write phase is much longer than handshake

wfurt: or would that happen in warmup phase?

src/BenchmarksApps/SslStream/Server/Program.cs
wfurt: I don't know what is precedent in the code base. I personally would be fine with leaving debug hooks but they should probably deserve comments or documentation .... so new persons can find and use them

src/BenchmarksApps/SslStream/Server/Program.cs
```diff
@@ -0,0 +1,204 @@
...
+        await Task.WhenAll(writeTask, readTask).ConfigureAwait(false);
+    }
+
+    static async Task ProcessClient(Socket socket, ServerOptions options, SslServerAuthenticationOptions sslOptions, CancellationToken cancellationToken)
+    {
+        try
+        {
+            using var stream = await EstablishSslStreamAsync(socket, sslOptions, cancellationToken).ConfigureAwait(false);
+
+            if (stream.NegotiatedApplicationProtocol.Protocol.Span.SequenceEqual("Handshake"u8))
```
wfurt, line 123: any reason why not to use existing values? There are some optimizations for common values (like h2) and this is going to be somewhat more expensive. I'm not sure if that matters but I feel it would be nice if the benchmark represents common scenario.

scenarios/sslstream.benchmarks.yml
rzikm: It was default before, but @carnaviire argued that we should be explicit about the version for consistency.

src/BenchmarksApps/SslStream/Server/Program.cs
rzikm: After we spoke offline about reporting connection errors, I think we need to leave this in to have opportunity to actually report them.

src/BenchmarksApps/SslStream/Client/Program.cs
rzikm: I copied default values from HttpClient benchmarks, both warmup and duration are 15s by default.

rzikm: handshake is not accounted for in either of those in the readwrite scenario

src/BenchmarksApps/SslStream/Server/Program.cs
rzikm: I wanted to have the server agnostic about what scenario (i.e. only client chooses the scenario and you don't have to restart the server) and didn't want to do the selection by sending an explicit message from the client.

We potentially could assign "h2" and "h3" values to "handshake" and "readwrite" scenarios. to make sure the optimization kicks in

src/BenchmarksApps/SslStream/Client/Program.cs
rzikm: I would also expect differences in Context vs non-context. And once we implement TLS resume for client certs (https://github.com/dotnet/runtime/issues/94561), then I would expect cert callback to be the slowest.

src/BenchmarksApps/SslStream/Server/Program.cs
rzikm: Changed to report number of errors.

src/BenchmarksApps/SslStream/Client/Program.cs
```diff
@@ -0,0 +1,249 @@
...
+
+    static SslClientAuthenticationOptions CreateSslClientAuthenticationOptions(ClientOptions options)
+    {
+        var sslOptions = new SslClientAuthenticationOptions
+        {
+            TargetHost = options.TlsHostName ?? options.Hostname,
+            RemoteCertificateValidationCallback = delegate { return true; },
+            ApplicationProtocols = new List<SslApplicationProtocol> { new(options.Scenario.ToString()) },
+#if NET8_0_OR_GREATER
+            AllowTlsResume = options.AllowTlsResume,
```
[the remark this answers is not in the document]
rzikm: I'd prefer to keep this in as a knob to allow us benchmarking TLS resume vs non-resume scenarios, e.g. when we implement it on linux with client certs.

rzikm:
@CarnaViire, @wfurt  PTAL

scenarios/sslstream.benchmarks.yml
```diff
@@ -0,0 +1,95 @@
...
+
+  sslClient:
+    source:
+      repository: https://github.com/aspnet/benchmarks.git
+      branchOrCommit: main
+      project: src/BenchmarksApps/SslStream/Client/SslStreamClient.csproj
+    isConsoleApp: true
+    waitForExit: true
+    timeout: 300 #seconds
+    arguments: "--host {{secondaryAddress}} --port {{serverPort}} --receive-buffer-size {{receiveBufferSize}} --send-buffer-size {{sendBufferSize}} --tls-version {{tlsVersion}} --cert-selection {{clientCertSelection}} --scenario {{scenario}} --warmup 00:00:{{warmup}} --duration 00:00:{{duration}}"
```
CarnaViire, line 95: hmmm I think the first one should be "host" instead of "secondaryAddress" 🤔 so that if you change the address, you change it in variables only

```suggestion
    arguments: "--host {{host}} --port {{serverPort}} --receive-buffer-size {{receiveBufferSize}} --send-buffer-size {{sendBufferSize}} --tls-version {{tlsVersion}} --cert-selection {{clientCertSelection}} --scenario {{scenario}} --warmup 00:00:{{warmup}} --duration 00:00:{{duration}}"

```

src/BenchmarksApps/SslStream/Client/Program.cs
```diff
@@ -0,0 +1,249 @@
...
+        }
+
+        using var sock = new Socket(SocketType.Stream, ProtocolType.Tcp);
+        await sock.ConnectAsync(options.Hostname, options.Port).ConfigureAwait(false);
+        using var stream = await EstablishSslStreamAsync(sock, options).ConfigureAwait(false);
+
+        byte[] recvBuffer = new byte[options.ReceiveBufferSize];
+
+        s_isRunning = true;
+        var writeTask = Task.Run(() => WritingTask(stream, options.SendBufferSize));
```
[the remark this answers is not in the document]
CarnaViire: This is good point; we might want to add a comment just so we would remember the reasoning in the future 🤔 you don't have to do this in this PR though

src/BenchmarksApps/SslStream/Server/Program.cs
```diff
@@ -0,0 +1,205 @@
...
+                var clientSock = await sock.AcceptAsync(ctrlC.Token).ConfigureAwait(false);
+                _ = Task.Run(() => ProcessClient(clientSock, options, sslOptions, ctrlC.Token));
+            }
+        }
+        catch (OperationCanceledException)
+        {
+            // Expected, just return
+        }
+
+        LogMetric("sslstream/errors", s_errors);
```
CarnaViire, line 57: Does this _actually_ report the error count when there are any? 🤔 I mean, within the Crank environment

src/BenchmarksApps/SslStream/Server/Program.cs
```diff
@@ -0,0 +1,205 @@
...
+            }
+        }
+        catch (IOException e) when (e.InnerException is SocketException ex && ex.SocketErrorCode == SocketError.ConnectionReset)
+        {
+            // client closed the connection on us, this is expected as clients
+            // simply close the socket after the test is done.
+        }
+        catch (Exception e)
+        {
+            s_errors++;
```
CarnaViire, line 145: this needs to be an interlocked operation

src/BenchmarksApps/SslStream/Server/Program.cs
rzikm: Yes

src/BenchmarksApps/SslStream/Client/Program.cs
```diff
@@ -0,0 +1,258 @@
...
+                await sock.ConnectAsync(options.Hostname, options.Port).ConfigureAwait(false);
+                await using var stream = await EstablishSslStreamAsync(sock, options).ConfigureAwait(false);
+                s_counters.TotalHandshakes++;
+            }
+        }
+
+        s_isRunning = true;
+        var task = Task.Run(() => DoRun(options));
+        await Task.Delay(options.Warmup).ConfigureAwait(false);
+        s_isRunning = false;
```
CarnaViire, line 89: Hmmm this wasn't _exactly_ what I had in mind, I remember that the idea behind the HttpClient benchmark approach was to not actually _stop_ between warmup and the "real" metric collection -- so that any additional cost of creating/growing a state machine etc will not affect the metrics (that being said, I don't know _exactly_ how much it affects the metrics; we had this more as a theoretical discussion with Geoff and Cory when I did the initial work on HttpClient benchmarks)

We can leave it as it is right now, I don't want to hold this back much longer, but we might want to address this later.

CarnaViire: (now I see that read-write scenarios seem to follow the pattern from HttpClient benchmarks, but handshakes don't 😅)

scenarios/sslstream.benchmarks.yml
```diff
@@ -0,0 +1,89 @@
...
+      job: sslServer
+      agent: secondary
+      variables:
+        receiveBufferSize: 0
+    client:
+      job: sslClient
+      agent: main
+      variables:
+        scenario: ReadWrite
+        sendBufferSize: 0
```
CarnaViire, line 46: I think this and `client-write` scenarios should also include `host: "{{secondaryAddress}}"`, shouldn't they? 🤔

CarnaViire:
Ping @rzikm? 😊

src/BenchmarksApps/SslStream/Client/Program.cs
```diff
@@ -0,0 +1,258 @@
...
+        BenchmarksEventSource.Register("sslstream/handshake/mean", Operations.Avg, Operations.Avg, "Mean handshake duration (ms)", "Handshakes duration in milliseconds - mean", "n3");
+
+        static async Task DoRun(ClientOptions options)
+        {
+            while (s_isRunning)
+            {
+                using var sock = new Socket(SocketType.Stream, ProtocolType.Tcp);
+                await sock.ConnectAsync(options.Hostname, options.Port).ConfigureAwait(false);
+                await using var stream = await EstablishSslStreamAsync(sock, options).ConfigureAwait(false);
+                s_counters.TotalHandshakes++;
```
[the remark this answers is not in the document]
rzikm: For measuring the handshake duration, it does not make sense IMO to run multiple clients, since we are essentially measuring latency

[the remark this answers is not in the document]
rzikm: we can expand this to multiple threads in the future if we find use for it

src/BenchmarksApps/SslStream/Client/Program.cs
rzikm: Converted this to use similar approach as in httpclient benchmarks, and even made it report percentiles

scenarios/sslstream.benchmarks.yml
```diff
@@ -0,0 +1,89 @@
...
+
+  sslClient:
+    source:
+      repository: https://github.com/aspnet/benchmarks.git
+      branchOrCommit: main
+      project: src/BenchmarksApps/SslStream/Client/SslStreamClient.csproj
+    isConsoleApp: true
+    waitForExit: true
+    timeout: 300 #seconds
+    arguments: "--host {{host}} --port {{serverPort}} --receive-buffer-size {{receiveBufferSize}} --send-buffer-size {{sendBufferSize}} --tls-version {{tlsVersion}} --cert-selection {{clientCertSelection}} --scenario {{scenario}} --warmup 00:00:{{warmup}} --duration 00:00:{{duration}}"
```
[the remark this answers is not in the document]
rzikm: fixed

src/BenchmarksApps/SslStream/Client/Program.cs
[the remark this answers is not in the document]
rzikm: added comment

CarnaViire:
LGTM, but why there is a submodule added? 😅
image

rzikm:
No Idea how it got there, should be gone now

CarnaViire (approved):

PR #1941, opened 2024-02-01
Add pr and ci support for Arm64 Azure VM runs

DrewScoggins:


DrewScoggins:
@sebastienros PTAL

DrewScoggins:
This still needs to wait for the service connection to be added to the AzDO by the engineering team.

build/azure.profile.yml
```diff
@@ -50,3 +50,53 @@ profiles:
...
+
+  aspnet-azurearm64-lin-relay:
+    variables:
+      serverPort: 5000
+      serverAddress: 10.0.0.4
+      cores: 4
+    jobs: 
+      db:
+        endpoints: 
+          - https://aspnetperf.servicebus.windows.net/azuredb
```
sebastienros, line 87: Did you mean `/azurearm64db` (same comment for the other endpoints)

DrewScoggins: Yes, good catch.

build/benchmarks.matrix.azure.yml
```diff
@@ -15,24 +15,44 @@ groups:
       profiles:
       - aspnet-azure-lin
 
+    - name: Trends Database Azure Arm64 Linux
```
sebastienros, line 18: Two jobs per group means it needs two queues at the top of the file. So we need to create another one in AzDo, might need to ask First Responders for that. `ASP.NET Benchmarks Azure Queue 2` for instance (not mentioning a name so we can use it for something else later). And create a new queue in the service bus service for it.

DrewScoggins: Sure, I will use a generic name for now.

PR #1944, opened 2024-02-27
Add Lighthouse benchmarks

MackinnonBuck:
## Add Lighthouse benchmarks

This PR includes the following changes:
* Added Lighthouse testing job
  * Runs against the specified URL, so it's a general-purpose job that's not tied to a specific application
  * Allows for running Lighthouse multiple times and averaging the results
  * Allows an additional "warm-up" run so that the tests run against a fully initialized server
    * Note: The warm-up does not affect the client. We currently clear the Chrome cache between each run, but we could consider adding tests that run against a client cached assets.
  * Tests the following metrics:
    * First Contentful Paint
    * Interaction to Next Paint †
    * Largest Contentful Paint
    * Total Blocking Time
    * Time to Interactive
* Added a "Run template" job
  * Creates, publishes, and runs a project created from an arbitrary project template
  * It's set up in a way that makes it easy to add additional template scenarios (only `.yml` config changes required)
* Added Lighthouse scenarios for the Blazor Web App template:
  * Includes:
    * With Server interactivity
    * With WebAssembly interactivity

Still TODO:
- [x] Add more Lighthouse scenarios
  - [x] Blazor Web App with Auto interactivity
  - [ ] Blazor WebAssembly standalone
- [x] † Add support for [Lighthouse user flows](https://web.dev/articles/lighthouse-user-flows). This will allow us to collect metrics for scenarios other than loading pages (e.g., typing into a text field, clicking a button/link, etc.).
- [x] Utilize the changes from https://github.com/dotnet/crank/pull/703
- [x] Document how to run these benchmarks locally

Fixes https://github.com/dotnet/aspnetcore/issues/29303


sebastienros (approved):
Pending changing source branches to `main`
Next PR will be adding it to the CI

PR #1953, opened 2024-03-11
Add set rate runs for Azure VMs

DrewScoggins:
I was planning on putting this data into a new SQL table called `ConstrainedBenchmarks` if you think it should go somewhere else let me know.


build/benchmarks.matrix.azure.yml
```diff
@@ -56,4 +56,15 @@ groups:
       template: baselines-scenarios.yml
       profiles:
       - aspnet-azurearm64-lin
+  
+  - jobs:
+      - name: Constant Requests Azure Arm64 Linux
+        template: rate-limited-scenarios.yml
+        profiles:
+        - aspnet-azurearm64-lin
```
sebastienros, line 64: What about using the current containers-scenarios file and also run the other frameworks with the same frequency. This way all the combinations will work whatever machine is selected in the dashboard, and less templates to maintain.

DrewScoggins: I wasn't sure if we really cared about running against the actual CPU and memory constrained runs, but I see that we do run an unlimited scenario, so we can just ignore the other results.

build/containers-scenarios.yml
```diff
@@ -40,6 +40,8 @@ parameters:
       arguments: --variable rate=1000 --property rate=1000 --load.job bombardier
     - displayName: 10000
       arguments: --variable rate=10000 --property rate=10000 --load.job bombardier
+    - displayName: 20000
+      arguments: --variable rate=10000 --property rate=10000 --load.job bombardier
```
sebastienros, line 44:
```suggestion
      arguments: --variable rate=20000 --property rate=20000 --load.job bombardier
```

build/containers-scenarios.yml
```diff
@@ -40,6 +40,8 @@ parameters:
       arguments: --variable rate=1000 --property rate=1000 --load.job bombardier
     - displayName: 10000
       arguments: --variable rate=10000 --property rate=10000 --load.job bombardier
+    - displayName: 20000
+      arguments: --variable rate=20000 --property rate=20000 --load.job bombardier
```
sebastienros, line 44:
```suggestion
    - displayName: 20000
      arguments: --variable rate=20000 --property rate=20000 --load.job bombardier
```

Let's remove it for now and add it if it's asked for. That would double the number of runs, and for fortunes it's already at 50% (and more) of CPU with 10K.

PR #1955, opened 2024-03-13
Add SslStream Benchmarks to regular CI runs

rzikm:


build/sslstream-scenarios.yml
```diff
@@ -0,0 +1,88 @@
...
+                    displayName: ${{ version.displayName }} ${{ s.displayName }} ${{ res.displayName }}
+                    inputs:
+                      connectedServiceName: ${{ parameters.connection }}
+                      waitForCompletion: true
+                      useDataContractSerializer: "false"
+                      messageBody: |
+                        {
+                          "condition": "(${{ parameters.condition }}) && ((new Date().getUTCHours() - 7 + 24) % 24 >= 13)",
+                          "name": "crank",
+                          "args": [ "--client.framework net9.0 --server.framework net9.0 --command-line-property --table HttpClientBenchmarks --sql SQL_CONNECTION_STRING --chart --session $(session) ${{ parameters.arguments }} --no-metadata --no-measurements ${{ s.arguments }} ${{ version.arguments }} ${{ res.arguments }} ${{ c.arguments }}" ]
```
sebastienros, line 68: The existing benchmarks are already stored in the HttpClientBenchmarks table but these ones have different properties so this would make sense to use a distinct table, and even more if they have to show up in a different PBI page with custom filters (concurrency, version, ...).

e.g., SslStreamBenchmarks

build/sslstream-scenarios.yml
```diff
@@ -0,0 +1,88 @@
...
+steps:
+  # handshake scenarios
+  - ${{ each s in parameters.handshakeScenarios }}:
+      - ${{ each version in parameters.tlsVersion }}:
+          - ${{ each res in parameters.tlsResume }}:
+              - ${{ each c in parameters.concurrency }}:
+                  - task: PublishToAzureServiceBus@1
+                    condition: succeededOrFailed()
+                    timeoutInMinutes: 10
+                    displayName: ${{ version.displayName }} ${{ s.displayName }} ${{ res.displayName }}
```
sebastienros, line 59: c.displayname too?

build/sslstream-scenarios.yml
```diff
@@ -0,0 +1,88 @@
...
+                    inputs:
+                      connectedServiceName: ${{ parameters.connection }}
+                      waitForCompletion: true
+                      useDataContractSerializer: "false"
+                      messageBody: |
+                        {
+                          "condition": "(${{ parameters.condition }}) && ((new Date().getUTCHours() - 7 + 24) % 24 >= 13)",
```
sebastienros, line 66:
```suggestion
                          "condition": "(${{ parameters.condition }}) && (Math.round(Date.now() / 43200000) % 2 == 0 )",
```

This is another way to express once every 2 half-days which I use now. Another example of how it's used: https://github.com/aspnet/Benchmarks/blob/main/build/containers-scenarios.yml#L53

build/sslstream-scenarios.yml
sebastienros: Please add this argument `--description \"${{ version.displayName }} ${{ s.displayName }} ${{ res.displayName }} ${{ c.displayName }} $(System.JobDisplayName)\" --property buildId=\"$(buildId)\" --property buildNumber=\"$(buildNumber)\"` which will allow us to add regression detection.

I use this field and the scenario name to group all runs (by unique combination of arguments).

CarnaViire: also, it seems that `$(sslStreamJobs)` is not added anywhere, we'd need to pass it as well so the `--config` argument would be specified in the resulting command line

sebastienros: @CarnaViire it's coming from job-variables.yml which is added in the main ADO template.

build/sslstream-scenarios.yml
```diff
@@ -0,0 +1,88 @@
...
+    default: ""
+  - name: condition
+    type: string
+    default: "true"
+
+  - name: readWriteScenarios
+    type: object
+    default:
+      - displayName: "read-write"
+        arguments: --scenario read-write --client.options.collectCounters true
```
CarnaViire, line 18:
```suggestion
        arguments: --scenario read-write --client.options.collectCounters true --property scenario=read-write
```

build/sslstream-scenarios.yml
```diff
@@ -0,0 +1,88 @@
...
+
+  - name: handshakeScenarios
+    type: object
+    default:
+      - displayName: "read-write"
+        arguments: --scenario handshake --client.options.collectCounters true
```
CarnaViire, line 24:
```suggestion
        arguments: --scenario handshake --client.options.collectCounters true --property scenario=handshake
```

sebastienros: Yes, the scenario property should be set, but this can be nicer, like "Handshake". This will be used in the PBI dashboard as a list to filter charts.

CarnaViire (approved):
LGTM, thanks!

sebastienros:
LGTM2

Any feedback on the process? What needs more documentation, what feels wrong/weird, ...

rzikm:
> LGTM2
> 
> Any feedback on the process? What needs more documentation, what feels wrong/weird, ...

Natalia gave me some pointers before I started, so I knew what to do and knew where to take inspiration from. I did not look for any existing documentation, but if I were going to this blind then I think document with following would be useful:
- which files to edit (and that one is generated using some templating engine)
- any constraints (e.g. the benchmark matrix yml has groups such that no two jobs use the same profile/machines at the same time)
- some info on how the data flows from the benchmarks to the board (e.g. how to setup saving the results to the database, and how/which properties to add so that the data can be processed)

PR #1968, opened 2024-04-01
Add Siryn machine CI runs

LoopedBard3:
Add Siryn ci runs, move SslStream ci run to new slot, and add ci.profile for intel-load-app so intel-load can be used as the app.

### DIFF
--- a/build/benchmarks.matrix.01.yml
+++ b/build/benchmarks.matrix.01.yml
@@ -219,6 +219,13 @@ groups:
       - amd-lin-load
       - amd-lin2-db # required by nodejs and tfb-database argument
 
+    - name: Trends Siryn Linux
+      template: trend-scenarios.yml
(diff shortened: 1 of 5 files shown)


build/benchmarks.matrix.01.yml
```diff
@@ -219,6 +219,13 @@ groups:
       - amd-lin-load
       - amd-lin2-db # required by nodejs and tfb-database argument
 
+    - name: Trends Siryn Linux
+      template: trend-scenarios.yml
+      profiles:
+      - siryn-lin-app
```
sebastienros, line 225: Only the database version needs a db agent

LoopedBard3: Good catch, fixed 👍

sebastienros (approved):

PR #1970, opened 2024-04-08
Socket benchmarks initial implementation

liveans:
Adds benchmark for Socket
Two different scenario implemented: ConnectionEstablishment, ReadWrite


sebastienros:
FYI, a related one I made to measure how many packets per second a set of machines can handle.

https://github.com/sebastienros/PacketsPerSecond/blob/master/Program.cs

liveans:
Currently dealing with a task in QUIC, I plan to get back to this early next week.

/cc: @CarnaViire

liveans:
It's ready for review now @CarnaViire, there are fields and methods which is duplicating with Tls but decided not to touch them for now, I'll address them as a follow-up later, want merge this first.

src/System/Net/Benchmarks/Common/Shared/SingleStreamConnection.cs
```diff
@@ -3,9 +3,18 @@
 
 namespace System.Net.Benchmarks;
 
-internal abstract class SingleStreamConnection<TStream> : IConnection
```
liveans, line 6: There were no abstract method in this class, so I deleted it.

scenarios/socket.benchmarks.yml
```diff
@@ -0,0 +1,134 @@
...
+        receiveBufferSize: 0
+      onConfigure:
+        - '
+          if (String(job.Variables.receiveBufferSize) != "0")
+          {
+            throw "Incorrect receiveBufferSize for client-write scenario. Expected: 0, got: " + job.Variables.receiveBufferSize;
+          }
+          '
+
+  connection-establishment:
```
CarnaViire, line 90: It is generally better to have shorter names, if possible, bc it is easier to use them in command line :)
```suggestion
  connect:
```

src/System/Net/Benchmarks/Apps/Socket/Client/Program.cs
```diff
@@ -0,0 +1,247 @@
...
+            catch (Exception ex)
+            {
+                Log("Exception: " + ex);
+                exceptionRequests++;
+            }
+        }
+        var elapsed = sw.ElapsedTicks * 1.0 / Stopwatch.Frequency;
+        return (successRequests / elapsed, exceptionRequests);
+    }
+}
```
CarnaViire, line 247: Can you pls unify the file endings so that all of the files end with a new line?

src/System/Net/Benchmarks/Apps/Socket/Client/Program.cs
```diff
@@ -0,0 +1,247 @@
...
+        {
+            if (isWarmup && !IsWarmup)
+            {
+                isWarmup = false;
+                values.Clear();
+                OnWarmupCompleted();
+            }
+            sw.Restart();
+            await using var connection = await establishConnectionAsync(options).ConfigureAwait(false);
+            await using var stream = await connection.EstablishStreamAsync(options).ConfigureAwait(false);
```
CarnaViire, line 116: Wait, I am a bit confused... we discussed that we need socket benchmarks that call actual Socket APIs :) so that we can have/add scenarios that are impossible to do using Stream APIs (e.g. UDP socket benchmarks)

When we've discussed NetworkStream previously, in my mind that was about an _additional_ benchmark, more intended to compare with SslStream to see the TLS overhead

liveans: I see, looks like I got confused from our last discussion as well, and didn't actually realize I should be using direct socket APIs here, I'll change the design to use Socket APIs directly.

src/System/Net/Benchmarks/Common/Shared/SingleStreamConnection.cs
CarnaViire: I believe we'd need to revert the changes in this file.

It is abstract because it cannot be used directly. The protected `InnerStream` property is technically playing the role of the "abstract method". There's no actual abstract method bc the class is used as a base class in both server and client, and they have different consuming methods.

We can pull the full single-stream client connection implementation -- but only a part of server connection implementation -- "up". But it would still leave `SingleStreamConnection` abstract.

There's how I see it:

<details>

**1. (move "up" from `Tls` to `Common`)**
```diff
-   ITlsBenchmarkClientConnection
+   IStreamingClientConnection<TOptions>
```

**2. add `IStreamingServerConnection : IConnection` with `AcceptInboundStreamAsync` method pulled up from `ITlsBenchmarkServerConnection`**

**3. (remove `AcceptInboundStreamAsync` from `ITlsBenchmarkServerConnection`)**
```diff
-   ITlsBenchmarkServerConnection : IConnection
+   ITlsBenchmarkServerConnection : IStreamingServerConnection
```

**4. Keep `SingleStreamConnection` abstract**

**5. (move "up" from `Apps` to `Common`)**
```diff
-   SslStreamClientConnection : SingleStreamConnection<SslStream>, ITlsBenchmarkClientConnection
+   SingleStreamConnectionClientConnection<TStream, TOptions> : SingleStreamConnection<TStream>, IStreamingClientConnection<TOptions>
```

**6. add `SingleStreamServerConnection<TStream, TOptions> : SingleStreamConnection<TStream>, IStreamingClientConnection<TOptions>` with `AcceptInboundStreamAsync` method impl pulled up from `SslStreamBenchmarkServerConnection`**

**7. (remove `AcceptInboundStreamAsync` from `SslStreamBenchmarkServerConnection `)**
```diff
-   SslStreamBenchmarkServerConnection : SingleStreamConnection<SslStream>, ITlsBenchmarkServerConnection
+   SslStreamBenchmarkServerConnection : SingleStreamServerConnection<SslStream>, ITlsBenchmarkServerConnection
```

</details>

This definitely can be done as a follow-up (I mean, not in this PR), and it is even still up to discussion. It technically might be more trouble than worth, but it can support the positioning of "NetworkStream" benchmarks alongside Tls benchmarks as a group of "streaming" benchmarks.

src/System/Net/Benchmarks/Apps/Socket/Client/Program.cs
```diff
@@ -0,0 +1,247 @@
+﻿using System.Diagnostics;
+using System.Net;
+using System.Net.Benchmarks;
+using System.Net.Benchmarks.SocketBenchmark.Client;
+using System.Net.Benchmarks.SocketBenchmark.Shared;
+using System.Net.Sockets;
+
+return await SocketBenchmarkClient.RunAsync(args).ConfigureAwait(false);
+
+internal class SocketBenchmarkClient : BenchmarkClient<SocketClientOptions>
```
CarnaViire, line 10: I was thinking, and I still believe we'd need to separate these benchmarks (which are technically `NetworkStream` benchmarks) from the to-be-added pure-Socket-API benchmarks. Otherwise it would be hard to navigate within the benchmark class -- this one is big enough already.

And creating a socket is technically a two-liner (`var socket = new Socket(...); await socket.ConnectAsync(...);`) and it can/should be potentially different in new benchmarks so we can maybe set some socket options based on passed parameters. So IMO there's not really much to reuse? Or am I missing something?

So, currently the proposal is to rename
`SocketBenchmark` -> `NetworkStreamBenchmark`,
`src/System/Net/Benchmarks/Apps/Socket/` -> `src/System/Net/Benchmarks/Apps/Streaming/`,
and maybe even
`src/System/Net/Benchmarks/Apps/Tls/` -> `src/System/Net/Benchmarks/Apps/Streaming/Tls/`

WDYT?

cc @rzikm for additional opinions

liveans: > I was thinking, and I still believe we'd need to separate these benchmarks (which are technically NetworkStream benchmarks) from the to-be-added pure-Socket-API benchmarks. Otherwise it would be hard to navigate within the benchmark class -- this one is big enough already.

After thinking a while on my own, I totally agree with this. I'm planning to change this as `NetworkStreamBenchmark` and add another Benchmark for pure Socket APIs, because we can extend those Socket Benchmarks with UDP and possibly(?) with other protocols as well in the future.

> And creating a socket is technically a two-liner (var socket = new Socket(...); await socket.ConnectAsync(...);) and it can/should be potentially different in new benchmarks so we can maybe set some socket options based on passed parameters. So IMO there's not really much to reuse? Or am I missing something?

I also think that we don't have much to reuse here and it is highly possible that some benchmarks can customize their usages of socket while creating connection (even during connection?), maybe we can create a `ConnectCallback` or something like that as we have in `SocketsHttpHandler`. (This is what I'm thinking from top of my head)

> So, currently the proposal is to rename
SocketBenchmark -> NetworkStreamBenchmark,
src/System/Net/Benchmarks/Apps/Socket/ -> src/System/Net/Benchmarks/Apps/Streaming/,
and maybe even
src/System/Net/Benchmarks/Apps/Tls/ -> src/System/Net/Benchmarks/Apps/Streaming/Tls/

The only question that comes in my mind is, should we keep QUIC under Tls folder here. Technically, it's a stream protocol as well but do we want to keep it as separated? Or you don't have any problem with moving it with Tls?

src/System/Net/Benchmarks/Common/Shared/SingleStreamConnection.cs
liveans: I'm planning to do this as a follow-up, I don't want to mix refactoring with this PR.

src/System/Net/Benchmarks/Common/Shared/MetricName.cs
```diff
@@ -18,4 +18,5 @@ public static class MetricName
     public const string Rps = "/rps";
     public const string Errors = "/errors";
     public const string Handshake = "/handshake";
+    public const string ConnectionEstablishment = "/connection-establishment";
```
CarnaViire, line 21: Nit: I think it might be better to unify scenario (and metric) name with the job name: connection establishment -> connect

src/System/Net/Benchmarks/Common/Shared/SingleStreamConnection.cs
```diff
@@ -3,9 +3,18 @@
...
+internal class SingleStreamConnection<TStream> : IConnection
```
CarnaViire, line 6: Nit: Can we pls revert the changes in this file as I asked in https://github.com/aspnet/Benchmarks/pull/1970/files#r1611511135?

CarnaViire (approved):
LGTM, thanks!

PR #1972, opened 2024-04-15
Common System.Net benchmark setup + QUIC benchmark to compare with SslStream

CarnaViire:
Also:
- Moved SslStream benchmark to the common setup and fixed minor bugs.
- Added a single solution for all System.Net benchmarks & some msbuild props.
- Added a second level of abstraction to adapt for non-stream-based and/or non-tls-specific benchmarks  (e.g. Sockets).
- Added a new "RPS" scenario to QUIC & SslStream benchmarks.
- Validated SslStream benchmark numbers didn't change with the new setup.

cc @rzikm @liveans


build/sslstream-scenarios.yml
```diff
@@ -43,9 +43,9 @@ parameters:
     type: object
     default:
       - displayName: "1 connection"
-        arguments: --variable concurrency=1 --property concurrency=1
+        arguments: --variable connections=1 --property concurrency=1
```
rzikm, line 46: Should we also change the property label?

```suggestion
        arguments: --variable connections=1 --property connections=1

```

build/sslstream-scenarios.yml
```diff
@@ -43,9 +43,9 @@ parameters:
     type: object
     default:
       - displayName: "1 connection"
-        arguments: --variable concurrency=1 --property concurrency=1
+        arguments: --variable connections=1 --property concurrency=1
       - displayName: "16 connections"
-        arguments: --variable concurrency=16 --property concurrency=16
+        arguments: --variable connections=16 --property concurrency=16
```
rzikm, line 48:
```suggestion
        arguments: --variable connections=16 --property connections=16

```

src/System/Net/Benchmarks/Apps/Tls/Quic/Readme.md
```diff
@@ -0,0 +1,38 @@
+# Quic Benchmark Apps
+
+Contains a pair of applications for measuring the performance of  and QuicStream.
```
rzikm, line 3:
```suggestion
Contains a pair of applications for measuring the performance of QuicStream.
```

scenarios/sslstream.benchmarks.yml
```diff
@@ -12,10 +12,10 @@ variables:
   tlsVersion: "1.3" # <1.2|1.3>
   serverCertSelection: "CertContext" # <Callback|Certificate|CertContext>
   clientCertSelection: "CertContext" # <Callback|Collection|CertContext>
-  allowTlsResume: true
+  allowTlsResume: false # can be true for .NET 8+ only, defaults to false for parity with older versions (and QUIC)
```
rzikm, line 15: TLS Resume is on by default for platforms where it is implemented (.NET 7+), IMO users don't explicitly disable it unless there is a problem so we should use it benchmarks I think. (btw it should affect only the handshake benchmarks)

src/System/Net/Benchmarks/Apps/Tls/SslStream/Server/SslStreamServerOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Security.Authentication;
+
+namespace System.Net.Benchmarks.Tls.SslStreamBenchmark;
+
+public class SslStreamServerOptions : TlsBenchmarkServerOptions, ISslStreamExtraOptions
+{
+    public bool DisableTlsResume { get; set; }
+    public bool AllowTlsResume { get; set; }
```
rzikm, line 11: Those two are inverse of each other, so having only one should be enough.

rzikm (approved):
Looks great so far! Few comments

build/sslstream-scenarios.yml
CarnaViire: I was going back and forth on this. On one hand, we already should have this table filled with some previous data, so renaming the column won't look pretty there... On the other hand, it would be nice to have QuicStream's result in the same table to be able to show them back to back in PowerBI. So I was thinking -- I will rename the table as well (to a common name) and it will also automatically solve the column problem :D

scenarios/sslstream.benchmarks.yml
CarnaViire: Ok, makes sense.

liveans (approved):
Looks great!

CarnaViire:
This PR doesn't touch any code that's outside our (NCL) direct responsibility (beside adding 2 props to `build/dependencies.props`). I'm going to merge this, given I have 2 approves, to unblock @liveans work.

PR #1979, opened 2024-04-23
Add large Json scenario

BrennanConroy:
* Create a new Json project where we can add crazy 🤪 json scenarios.
* Adds new json scenario where we send a large json payload (arbitrarily chose 400k string).
  - Want this in so we can see improvements from adding Pipelines to Json


src/BenchmarksApps/Json/JsonMiddleware.cs
```diff
@@ -0,0 +1,58 @@
...
+    private const int _jsonFramingSize = 14;
+
+    private readonly int _bufferSize;
+    private readonly RequestDelegate _next;
+
+    private readonly string _message = new string("Hello, World!");
+
+    public JsonMiddleware(RequestDelegate next)
+    {
+        if (!int.TryParse(Environment.GetEnvironmentVariable("JSONSIZE"), out var length))
```
davidfowl, line 19: No IConfiguration??? 😄

src/BenchmarksApps/Json/Properties/launchSettings.json
```diff
@@ -0,0 +1,29 @@
...
+        "ASPNETCORE_ENVIRONMENT": "Development"
+      }
+    },
+    "IIS Express": {
+      "commandName": "IISExpress",
+      "launchBrowser": true,
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development"
+      }
+    }
```
DamianEdwards, line 27: Delete this

src/BenchmarksApps/Json/Properties/launchSettings.json
```diff
@@ -0,0 +1,29 @@
+﻿{
+  "$schema": "http://json.schemastore.org/launchsettings.json",
+  "iisSettings": {
+    "windowsAuthentication": false,
+    "anonymousAuthentication": true,
+    "iisExpress": {
+      "applicationUrl": "http://localhost:40856",
+      "sslPort": 0
+    }
+  },
```
DamianEdwards, line 10: Delete this

scenarios/json.benchmarks.yml
```diff
@@ -79,3 +92,16 @@ scenarios:
...
         path: /json
         presetHeaders: json
+
+  json_400k:
+    application:
+      job: jsonbenchmarks
+      variables:
+        scenario: json
+      environmentVariables:
+        JSONSIZE: 400000
```
sebastienros, line 102: FYI could be a command line argument plus variable and have a default value. Something like "connections" in client scenarios.

scenarios/json.benchmarks.yml
```diff
@@ -33,6 +34,18 @@ jobs:
...
+    source:
+      repository: https://github.com/aspnet/benchmarks.git
+      branchOrCommit: main
+      project: src/BenchmarksApps/Json/Json.csproj
+    readyStateText: Application started.
+    variables:
+      protocol: http
+      server: Kestrel
+      scenario: json
+    arguments: "--nonInteractive true --scenarios {{scenario}} --urls {{serverScheme}}://{{serverAddress}}:{{serverPort}} --server {{server}} --protocol {{protocol}}"
```
sebastienros, line 47:
```suggestion
    variables:
    arguments: "--urls {{serverScheme}}://{{serverAddress}}:{{serverPort}}"
```

I don't think these are used, probably a copy from the main Benchmarks project.

build/trend-scenarios.yml
```diff
@@ -82,6 +82,8 @@ parameters:
     arguments: --scenario mapaction $(jsonJobs) --property scenario=JsonMapAction --property protocol=http
   - displayName: Json Minimal APIs
     arguments: --scenario json $(minimalJobs) --property scenario=JsonMinimalApis --property protocol=http
+  - displayName: Json 400k
+    arguments: --scenario json_400k --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/json.benchmarks.yml --property scenario=Json400k
```
sebastienros, line 86: You don't have to create specific scenarios if this is mostly for CI, just one and set the variable in this line for each value you want to test. More flexible.

BrennanConroy: I don't understand this comment.

sebastienros: `--scenario json_400k` means you created a scenario named `json_400k`. Now let's say you want to test with 1, 42, 1000, ... you would create a new scenario for each. It's nice if people need to chose from these, but I believe in our case this is just for CI, you can create a single scneario called "AllTheJsonsizes" (joking), and use the same scenario name, but change a variable which contains the size. In the end this file (azdo script) would look like:

```
- displayName: Json 400k
  arguments: --scenario AllTheJsonsizes --variable jsonSize=400000 --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/json.benchmarks.yml --property scenario=Json400k

- displayName: Json 1k
  arguments: --scenario AllTheJsonsizes --variable jsonSize=1000 --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/json.benchmarks.yml --property scenario=Json1K
...
```

So you don't have to create new scenarios. The same way we can vary the number of connections for Plaintext without having to create several scenarios.

sebastienros (approved):

Issue #1983, opened 2024-05-10
Containers ASP.NET Composite benchmarks are broken


sebastienros:
```
Step 5/8 : FROM mcr.microsoft.com/dotnet/nightly/aspnet:9.0-alpine-composite AS runtime
manifest for mcr.microsoft.com/dotnet/nightly/aspnet:9.0-alpine-composite not found: manifest unknown: manifest tagged by "9.0-alpine-composite" is not found
```

in https://github.com/aspnet/Benchmarks/blob/main/docker/container-matrix/Middleware-composite.dockerfile

@ivdiazsa

ivdiazsa:
This looks extremely strange. I will take a look

ivdiazsa:
I've found the problem. They changed `9.0-alpine-composite` to `9.0-preview-alpine-composite`. I'll send a PR with the updated change

PR #1984, opened 2024-05-13
Fix ASP.NET Preview Tags

ivdiazsa:
This PR fixes issue #1983. The Middleware Dockerfiles were looking for the tag `9.0-alpine`, but they changed it to be `9.0-preview-alpine` instead. This PR updates the dockerfiles to reference the new tag.

cc @sebastienros


sebastienros:
Thank you!

PR #1985, opened 2024-05-15
Add Json 5M

BrennanConroy:


sebastienros (approved):
Could be possible that at max throughput we already reach the network bandwidth limit. Would be interesting to compare with a different stack to check latencies or other things than RPS in this case.

PR #1987, opened 2024-05-20
Update IPs to VMs in new subnets

DrewScoggins:


LoopedBard3 (approved):
Looks good assuming the IPs are correct and the core update is correct.

PR #1990, opened 2024-05-23
Update Go modules

LoopedBard3:
Updated Go modules using `go get -u`. Is there any other testing that needs to be done before merge?


DrewScoggins (approved):
LGTM

LoopedBard3:
@sebastienros is there any testing or other steps that should be done before merging?

PR #1995, opened 2024-06-05
Update azdo worker to use certificates authentication

sebastienros:


sebastienros:
/cc @LoopedBard3 @DrewScoggins 

This are the scripts which are installed for running AzdoDevOpsWorker. This should get updated automatically, but the Azure machines will need the new ENVs and the certificate.

PR #1998, opened 2024-06-27
Add option to enable multuple H/3 connections

ManickaP:
https://github.com/dotnet/runtime/pull/101531 follow up

cc @CarnaViire

### DIFF
--- a/build/httpclient-scenarios.yml
+++ b/build/httpclient-scenarios.yml
@@ -19,13 +19,13 @@ parameters:
   type: string
   default: '--client.framework net9.0 --server.framework net8.0' # net8.0 for server is a temporary workaround
 
-- name: getScenarios 
+- name: getScenarios
   type: object
   default:
   - displayName: "GET"
     arguments: --scenario httpclient-kestrel-get --variable useHttpMessageInvoker=true --property server=kestrel --property client=dotnetinvoker --property method=get --client.options.collectCounters true
 
-- name: postScenarios 
+- name: postScenarios
   type: object
   default:
   - displayName: "POST"
@@ -41,7 +41,9 @@ parameters:
     - displayName: "HTTP/2.0 (mult conn)"
       arguments: --variable httpVersion=2.0 --variable useHttps=true --variable http20EnableMultipleConnections=true --property httpversion=h2multconn
     - displayName: "HTTP/3.0"
-      arguments: --variable httpVersion=3.0 --variable useHttps=true --property httpversion=h3
+      arguments: --variable httpVersion=3.0 --variable useHttps=true --variable http30EnableMultipleConnections=false --property httpversion=h3
+    - displayName: "HTTP/3.0 (mult conn)"
+      arguments: --variable httpVersion=3.0 --variable useHttps=true --variable http30EnableMultipleConnections=true --property httpversion=h3mulconn
 
 - name: clientsxThreads
   type: object
--- a/build/httpclient-scenarios.yml
+++ b/build/httpclient-scenarios.yml
@@ -43,7 +43,7 @@ parameters:
     - displayName: "HTTP/3.0"
       arguments: --variable httpVersion=3.0 --variable useHttps=true --variable http30EnableMultipleConnections=false --property httpversion=h3
     - displayName: "HTTP/3.0 (mult conn)"
-      arguments: --variable httpVersion=3.0 --variable useHttps=true --variable http30EnableMultipleConnections=true --property httpversion=h3mulconn
+      arguments: --variable httpVersion=3.0 --variable useHttps=true --variable http30EnableMultipleConnections=true --property httpversion=h3multconn
 
 - name: clientsxThreads
   type: object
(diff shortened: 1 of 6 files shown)


src/BenchmarksApps/HttpClientBenchmarks/Clients/HttpClient/ClientOptionsBinder.cs
```diff
@@ -15,6 +15,7 @@ public class ClientOptionsBinder : BinderBase<ClientOptions>
         public static Option<int> ConcurrencyPerHttpClientOption { get; } = new Option<int>("--concurrencyPerHttpClient", () => 1, "Number of concurrect requests per one HttpClient");
         public static Option<int> Http11MaxConnectionsPerServerOption { get; } = new Option<int>("--http11MaxConnectionsPerServer", () => 0, "Max number of HTTP/1.1 connections per server, 0 for unlimited");
         public static Option<bool> Http20EnableMultipleConnectionsOption { get; } = new Option<bool>("--http20EnableMultipleConnections", () => true, "Enable multiple HTTP/2.0 connections");
+        public static Option<bool> Http30EnableMultipleConnectionsOption { get; } = new Option<bool>("--http30EnableMultipleConnections", () => true, "Enable multiple HTTP/3.0 connections");
```
CarnaViire, line 18: I wonder if we maybe should keep the default as false for the time being.
Otherwise if we merge it like that, all the CI tests will start using that by default, but I think it should be a clear parameter as with H/2 (h2 vs h2multiconn)
It's not a blocker, just a way not to mix up measurements.

Or we can change both crank config yml and ci test config yml right away to include the parameter

ManickaP: ???
Do you want me to change something here or not?

CarnaViire: **(1)**
"Keep the default as false" -- Minimal change
```diff
- ... new Option<bool>(..., () => true, ...);
+ ... new Option<bool>(..., () => false, ...);
```

-OR-

**(2)**
"Change crank config yml" -- Modify [scenarios/httpclient.benchmarks.yml](https://github.com/aspnet/Benchmarks/blob/9c454b7357ac0b032c48535df236b8bbb9588052/scenarios/httpclient.benchmarks.yml) to include new parameter as a variable (with default value = false), e.g.: https://github.com/aspnet/Benchmarks/blob/9c454b7357ac0b032c48535df236b8bbb9588052/scenarios/httpclient.benchmarks.yml#L27
and pass it to client, e.g.: `--http20EnableMultipleConnections {{http20EnableMultipleConnections}}`
https://github.com/aspnet/Benchmarks/blob/9c454b7357ac0b032c48535df236b8bbb9588052/scenarios/httpclient.benchmarks.yml#L171

-OR-

**(3)**
"Change both crank config yml and ci test config yml" -- (2) AND modify [build/httpclient-scenarios.yml](https://github.com/aspnet/Benchmarks/blob/9c454b7357ac0b032c48535df236b8bbb9588052/build/httpclient-scenarios.yml) to include new parameter with both true and false to test matrix, e.g.:
https://github.com/aspnet/Benchmarks/blob/9c454b7357ac0b032c48535df236b8bbb9588052/build/httpclient-scenarios.yml#L31-L34
and as `httpVersion` is already passed to the scenario (we're _not_ adding a new dimension), it should then appear automatically in "pretty graphs" as soon as the new value will appear in the result table.

---

> Do you want me to change something here or not?

It is hard for me to understand how urgently you'd like to merge the changes (I guess not really, given the PR's in draft), and how much effort are you willing to put into it.

My _preferred_ option is (3) but it _can_ be done as a follow-up, as long as the default is false.

But I am open to the discussion about the default as well.

ManickaP: > It is hard for me to understand how urgently you'd like to merge the changes

Not urgent at all, it just wasn't obvious to me from the original comment whether you expect any changes to this PR or not.

build/httpclient-scenarios.yml
```diff
@@ -33,7 +33,9 @@ parameters:
     - displayName: "HTTP/2.0 (mult conn)"
       arguments: --variable httpVersion=2.0 --variable useHttps=true --variable http20EnableMultipleConnections=true --property httpversion=h2multconn
     - displayName: "HTTP/3.0"
-      arguments: --variable httpVersion=3.0 --variable useHttps=true --property httpversion=h3
+      arguments: --variable httpVersion=3.0 --variable useHttps=true --variable http30EnableMultipleConnections=false --property httpversion=h3
+    - displayName: "HTTP/3.0 (mult conn)"
+      arguments: --variable httpVersion=3.0 --variable useHttps=true --variable http30EnableMultipleConnections=true --property httpversion=h3mulconn
```
CarnaViire, line 38:
```suggestion
      arguments: --variable httpVersion=3.0 --variable useHttps=true --variable http30EnableMultipleConnections=true --property httpversion=h3multconn
```

CarnaViire:
Changes LGTM -- have you tried to use the updated config YML to verify it works?

You can do it this way:

1) in `<local-path-to-repo>/scenarios/httpclient.benchmarks.yml`, change the pointers to the sources for the client job (`jobs.httpClient.source.repository` and `jobs.httpClient.source.branchOrCommit`) to point at your branch:

```yml
      repository: https://github.com/ManickaP/aspnet-benchmarks.git
      branchOrCommit: h3-mul-conns
```

2) use this local config file (instead of the remote one) in crank, and verify that both `true` and `false` values for the new parameter are passed and used correctly, e.g.:

```diff
-crank --config https://raw.githubusercontent.com/..../httpclient.benchmarks.yml --scenario ....
+crank --config <local-path-to-repo>/scenarios/httpclient.benchmarks.yml         --scenario .... --variable http30EnableMultipleConnections=false
```

#### HOW-TO: verify parameter is passed correctly

1) go to the job metadata on the agent (e.g. http://asp-citrine-lin:5001/jobs/100) and check the command line used to start the app, e.g.:

```json
{
"driverVersion": ....,
....
"arguments": "--address .... --http30EnableMultipleConnections false ....",
....
}
```

-AND/OR-

2) go to the job logs on the agent (e.g. http://asp-citrine-lin:5001/jobs/100/output) and check the options printed, e.g.

```
[10:10:50.344] HttpClient benchmark
[10:10:50.353] Options: Address=....; .... Http30EnableMultipleConnections=False; ....
....
```

CarnaViire (approved):
LGTM, thanks!

PR #2001, opened 2024-07-09
Force ServerGC in nativeaot scenarios

sebastienros:
FYI @MichalStrehovsky @eerhardt


sebastienros:
Not sure if the way we configure Workstation GC needs to be updated too (@mrsharm)

MichalStrehovsky:
I think all that was needed was to add `DOTNET_GCDynamicAdaptationMode=0` to the JIT-based legs. All the other additions look redundant.

PR #2010, opened 2024-08-01
Use built-in M.A.OpenApi package for AoT app

captainsafia:
Updates the test app for AoT to use the new built-in M.A.OpenAPI package and removes the OpenApiFeature.


src/BenchmarksApps/TodosApi/TodosApi.csproj
```diff
@@ -1,25 +1,23 @@
...
     <DefineConstants Condition=" '$(EnableLogging)' == 'true' ">$(DefineConstants);ENABLE_LOGGING</DefineConstants>
   </PropertyGroup>
   
   <ItemGroup>
-    <ProjectReference Include="..\AspNetCore.OpenApi\AspNetCore.OpenApi.csproj" />
     <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="$(MicrosoftAspNetCoreAppPackageVersion)" />
     <PackageReference Include="Npgsql" Version="$(NpgsqlVersion80)" />
     <PackageReference Include="Nanorm.Npgsql" Version="0.0.5" />
-    <PackageReference Include="Swashbuckle.AspNetCore" Version="6.5.0" />
+    <PackageReference Include="Microsoft.AspNetCore.OpenApi" Version="$(MicrosoftAspNetCoreAppPackageVersion)" />
```
captainsafia, line 20: @sebastienros This change will break the app if it runs on non-net9.0 targets because the `AddOpenApi` API that we consume is only introduced in 9.0 versions of the package. I see that `MicrosoftAspNetCoreAppPackageVersion` support multipls versions but it seems based on our discussions that we're only running on net9.0?

src/BenchmarksApps/TodosApi/TodosApi.json
```diff
@@ -45,7 +45,13 @@
         ],
         "operationId": "CreateTodo",
         "requestBody": {
-          "content": { },
+          "content": {
```
captainsafia, line 48: For whatever reason, request body schemas weren't populated in the previous version of the document.

src/BenchmarksApps/TodosApi/TodosApi.json
```diff
@@ -12,7 +12,7 @@
         ],
         "responses": {
           "200": {
-            "description": "Success"
+            "description": "OK"
```
captainsafia, line 15: Swashbuckle uses custom mappings for it's descriptions. We use the `ReasonPhrases` API.

src/BenchmarksApps/TodosApi/TodosApi.json
```diff
@@ -58,6 +64,16 @@
                 }
               }
             }
+          },
+          "201": {
```
captainsafia, line 68: Similarily, it looks like annotations from `IEndpointMetaataProvider` implementations weren't applied before.

src/BenchmarksApps/TodosApi/TodoApi.cs
```diff
@@ -94,6 +94,7 @@ UPDATE Todos
 }
 
 [JsonSerializable(typeof(Todo))]
+[JsonSerializable(typeof(bool?))]
```
captainsafia, line 97: We call into STJ to generate schemas for all types, including simple parameters bound directly from the route (like `bool? isCompleted`). There's some scenarios where this means that users need to add additional JsonSerializable attributes for types that haven't already been captured.

src/BenchmarksApps/TodosApi/TodosApi.json
```diff
@@ -142,15 +167,20 @@
             "name": "id",
             "in": "path",
             "required": true,
-            "style": "simple",
```
captainsafia, line 145: `"simple"` is the default style for path-parameters so there's no need to set it explicitly.

src/BenchmarksApps/TodosApi/TodosApi.json
```diff
@@ -201,23 +239,31 @@
             "name": "title",
             "in": "query",
             "required": true,
-            "style": "form",
```
captainsafia, line 204: `"form"` is the default style for parameters in the query so there's no need to set it explicitly.

src/BenchmarksApps/TodosApi/TodosApi.json
```diff
@@ -319,11 +377,9 @@
               "items": {
                 "type": "string"
               }
-            },
-            "nullable": true
```
captainsafia, line 323: `HttpValidationProblemDetails` can't actually be `null` when returned via the response.

sebastienros (approved):

Issue #2016, opened 2024-08-20
Possible seconds/milliseconds confusion in TcpClient


rgesteve:
Two of the three uses of `Task.Delay` (using `int` as argument) pass a value seemingly meant to be in seconds, but [docs](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.delay?view=net-8.0#system-threading-tasks-task-delay(system-int32)) says the arg is interpreted to be in *milliseconds*.   Should these values be scaled by 1,000 at some point in the code? 

The occurrences I mean are:
* https://github.com/aspnet/Benchmarks/blob/main/src/TcpClient/Program.cs#L93
* https://github.com/aspnet/Benchmarks/blob/main/src/TcpClient/Program.cs#L99

Apologies if this is an error of my reading of the docs/comments.

sebastienros:
You are correct. Either apply the factor, or better change the variables to be Timespans (calling it `Warmup` and `Duration`) and create the values correctly when initialized.

PR #2017, opened 2024-08-21
Change Duration and Warmup representation to reflect they're timestamps

rgesteve:
Begins addressing #2016 , following @sebastienros guidance (mostly, variable names still have "Seconds" suffix).  Also, not sure it follows style.  But as far as I can tell it works (I tested using .net 8).

Also: compiler (.net 8) complains about missing `await`, but I didn't make those changes since they don't seem related to the specific problem.


sebastienros:
thank you!

PR #2027, opened 2024-10-23
Razor Slices in Fortunes Middleware

BrennanConroy:
Massively reduces allocations in the Fortunes middleware by using `byte[]` instead of `string` for the `Message` property in the `Fortune` class and writing directly to the `PipeWriter` instead of to a `StringBuilder`. ~2.2GB/sec -> ~1.3GB/sec

Additionally, we've added templating which I believe makes it follow the guidelines for a proper Fortunes benchmark according to Techempower.

### DIFF
--- a/src/Benchmarks/Data/EfDb.cs
+++ b/src/Benchmarks/Data/EfDb.cs
@@ -83,7 +83,7 @@ namespace Benchmarks.Data
                 result.Add(fortune);
             }
 
-            result.Add(new Fortune { Message = "Additional fortune added at request time." });
+            result.Add(new Fortune { Message = "Additional fortune added at request time."u8.ToArray() });
             
             result.Sort();
 
--- a/src/Benchmarks/Data/EfDb.cs
+++ b/src/Benchmarks/Data/EfDb.cs
@@ -18,6 +18,8 @@ namespace Benchmarks.Data
         private readonly IRandom _random;
         private readonly ApplicationDbContext _dbContext;
 
+        private readonly byte[] AdditionalFortune = "Additional fortune added at request time."u8.ToArray();
+
         public EfDb(IRandom random, ApplicationDbContext dbContext, IOptions<AppSettings> appSettings)
         {
             _random = random;
@@ -83,7 +85,7 @@ namespace Benchmarks.Data
                 result.Add(fortune);
             }
 
-            result.Add(new Fortune { Message = "Additional fortune added at request time."u8.ToArray() });
+            result.Add(new Fortune { Message = AdditionalFortune });
             
             result.Sort();
 
--- a/src/Benchmarks/Data/Fortune.cs
+++ b/src/Benchmarks/Data/Fortune.cs
@@ -19,10 +19,10 @@ namespace Benchmarks.Data
         public int _Id { get; set; }
 
         [Column("message")]
-        [StringLength(2048)]
+        [MaxLength(2048)]
         [IgnoreDataMember]
         [Required]
-        public string Message { get; set; }
+        public byte[] Message { get; set; }
         
         public int CompareTo(object obj)
         {
@@ -31,7 +31,7 @@ namespace Benchmarks.Data
 
         public int CompareTo(Fortune other)
         {
-            return String.CompareOrdinal(Message, other.Message);
+            return Message.AsSpan().SequenceCompareTo(other.Message.AsSpan());
         }
     }
 }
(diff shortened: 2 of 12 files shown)


src/Benchmarks/Middleware/MiddlewareHelpers.cs
```diff
@@ -30,27 +29,15 @@ public static int GetMultipleQueriesQueryCount(HttpContext httpContext)
...
         }
 
-        public static Task RenderFortunesHtml(IEnumerable<Fortune> model, HttpContext httpContext, HtmlEncoder htmlEncoder)
+        public static async Task RenderFortunesHtml(IEnumerable<Fortune> model, HttpContext httpContext,
+            HtmlEncoder htmlEncoder, SliceFactory<IEnumerable<Fortune>> fortunesFactory)
         {
             httpContext.Response.StatusCode = StatusCodes.Status200OK;
             httpContext.Response.ContentType = "text/html; charset=UTF-8";
 
-            var sb = StringBuilderCache.Acquire();
```
sebastienros, line 38: Can we get rid of `StringBuilderCache` then? Haven't checked if anything else uses it.

src/Benchmarks/Data/Fortune.cs
```diff
@@ -19,10 +19,10 @@ public class Fortune : IComparable<Fortune>, IComparable
         public int _Id { get; set; }
 
         [Column("message")]
-        [StringLength(2048)]
+        [MaxLength(2048)]
         [IgnoreDataMember]
         [Required]
-        public string Message { get; set; }
+        public byte[] Message { get; set; }
```
sebastienros, line 25: Not sure that's valid in terms of requirements.

src/Benchmarks/Data/EfDb.cs
```diff
@@ -83,7 +83,7 @@ public async Task<IEnumerable<Fortune>> LoadFortunesRows()
                 result.Add(fortune);
             }
 
-            result.Add(new Fortune { Message = "Additional fortune added at request time." });
+            result.Add(new Fortune { Message = "Additional fortune added at request time."u8.ToArray() });
```
sebastienros, line 86: Does it reuse the array, or should it be declared as a class member.

src/Benchmarks/Data/Fortune.cs
DamianEdwards: It is. Nothing wrong with UTF8 encoded bytes for the message property.

src/Benchmarks/Middleware/MiddlewareHelpers.cs
BrennanConroy: https://github.com/aspnet/Benchmarks/blob/05ad4582116b62443cefe5bef4086bfe159d02a2/src/Benchmarks/Data/BatchUpdateString.cs#L28

src/Benchmarks/Data/EfDb.cs
BrennanConroy: > Within the scope of the request, a new Fortune object must be constructed and added to the list. This confirms that the data structure is dynamic-sized. The new fortune is not persisted to the database; it is ephemeral for the scope of the request.

It's a little vague on the details, but I feel like it should be created per request.

DamianEdwards: Can make it a member like https://github.com/aspnet/Benchmarks/blob/05ad4582116b62443cefe5bef4086bfe159d02a2/src/BenchmarksApps/TechEmpower/PlatformBenchmarks/Data/RawDb.cs#L291

BrennanConroy: I guess a hardcoded string isn't created per request either.

sebastienros:
Just to have an idea of how much different this is

| Method    | Mean      | Error     | StdDev    | Gen0   | Allocated |
|---------- |----------:|----------:|----------:|-------:|----------:|
| ToArray   | 4.4749 ns | 0.5935 ns | 0.0325 ns | 0.0077 |      72 B |
| NoToArray | 0.2429 ns | 0.3917 ns | 0.0215 ns |      - |         - |
| Static    | 0.0000 ns | 0.0000 ns | 0.0000 ns |      - |         - |

```c#
public class Benchmarks
{
    private static readonly byte[] _message = "Additional fortune added at request time."u8.ToArray();

    [Benchmark]
    public byte[] ToArray()
    {
        var message = "Additional fortune added at request time."u8.ToArray();
        return message;
    }

    [Benchmark]
    public ReadOnlySpan<byte> NoToArray()
    {
        var message = "Additional fortune added at request time."u8;
        return message;
    }

    [Benchmark]
    public ReadOnlySpan<byte> Static()
    {
        return _message;
    }
}
```

DamianEdwards (approved):
Note that the pattern Razor Slices uses in the latest version is quite different, with no lookup of template via a string path and caching of a delegate anymore. It's all done at build time via a source generator and you get types to represent each template. But this matches what we're doing in the other benchmarks we use Razor Slices in right now.

PR #2044, opened 2025-01-14
fix: integrate mutual TLS scenario with underlying http.sys server 

DeagleGross:
mTLS (mutual TLS) requires a separate setup from the server side. mTLS is valid, when clientCert is negotiated on the connection establishment, meaning when ASP.NET code calls `context.Connection.ClientCertificate`: it should be already available and there must not be any need to invoke `context.Connection.GetClientCertificateAsync()`

in case of Kestrel setup is extremely easy - just set a specific enum value:
```csharp
options.ClientCertificateMode = ClientCertificateMode.RequireCertificate;
options.ClientCertificateValidation = AllowAnyCertificateValidationWithLogging;
```
however in case of HTTP.SYS the only option is to execute a [netsh command](https://learn.microsoft.com/en-us/windows-server/networking/technologies/netsh/netsh-http#add-sslcert) `netsh http add sslcert`.

I have added a code in ASP.NET application, which on start binds the test certificate using netsh command, and then using an applicationLifetime `ApplicationStopping` callback it deletes the cert binding from the machine. It can be seem from the logs (below).

Basically the order is the following:
1) bind the test cert to http.sys with `clientcertnegotiation=enable` to enable client cert negotiation
_optionally) `netsh http show sslcert` can show the ssl cert bindings on a machine. Important property is "Negotiate Client Certificate : Enabled"_
2) execute the benchmark (in other words process the traffic). If verbose logging of cert validation is enabled, you should see message "client certificate (...) already exists on the connection", meaning that client cert negotiation happened on the connection establishment
3) in the end of the benchmark run during a graceful shutdown of an app a rollback of http.sys configuration will happen - we will execute `netsh http delete sslcert`

<blockquote>
args: --urls https://10.0.0.110:5000 --mTLS true --certValidationConsoleEnabled true --statsEnabled false --tlsRenegotiation true --httpSysLogs true

Setting up mTLS for http.sys
Loaded `testCert.pfx` from local file system
Added `testCert.pfx` to localMachine cert store
Executing command: `netsh http add sslcert ipport=10.0.0.110:5000 certstorename=MY certhash=6A283FAAA7FFFC129DF24CC0E7521186C7926219 appid={75e4f5b6-776f-448f-8a36-64e964bf2147} clientcertnegotiation=enable`
Configured http.sys settings for mTLS
Registered client cert validation middleware
Executing command: `netsh http show sslcert`

SSL Certificate bindings: 
------------------------- 
    IP:port                      : 10.0.0.110:5000
    Certificate Hash             : 6a283faaa7fffc129df24cc0e7521186c7926219
    Application ID               : {75e4f5b6-776f-448f-8a36-64e964bf2147}
    Certificate Store Name       : MY
    Verify Client Certificate Revocation : Enabled
    Verify Revocation Using Cached Client Certificate Only : Disabled
    Usage Check                  : Enabled
    Revocation Freshness Time    : 0
    URL Retrieval Timeout        : 0
    Ctl Identifier               : (null)
    Ctl Store Name               : (null)
    DS Mapper Usage              : Disabled
    **Negotiate Client Certificate : Enabled**
    Reject Connections           : Disabled
    Disable HTTP2                : Not Set
    Disable QUIC                 : Not Set
    Disable TLS1.2               : Not Set
    Disable TLS1.3               : Not Set
    Disable OCSP Stapling        : Not Set
    Enable Token Binding         : Not Set
    Log Extended Events          : Not Set
    Disable Legacy TLS Versions  : Not Set
    Enable Session Ticket        : Not Set

Application Info:
---
	mTLS is enabled (client cert is required)
	enabled logging certificate validation events to console
	listening endpoints: https://10.0.0.110:5000

Application started.
client certificate (6A283FAAA7FFFC129DF24CC0E7521186C7926219) already exists on the connection -648518319766231673
client certificate (6A283FAAA7FFFC129DF24CC0E7521186C7926219) already exists on the connection -216172777013495157
client certificate (6A283FAAA7FFFC129DF24CC0E7521186C7926219) already exists on the connection -144115161500731759
client certificate (6A283FAAA7FFFC129DF24CC0E7521186C7926219) already exists on the connection -576460747203134805
client certificate (6A283FAAA7FFFC129DF24CC0E7521186C7926219) already exists on the connection -432345559127283053
...
...
...
Disabling mTLS for http.sys
Executing command: `netsh http delete sslcert ipport=10.0.0.110:5000`
Disabled http.sys settings for mTLS
</blockquote>

### DIFF
--- a/src/BenchmarksApps/TLS/HttpSys/NetShWrapper.cs
+++ b/src/BenchmarksApps/TLS/HttpSys/NetShWrapper.cs
@@ -5,6 +5,18 @@ namespace HttpSys
 {
     public static class NetShWrapper
     {
+        public static void DisableHttpSysMutualTlsIfExists(string ipPort)
+        {
+            try
+            {
+                DisableHttpSysMutualTls(ipPort);
+            }
+            catch
+            {
+                // ignore
+            }
+        }
+
         public static void DisableHttpSysMutualTls(string ipPort)
         {
             Console.WriteLine("Disabling mTLS for http.sys");
@@ -70,9 +82,10 @@ namespace HttpSys
         }
 
 #pragma warning disable SYSLIB0057 // Type or member is obsolete
-        private static X509Certificate2 LoadCertificate() => File.Exists("testCert.pfx")
-            ? new X509Certificate2("testCert.pfx", "testPassword", X509KeyStorageFlags.MachineKeySet | X509KeyStorageFlags.Exportable)
-            : new X509Certificate2("../testCert.pfx", "testPassword", X509KeyStorageFlags.MachineKeySet | X509KeyStorageFlags.Exportable);
+        private static X509Certificate2 LoadCertificate()
+            => File.Exists("testCert.pfx")
+            ? X509CertificateLoader.LoadPkcs12FromFile("testCert.pfx", "testPassword", X509KeyStorageFlags.MachineKeySet | X509KeyStorageFlags.Exportable)
+            : X509CertificateLoader.LoadPkcs12FromFile("../testCert.pfx", "testPassword", X509KeyStorageFlags.MachineKeySet | X509KeyStorageFlags.Exportable)
 #pragma warning restore SYSLIB0057 // Type or member is obsolete
     }
 }
--- a/src/BenchmarksApps/TLS/HttpSys/NetShWrapper.cs
+++ b/src/BenchmarksApps/TLS/HttpSys/NetShWrapper.cs
@@ -1,0 +1,73 @@
+﻿using System.Diagnostics;
+using System.Security.Cryptography.X509Certificates;
+
+namespace HttpSys
+{
+    public static class NetShWrapper
+    {
+        public static void DisableHttpSysMutualTls(string ipPort)
+        {
+            Console.WriteLine("Disabling mTLS for http.sys");
+
+            string command = $"http delete sslcert ipport={ipPort}";
+            ExecuteNetShCommand(command);
+
+            Console.WriteLine("Disabled http.sys settings for mTLS");
+        }
+
+        public static bool HttpSysCertBindingExists(string ipPort)
+        {
+            return true;
+        }
+
+        public static void EnableHttpSysMutualTls(string ipPort)
+        {
+            Console.WriteLine("Setting up mTLS for http.sys");
+
+            var certificate = LoadCertificate();
+            Console.WriteLine("Loaded `testCert.pfx` from local file system");
+            using (var store = new X509Store(StoreName.My, StoreLocation.LocalMachine))
+            {
+                store.Open(OpenFlags.ReadWrite);
+                store.Add(certificate);
+                Console.WriteLine("Added `testCert.pfx` to localMachine cert store");
+                store.Close();
+            }
+
+            string certThumbprint = certificate.Thumbprint;
+            string appId = Guid.NewGuid().ToString();
+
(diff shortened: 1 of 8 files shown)


src/BenchmarksApps/TLS/HttpSys/NetShWrapper.cs
```diff
@@ -0,0 +1,78 @@
...
+                Console.WriteLine(output);
+            }
+
+            if (process.ExitCode != 0)
+            {
+                throw new InvalidOperationException($"netsh command execution failure: {output}");
+            }
+        }
+
+#pragma warning disable SYSLIB0057 // Type or member is obsolete
```
BrennanConroy, line 72: Can we use the non-obsolete APIs instead?
https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0057#workaround

src/BenchmarksApps/TLS/HttpSys/Program.cs
```diff
@@ -40,6 +43,37 @@
 }
 
 if (mTlsEnabled)
+{
+    var hostAppLifetime = app.Services.GetService<IHostApplicationLifetime>();
+    hostAppLifetime!.ApplicationStopping.Register(OnShutdown);
```
BrennanConroy, line 48:
```suggestion
    var hostAppLifetime = app.Services.GetRequiredService<IHostApplicationLifetime>();
    hostAppLifetime.ApplicationStopping.Register(OnShutdown);
```

BrennanConroy (approved):
Nice, I am slightly worried if netsh fails to delete the configuration and leaves the machine in a setup state.

src/BenchmarksApps/TLS/HttpSys/NetShWrapper.cs
DeagleGross: I tried those APIs, but they are not allowing loading `pfx` file. I.e. [X509CertificateLoader.LoadCertificateFromFile](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificateloader.loadcertificatefromfile?view=net-9.0) loads a single X.509 certificate (in either the **PEM** or **DER** encoding) from the specified file (**not PFX**).

do u know if there is a non-obsolete API which allows it?

DeagleGross:
> Nice, I am slightly worried if netsh fails to delete the configuration and leaves the machine in a setup state.

I am as well. I am thinking about 2 approaches:
1) leave it as is relying on re-running benchmark again eventually on same machine, which will bring machine back to a proper setup. I need to make an additional verification check in the startup then
2) add the `after-script` to perform sslcert binding deletion.

i will try to do №2 a bit later today - it should be more robust

src/BenchmarksApps/TLS/HttpSys/NetShWrapper.cs
BrennanConroy: https://learn.microsoft.com/dotnet/api/system.security.cryptography.x509certificates.x509certificateloader.loadpkcs12?view=net-9.0 says it loads a pfx. If that doesn't work we can ask someone from runtime.

BrennanConroy: Should be able to remove pragmas now?

DeagleGross: changed to new API, removed pragmas, thanks for suggestion!

DeagleGross:
decided to change the port to 8080 so that other tests are not impacted by http.sys configuration changes (if there would be) - just for safety.

PR #2049, opened 2025-01-20
feat: control SSL protocol version on tls jobs

DeagleGross:
Adding an ability to specify the tls version for the job.

In case of Kestrel, added an option to do it on server side. In case of HTTP.SYS the only available option is to modify the underlying registry (which is unsafe and can fail / leave machine in a transient state). So the easiest is to explicitly set the `SslProtocols` property on the client causing server to establish a connection using specified tls protocol version.

the option `load.httpClient.variables.sslProtocols` is being contributed here: https://github.com/dotnet/crank/pull/786 (current PR should be merged after the crank one)


src/BenchmarksApps/TLS/Kestrel/Program.cs
```diff
@@ -81,6 +94,27 @@ bool AllowAnyCertificateValidationWithLogging(X509Certificate2 certificate, X509
...
+            logged = true;
+
+            var tlsHandshakeFeature = context.Features.GetRequiredFeature<ITlsHandshakeFeature>();
+
+            Console.WriteLine("Request details:");
+            Console.WriteLine("-----");
+            Console.WriteLine("TLS: " + tlsHandshakeFeature.Protocol);
+            Console.WriteLine("-----");
+        }
+        await next();
```
BrennanConroy, line 114: nit: `await next(context);`

src/BenchmarksApps/TLS/Kestrel/Program.cs
```diff
@@ -157,4 +192,35 @@ static IPEndPoint CreateIPEndPoint(UrlPrefix urlPrefix)
...
+    {
+        return null;
+    }
+
+    foreach (var version in supportedTlsVersions.Split(','))
+    {
+        switch (version.Trim().ToLower())
+        {
+#pragma warning disable SYSLIB0039 // Type or member is obsolete
+            case "tls11":
```
BrennanConroy, line 210: Remove? I don't think we will want to do anything with TLS1.1

BrennanConroy (commented):
Are we going to add TLS1.3?
We should probably change the scenario names so tls1.2 and tls1.3 are distinguishable.

DeagleGross:
> Are we going to add TLS1.3?
I think we should not add more jobs, but simply run TLS1.3 when we would want to compare the perf. It feels like we are adding too many jobs at a moment if we would want to multiply by TLS / OpenSSL version (x3 jobs).

what do you think?

src/BenchmarksApps/TLS/Kestrel/Program.cs
DeagleGross: applied

src/BenchmarksApps/TLS/Kestrel/Program.cs
DeagleGross: agree, removed

src/BenchmarksApps/TLS/HttpSys/Program.cs
```diff
@@ -30,6 +36,27 @@
...
+            logged = true;
+
+            var tlsHandshakeFeature = context.Features.GetRequiredFeature<ITlsHandshakeFeature>();
+
+            Console.WriteLine("Request details:");
+            Console.WriteLine("-----");
+            Console.WriteLine("TLS: " + tlsHandshakeFeature.Protocol);
+            Console.WriteLine("-----");
+        }
+        await next();
```
BrennanConroy, line 56: Same 😆

BrennanConroy (approved):

PR #2050, opened 2025-01-23
feat: docker-based benchmarks comparing OpenSSL perf

DeagleGross:
As we have heard OpenSSL 3.x.x has a perf degradation comparing OpenSSL 1.1.1.
It is critical to understand if there is a bottleneck in the OpenSSL package itself, that is why I packed app used in previously set up benchmarks into docker containers. Now we would be able to compare performance differing the OpenSSL package version easily.

example:
> crank --config https://raw.githubusercontent.com/DeagleGross/Benchmarks/refs/heads/dmkorolev/idna/openssl-versioning/scenarios/tls.benchmarks.yml --scenario tls-handshakes-docker-openssl-332 --profile aspnet-perf-lin --application.framework net10.0

results that I got:
| OpenSSL version | RPS |
| --- | --- |
| 3.3.2 (latest) | 2,290 |
| 1.1.1 | 2,305 |
| 3.0.15 | 1,430 |


src/BenchmarksApps/TLS/Kestrel/Dockerfile
```diff
@@ -0,0 +1,38 @@
+# This stage is used when running from VS in fast mode (Default for Debug configuration)
+FROM mcr.microsoft.com/dotnet/aspnet:9.0-alpine AS base
```
DeagleGross, line 2: base on azure mariner base image (linux vms)

DeagleGross: discussed offline; should not expand here and we will create a separate job to run based on azurelinux image

src/BenchmarksApps/TLS/Kestrel/Program.cs
```diff
@@ -219,4 +223,25 @@ static IPEndPoint CreateIPEndPoint(UrlPrefix urlPrefix)
     }
 
     return protocols;
+}
+
+static void LogOpenSSLVersion()
```
BrennanConroy, line 228: Looks like this code runs no matter what OS? Should probably check you're on linux?

BrennanConroy (approved):

src/BenchmarksApps/TLS/Kestrel/Program.cs
DeagleGross: added check that we are on linux or macos

PR #2058, opened 2025-02-05
Add intel-lin-relay

DrewScoggins:


sebastienros (approved):
You will need to update the host file on the azdo-workers (db) machines if they have to be used in CI.

sebastienros:
@DeagleGross I knew I had seen it somewhere

DeagleGross (approved):

PR #2080, opened 2025-03-31
feat: use different public-key length certs for TLS

DeagleGross:
I've noticed, that certificate length heavily contributes to the TLS speed (4096 bit vs 2048 bit). I have seen http.sys using 2048 bit length cert, and kestrel using a [testCert.pfx](https://github.com/aspnet/Benchmarks/blob/main/src/BenchmarksApps/Mvc/testCert.pfx) which is 4096 bit length, meaning tests are unfair.

Besides updating certificates on the machines, I prepared a separate nginx test (with a possibility to specify a cert length) and updated http.sys + kestrel benchmarks to have such a parameter (http.sys now issues a cert before each run and kestrel uses pre-created certs)

Here are the RPS values for some random run on a server using different cert lengths
| Cert length \ Server | Nginx | Kestrel | Http.Sys |
| -- | -- | -- | -- |
| 2048 | 2,167 | 1436 | 1250-1450 (depending on EC curve) |
| 4096 | 491 | 443 | 439 |


sebastienros:
fyi we already have some docker file and config and certs using nginx. At least if there is a reason to create a new one then some things should be reused: https://github.com/aspnet/Benchmarks/blob/main/docker/nginx/nginx-https.conf

PR #2081, opened 2025-04-02
Use List<T> instead of Array for Json object

BrennanConroy:
This will use the List converter which avoids STJ internally creating a list just to call ToArray to give it back to us.

Allocation rate went from ~2.6B/sec to ~2.4B/sec


sebastienros:
Do you mean that in STJ deserializing an array will always use an intermediate list?

sebastienros (approved):

BrennanConroy:
> Do you mean that in STJ deserializing an array will always use an intermediate list?

I won't claim "always". But using the debugger I followed it into the list creation and then ToArraying the list.

PR #2084, opened 2025-04-04
Suppress credscan false positives in tls.benchmarks.yml

akoeplinger:
Noticed this while looking at the code mirror, should unbreak the mirroring.
/cc @DeagleGross @sebastienros

**Important:** make sure the commit message contains this string:

```
**BYPASS_SECRET_SCANNING**
```


sebastienros:
Thanks @akoeplinger

PR #2088, opened 2025-05-07
Add Cobalt Cloud machine Azure CI profile and runs.

LoopedBard3:
Add Cobalt Cloud machine Azure CI profile and runs. We have 3 added Linux Cobalt machines, so I decided to add them to each of the run types. 

One thing that I am unsure about it that these machines are under a different virtual network as the cobalt machines were not available in the same data center as the other machines. I am not super familiar with the connection between the pipelines and the machines so please let me know if there is known work that will have to be done there.

Additional steps necessary to set this up:
- [x] Added cobaltcloud to the aspnetbenchmarks service bus (I am assuming service bus can be accessed across datacenters).
- [x] Added necessary environment variables to the cobalt-db-lin machine to act as the entry point for that network.
- [x] Setup the service bus cronjob and start the connection.
- [x] Did not need to add hosts file as the other VMs are pingable via their VM name.

### DIFF
--- a/build/azure.profile.yml
+++ b/build/azure.profile.yml
@@ -6,21 +6,21 @@ profiles:
       serverPort: 5000
       serverAddress: 10.0.4.4
       cores: 4
-    jobs: 
+    jobs:
       db:
-        endpoints: 
+        endpoints:
           - http://10.0.4.6:5001
         aliases:
           - extra
       application:
-        endpoints: 
+        endpoints:
           - http://10.0.4.4:5001
         variables:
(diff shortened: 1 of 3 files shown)


LoopedBard3:
For now, I added a new service bus queue for the cobalt machines as they are in a different datacenter and VNet from the others. While this currently works with the current queue setup, it would probably be best long-term to update the benchmarks.template.liquid template. Not sure what the best approach would be, but I would be open to recommendations if we want them in this PR.

sebastienros (approved):

PR #2097, opened 2025-06-17
Added MVC jobs for cobalt cloud runs.

LoopedBard3:
Added Mvc jobs for cobalt cloud runs. Also fixed some minor formatting for the container jobs definition in the matrix. I tested this from the AL3 machine using the about-sqlite scenario successfully.


DrewScoggins (approved):
LGTM

PR #2099, opened 2025-06-26
Update the benchmarks-ci-*.yml files and add supporting file for WIP scheduling flow.

LoopedBard3:
Update the benchmarks-ci-*.yml files with current target runs and add supporting file for WIP scheduling flow. The new flow is still WIP but the schedule updates are important to help ensure all expected runs are working. The new flow will consist of updating the new .json file with the runs we want and running a python program that will generate the schedule to run and generate the resulting yaml pipeline files.

## Unclear benchmark name changes:
Amd Linux -> Amd Linux2
Fixed Intel Gold Linux -> Gold Linux

### Arm 28/80 to be replaced by Cobalt
Baselines Arm 28 Linux
Baselines Database Arm 28 Linux
Crossgen Arm 28 Linux
Frameworks Arm 28 Linux
Frameworks Database Arm 28 Linux
GC Arm 28 Linux
Native Aot Arm 28 Linux
PGO Arm 28 Linux
Trends Arm 28 Linux
Trends Database Arm 28 Linux
Frameworks Arm 80 Linux
Frameworks Database Arm 80 Linux
Trends Arm 80 Linux
Trends Database Arm 80 Linux

### Siryn removed due to not being used/monitored
Baselines Database Siryn Linux
Baselines Siryn Linux
Mvc Siryn Linux
Trends Database Siryn Linux
Trends Siryn Linux

## Added Jobs (Goal of matching the currently running intel jobs)
Blazor Gold Linux
Build Gold Linux
Build Gold Windows
Containers Gold Linux
Crossgen Gold Linux
Crossgen Gold Windows
Custom Proxies Gold Linux
EF Core Gold Windows
GC Gold Linux
GC Gold Windows
Grpc Gold Linux
Grpc Gold Windows
HttpClient Gold Linux
MVC Gold Linux
MVC Gold Windows
NativeAOT Gold Linux
NativeAOT Gold Windows
PGO Gold Linux
PGO Gold Windows
Proxies Gold Linux
SignalR Gold Linux
SignalR Gold Windows
Single File Gold Linux
Single File Gold Windows
SslStream Gold Linux
WebSockets Gold Linux
WebSockets Gold Windows

PR #2101, opened 2025-07-16
Fix NativeAOT runs to include db machines.

LoopedBard3:
Fix NativeAOT runs to include db machines.

This was inadvertently incorrectly set in the initial PR switching to the new test addition flow: https://github.com/aspnet/Benchmarks/pull/2099.


DrewScoggins (approved):
LGTM

PR #2104, opened 2025-07-31
Switch to auto-scheduled setup for azure and azure_eastus2 ci runs

LoopedBard3:
Switch to auto-scheduled setup for azure and azure_eastus2 ci runs. This includes adding the .json configuration files for specifying the jobs to schedule, updating the yaml schedules themselves, and adding individual machine profiles for the azure machines. I have one more set of machines and jobs to add and then will finalize and add the scheduler I have been using to the repo.

Can't test run the pipelines as the profiles are needed but will make sure to watch the pipelines for success.

### DIFF
--- a/build/azure.profile.yml
+++ b/build/azure.profile.yml
@@ -400,3 +400,498 @@ profiles:
         aliases:
           - warmup
           - secondary
+
+  # Add individual machine profiles for azure-arm64-db similar to ci.profile.yml
+  azure-arm64-db-app:
+    variables:
+      serverAddress: 10.0.4.9
+      cores: TODO
+    agents:
+      application:
+        endpoints:
+          - http://10.0.4.9:5001
+        aliases:
+          - main
+
+  azure-arm64-db-db:
+    variables:
+      databaseServer: 10.0.4.9
+      downstreamAddress: 10.0.4.9
+    agents:
+      db:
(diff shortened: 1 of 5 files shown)


LoopedBard3:
Force-pushed to include the clean-up of non-required configuration options in the azure.yml file. (preferred_partners and default_profile are not required).

build/azure.profile.yml
```diff
@@ -399,4 +399,513 @@ profiles:
...
+      serverAddress: 10.0.4.9
+      cores: 4
+    agents:
+      application:
+        endpoints:
+          - http://10.0.4.9:5001
+        aliases:
+          - main
+
+  azure-db-db:
```
DrewScoggins, line 417: This and `azure-db-app` have the same IP address, is that correct?

build/azure.profile.yml
```diff
@@ -399,4 +399,513 @@ profiles:
...
+      downstreamAddress: 10.0.4.9
+    agents:
+      db:
+        endpoints:
+          - http://10.0.4.9:5001
+        aliases:
+          - downstream
+          - extra
+
+  azure-db-load:
```
DrewScoggins, line 429: Same as above.

DrewScoggins (approved):
LGTM

PR #2105, opened 2025-07-31
Add hosted cobalt crank profiles and runs

LoopedBard3:
Add hosted cobalt crank profiles and runs. This includes adding the cobalt hosted machine setups to the ci.setup.yml, aspnet.profile.yml, and aspnet.standard.profile.yml files, adding the hosted cobalt ci configuration file, and adding the hosted cobalt ci run itself. 

If we think including these with the already running benchmarks_ci runs makes more sense, we can do that. I will just be interested in ideas on what we think a good practice will be for ensuring the local vnet machine tests don't try to use the cobalt machines.

This also brings back testing for the ARM (80 and 26 core) scenarios removed here: 
https://github.com/aspnet/Benchmarks/pull/2099

A new pipeline will need to be created to run these.

### DIFF
--- a/build/benchmarks-ci-cobalt.yml
+++ b/build/benchmarks-ci-cobalt.yml
@@ -45,29 +45,87 @@ jobs:
       serviceBusNamespace: aspnetbenchmarks
       arguments: "$(ciProfile) --profile cobalt-hosted-lin-server-app --profile cobalt-hosted-lin-client-load --profile cobalt-hosted-lin-db-db "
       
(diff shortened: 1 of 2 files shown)


LoopedBard3:
> Does hosted mean these are the machines that are VMs that we have created on our physical hardware?

Yes.

DrewScoggins (approved):
LGTM

DrewScoggins (approved):
LGTM

PR #2136, opened 2025-12-04
Fix nuget restore in prbenchmarks.yml and regressionbot.yml

akoeplinger:
In the pipeline dotnet restore was accidentally not using the nuget.config in the repo since we ran dotnet from the checkout root. Instead we can set `workspaceRoot: true` on the crank checkout step to make it the default working directory.

This fixes network isolation violations in the pipelines.

/cc @LoopedBard3 @sebastienros

### DIFF
--- a/build/prbenchmarks.yml
+++ b/build/prbenchmarks.yml
@@ -73,9 +73,10 @@ extends:
         steps:
         - checkout: dotnet-crank
           path: crank
+          workspaceRepo: true
           fetchDepth: 1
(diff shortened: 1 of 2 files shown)


akoeplinger:
I ran an internal test build in https://dev.azure.com/dnceng/internal/_build/results?buildId=2853658&view=results

akoeplinger:
thanks. the OSS portal is down right now, will do once it's back up.

PR #2139, opened 2026-01-12
Temporarily disable the SslStream benchmark runs

LoopedBard3:
Temporarily disable the SslStream benchmark runs as the app does not seem to be exiting properly, causing follow-up test runs to timeout.


LoopedBard3:
Just pushed an updated, let me know if that is more in-line with the approach you mentioned.

PR #2150, opened 2026-02-09
Fix tar-fs CVE by overriding to 3.1.1

DrewScoggins:
Add tar-fs 3.1.1 override in package.json to address path traversal vulnerability (CVE) in tar-fs < 3.1.1, a transitive dependency via puppeteer/@puppeteer/browsers.


sebastienros:
Note I don't think the benchmark has run in more than a year /cc @MackinnonBuck

sebastienros (approved):

PR #2163, opened 2026-03-23
Add net11.0 framework to gRPC proxy load jobs

LoopedBard3:
The grpc-dotnet master branch now multi-targets net10.0 (v2.76.0), causing NETSDK1045 errors on benchmark agents with only SDK 9.0. Add --load.framework net11.0 to all gRPC proxy scenarios (YARP, NGinx, Envoy) so crank installs an SDK that supports the net10.0 transitive dependencies.

This is consistent with how grpc-scenarios.yml already handles this.

### DIFF
--- a/build/proxies-scenarios.yml
+++ b/build/proxies-scenarios.yml
@@ -31,7 +31,7 @@ parameters:
     condition: 'true'
   - displayName: YARP gRPC
     scenario: proxy-yarp-grpc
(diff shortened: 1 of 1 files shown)


sebastienros (approved):

LoopedBard3:
Tested proxy-yarp-grpc and the --load.framework net11.0 did fix the issue, so merging.

PR #2168, opened 2026-05-04
Rewrite framework dependency diff URLs to dotnet/dotnet (VMR)

LoopedBard3:
## Why

The `RegressionBot` issues filed under `dotnet/aspnetcore` include a per-dependency Changes table whose **diff links 404** for `Microsoft.AspNetCore.App` and `Microsoft.NETCore.App`. Concrete example: [dotnet/aspnetcore#66568](https://github.com/dotnet/aspnetcore/issues/66568) shows three rows that share the same VMR commit hashes (`0188adf...4c4e7f4`):

| Dependency | Linked compare URL | Works? |
|------------|-------------------|--------|
| `Microsoft.AspNetCore.Antiforgery.dll` | `dotnet/dotnet/compare/0188adf...4c4e7f4` | ✅ (already correct from VMR-stamped `[AssemblyMetadata]`) |
| `Microsoft.AspNetCore.App` | `dotnet/aspnetcore/compare/0188adf...4c4e7f4` | ❌ |
| `Microsoft.NETCore.App` | `dotnet/runtime/compare/0188adf...4c4e7f4` | ❌ |

After the move to the VMR (`dotnet/dotnet`), the .NET runtime and ASP.NET runtime are built from `dotnet/dotnet`, and the commit hashes Crank captures are dotnet/dotnet commits. The Crank agent (`Microsoft.Crank.Agent/Startup.cs:3263, 3296`) still hardcodes the legacy component-repo URLs (`dotnet/aspnetcore`, `dotnet/runtime`) for the synthetic `Microsoft.AspNetCore.App` and `Microsoft.NETCore.App` entries, so the rendered compare URLs are wrong.

## What

Rewrite the URL at template-render time inside the **Changes** block of all five issue-body templates:

* `build/regressions.config.yml` — `rps`, `published-size`, `start-time`
* `build/regressions.blazor.config.yml` — `download-size`, `first-ui`

The rewrite triggers when **either**:

* the dependency name is one of `Microsoft.AspNetCore.App` / `Microsoft.NETCore.App`, **or**
* the stored `RepositoryUrl` is one of the legacy component-repo URLs (`https://github.com/dotnet/aspnetcore`, `https://github.com/dotnet/runtime`).

Other dependencies (e.g. application-owned assemblies like `Antiforgery.dll`, or already-correct `dotnet/dotnet` entries from post-VMR assembly metadata) are left untouched.

```liquid
{%- assign repoUrl = diff.RepositoryUrl -%}
{%- if diff.Names contains 'Microsoft.AspNetCore.App'
     or diff.Names contains 'Microsoft.NETCore.App'
     or repoUrl == 'https://github.com/dotnet/aspnetcore'
     or repoUrl == 'https://github.com/dotnet/runtime' -%}
    {%- assign repoUrl = 'https://github.com/dotnet/dotnet' -%}
{%- endif -%}
{{ repoUrl }}/compare/{{ diff.PreviousCommitHash }}...{{ diff.CurrentCommitHash }}
```

## Validation

* Both YAML configs parse (`yaml.safe_load`).
* All affected templates parse and render through `FluidParser` against a synthetic 5-row regression covering: an application assembly with its own repo URL, an already-correct `dotnet/dotnet` assembly-metadata entry, the two synthetic framework deps with the wrong hardcoded URLs, and a hypothetical "other dep stamped with one of the legacy URLs" case.
* Rendered output for each of the 5 affected templates: **0** leftover `dotnet/aspnetcore` / `dotnet/runtime` URLs, **4** correctly-attributed `dotnet/dotnet` URLs, **1** preserved `aspnet/Benchmarks` URL on the application assembly.

## Scope / non-goals

* **Heals already-stored SQL data** at render time, so when the bot updates an existing issue (e.g. to mark it recovered), the regenerated body is corrected too.
* Doesn't fix the duplicate-runtime-assembly bug (`Microsoft.AspNetCore.Antiforgery.dll` showing up alongside the rolled-up `Microsoft.AspNetCore.App`) — that bug lives in the Crank agent's `IsAspNetCoreDependency` / `IsNetCoreDependency` predicates and is intentionally left for a future Crank-side change.
* A Crank-side fix is **deferred** because there's in-flight work to add support for running benchmarks against individual ASP.NET Core builds (with the current VMR-sourced flow becoming an option). Once that lands, the right `RepositoryUrl` for `Microsoft.AspNetCore.App` / `Microsoft.NETCore.App` will become conditional on the build source rather than a hard flip — best to do that work together.

## Rollout

The next scheduled run of `build/regressionbot.yml` (07:00/19:00 UTC) picks this up automatically — the pipeline rebuilds the bot from `dotnet-crank@main` and reads these YAMLs from the `self` checkout each run.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


DrewScoggins (approved):
LGTM. Assuming these are the only two templates that we need to update.

LoopedBard3:
Yup, these were the only ones I noticed to be updated.

PR #2169, opened 2026-05-05
Remove EOL'd citrine/perf relays + VMs from pod scheduler and profiles

LoopedBard3:
The following Service Bus relays (and their underlying VMs) were end-of-lifed and removed from the `aspnetrelay` connection list in Azure:

- `perflin` (`asp-perf-lin`, pod `intel-perflin`)
- `perfwin` (`asp-perf-win`)
- `citrinewin` (`asp-citrine-win`, pod `intel-win`)
- `citrinelin` (`asp-citrine-lin`, pod `intel-lin`)
- `citrineamd2` (`asp-citrine-amd2`, pod `amd-lin2`)

## Changes

**`build/benchmarks_ci_pods.json`** - drop the four pods whose SUT is EOL'd (`intel-lin`, `intel-win`, `intel-perflin`, `amd-lin2`). Every scenario still has at least one valid pod (`gold-lin` / `gold-win`), so no scenarios are removed; their `pods` arrays are pruned accordingly.

**`build/ci.profile.yml`** - remove EOL'd profile defs (`intel-lin-*`, `intel-win-*`, `intel-perflin-app`, `amd-lin2-*`).

**`build/benchmarks-ci-01.yml`, `build/benchmarks-ci-02.yml`** - regenerated via:

```
python scripts/pod-scheduler/main.py --config build/benchmarks_ci_pods.json --base-name benchmarks-ci --yaml-output build
```

**`scenarios/aspnet.profiles.yml`, `scenarios/aspnet.profiles.standard.yml`**
- Remove profiles whose APP endpoint is EOL'd: `aspnet-citrine-{lin,win,amd2}[-relay]`, `aspnet-citrine-amd-relay`, `aspnet-perf-{lin,win}[-relay]`.
- Repoint load/db endpoints in profiles whose SUT still works but whose load/db lived on `citrineamd2` / `asp-citrine-amd2`:

  | Profile | Was | Now |
  |---|---|---|
  | `aspnet-citrine-arm-lin[-relay]` | load = `citrineamd2` | `citrineload` |
  | `aspnet-citrine-arm-win[-relay]` *(yml only)* | db = `citrineamd2` | `citrinedb` |
  | `aspnet-siryn-arm-lin[-relay]` | load = `citrineamd2` | `citrineload` |
  | `aspnet-citrine-ampere` *(yml only)* | db & load = `asp-citrine-amd2` | `asp-citrine-db` / `asp-citrine-load` |

[...]

PR #2170, opened 2026-05-07
Switch pod-scheduler configs from JSON to YAML; add JSON Schema

LoopedBard3:
## Why

Follow-up to #2167. The pod-scheduler configs introduced there were JSON, which makes them awkward to read and edit — no inline comments, lots of quote/comma noise, and `type: 3` is a magic integer you have to grep `models.py` to understand. The named-key `machines: {sut: ..., load: ..., db: ...}` shape also turned out to be more friction than information.

This PR switches the source-of-truth configs to YAML, ships a JSON Schema, and simplifies the pod definition along the way.

### Format

- Move `build/benchmarks_ci{,_azure,_cobalt}_pods.json` → `.yml`. Each opens with:
  ```yaml
  # yaml-language-server: $schema=../scripts/pod-scheduler/pod-config.schema.json
  ```
  so VS Code / Cursor / the Red Hat YAML extension provide autocomplete, hover docs, and inline validation while editing.
- Block style throughout; inline comments document the things that used to live only in PR review threads (e.g., that `gold-lin` and `gold-win` share `gold-db`).

### Pod shape

`machines` and `profiles` are now **parallel arrays** of 1–3 unique non-empty strings. `machines[i]` runs `profiles[i]`, and the role of slot `i` is implicit by length (1 = SUT, 2 = SUT + load, 3 = SUT + load + db). No `sut`/`load`/`db` keys anywhere.

```yaml
- name: gold-lin
  machines:
  - gold-lin
  - gold-load
  - gold-db
  profiles:
  - gold-lin-app
  - gold-load-load
  - gold-db-db
```

The loader enforces, per pod:
- `len(machines) == len(profiles)` — each machine pairs with exactly one profile,
- both lists are unique within the pod,
- bools (the YAML `yes`/`no` trap) and other non-strings are rejected.

Global `machine → one profile` is intentionally **not** enforced — `azure2-db` legitimately serves the `db` role in `azure2-amd64` and the `load` role in `idna-amd-win` with different profiles in each. A regression test pins that down.

### Self-documenting scenario type

`type: 1 | 2 | 3` → `type: single | dual | triple` (case-insensitive). Integer form is still accepted for back-compat; bools and unknown strings raise `ConfigError`.

### Loader

- `yaml.safe_load` for any path. YAML is a strict superset of JSON, so existing `.json` configs continue to parse transparently — no extension dispatch.
- Numeric fields (`timeout`, `target_yaml_count`, `schedule_offset_hours`, `estimated_runtime`) go through helpers that reject bools and enforce range minimums so a `timeout: yes` can't quietly resolve to 1.

### Schema

- New `scripts/pod-scheduler/pod-config.schema.json` (Draft 2020-12). Schema descriptions double as grounding for LLM-driven edits.
- Validated locally with the `jsonschema` package; all three configs validate clean.
- `additionalProperties: false` on pods and scenarios catches misspelled keys at edit time (e.g. `mahcines:` instead of `machines:`).

### Model

- `Scenario.type` is a plain `int` (1/2/3 with `SCENARIO_TYPE_*` named constants). The previous `ScenarioType` IntEnum is gone.
- `Pod` stores `machines: List[str]` and `profiles: List[str]` directly — no per-role fields. `machines_for_type(t)` and `profiles_for_type(t)` are just slices.
- Adding a 4th role later only requires bumping the schema's `maxItems`, appending to `ROLE_NAMES`, and adding one alias entry. The rest of the code derives everything from `len(pod.machines)`.

### Generated pipelines

- `benchmarks-ci-01.yml`, `-02.yml`, `-azure.yml`, `-cobalt.yml` regenerated. The only delta vs HEAD is the embedded regen-command line in each file's header (`.json` → `.yml`); scheduling output is byte-identical, enforced by the snapshot test.

### Dependencies and tooling

- New `scripts/pod-scheduler/requirements.txt` declares the runtime PyYAML dependency. The README gets a Prerequisites subsection pointing at it so a clean checkout has a documented `pip install -r ...` step.

### Upstream sync

- The branch was rebased through upstream #2169, which retired the EOL'd `intel-lin`, `intel-win`, `intel-perflin`, and `amd-lin2` pods. The YAML config and pipeline YAMLs reflect that prune; Azure and Cobalt configs were unaffected.

### Tests

- 68 unit tests pass. New coverage includes:
  - All three string aliases for `type` (case variations) + integer back-compat.
  - Bool rejection on every numeric field (type, timeout, estimated_runtime, target_yaml_count, schedule_offset_hours) and on entries inside `machines`/`profiles`.
  - Pod construction invariants: length mismatch, empty list, too-long list, duplicate entries.
  - Same physical machine playing different roles across pods (global non-1:1 invariant).
  - `.json` back-compat via the YAML parser.

### Docs

- Top-level `README.md`, `build/README.md`, and `scripts/pod-scheduler/README.md` rewritten with YAML examples, the schema-directive convention, and the parallel-array pod shape.

## Verification

```bash
python -m pip install -r scripts/pod-scheduler/requirements.txt
cd scripts/pod-scheduler
python -m unittest discover tests
# Ran 68 tests in ~0.3s — OK
```

The snapshot test enforces that regenerating from the new YAML configs produces byte-identical pipeline YAML, so this is a config/format refactor with no behavioral change.

## Notes for reviewers

- `_format_source_path` keeps the repo-relative URL behavior; it now expects `.yml` paths to match how `main.py` is invoked from CI.
- `_parse_scenario_type` explicitly rejects `bool` because `yaml.safe_load` turns `yes`/`no`/`true`/`false` into Python bools, which would otherwise quietly resolve to `1`/`0` via the integer alias. The same bool-guard runs on every numeric field.
- The schema's `maxItems: 3` is the single remaining hardcoded role-count cap; everything else in the code path derives from `len(pod.machines)`.


build/benchmarks_ci_azure_pods.yml
```diff
@@ -0,0 +1,104 @@
...
+    machines: { sut: azure-server-arm64,  load: azure-client,  db: azure-db }
+    profiles: { sut: azure-server-arm64-app, load: azure-client-load, db: azure-db-db }
+
+  - name: azure2-amd64
+    machines: { sut: azure2-server-amd64, load: azure2-client, db: azure2-db }
+    profiles: { sut: azure2-server-amd64-app, load: azure2-client-load, db: azure2-db-db }
+
+  # eastus2 cobalt-cloud pods, merged in when the dedicated pipeline was retired.
+  - name: cobalt-cloud-lin
+    machines: { sut: cobalt-cloud-lin-server, load: cobalt-cloud-lin-client, db: cobalt-cloud-lin-db }
```
sebastienros: why does `machines` need names like `sut`,`load`, ...

build/benchmarks_ci_pods.yml
```diff
@@ -24,34 +24,40 @@ metadata:
...
+# 2 entries = [sut, load]          (dual-type scenarios)
+# 3 entries = [sut, load, db]      (triple-type scenarios)
+# The named-dict form ({sut: ..., load: ..., db: ...}) is still accepted
+# if you ever need to declare just a subset of roles explicitly.
 pods:
   - name: intel-lin
-    machines: { sut: intel-lin, load: intel-load,  db: intel-db }
-    profiles: { sut: intel-lin-app, load: intel-load-load,  db: intel-db-db }
+    machines: [intel-lin, intel-load,  intel-db]
+    profiles: [intel-lin-app, intel-load-load,  intel-db-db]
```
sebastienros: What I meant is that the one of these should not need the role name (application, load, db), scheduling doesn't care about it.

build/benchmarks_ci_pods.yml
```diff
@@ -24,34 +24,40 @@ metadata:
...
 
   - name: intel-win
     # intel-win shares intel-db with intel-lin; the scheduler keeps them out
     # of the same stage when both run a triple-type scenario.
-    machines: { sut: intel-win, load: intel-load2, db: intel-db }
-    profiles: { sut: intel-win-app, load: intel-load2-load, db: intel-db-db }
+    machines: [intel-win, intel-load2, intel-db]
```
sebastienros: Feels weird to use the json syntax in yml (though totally valid)
```
machines:
- intel-win
- intel-load2
- intel-db
```

Also note that YML "IS" JSON, so a JSON document is valid yaml, in case in python you'd want to only handle the yml parsing and not check the extension, just say it's yaml.

LoopedBard3: Went ahead and went forward with updated to the yml format.

build/benchmarks_ci_pods.yml
LoopedBard3: Went ahead and got rid of explicit keys entirely as the profiles selected determine the specific roles. If we wanted to require the role portion of a profile (app/load/db) we could do better verification, but that is part of why we have both machines and profiles.

PR #2185, opened 2026-07-23
Use 1ES images on internal benchmark pool

mmitche:
Replace MMS images with 1ES Windows and Ubuntu images and move the benchmark pipelines to NETCore1ESPool-Internal.

Validation:
- Regression-bot run 3030031 succeeded on the dev branch.
- Pull Request Benchmarks run 3030030 is queued/not started.
- Previous runs on ASP.NET-Performance-Controller-new failed because the 1ES image names were not registered in that pool.

### DIFF
--- a/build/prbenchmarks.yml
+++ b/build/prbenchmarks.yml
@@ -49,7 +49,7 @@ extends:
   parameters:
     pool:
       name: ASP.NET-Performance-Controller-new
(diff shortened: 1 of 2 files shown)


sebastienros:
Actually there might a reason for using a different pool. IIRC it was made by Ilya and John Prather because we needed custom permissions ... I need to search in my inbox.

